PR #78, opened 2014-11-21
Make Data.Tree.flatten a good producer

treeowl:
Write it using `build` to fuse with `foldr` et al.


m-renaud:
If you're still pursuing this could you add benchmarks for this? I'm hoping to get around to adding tests and benchmarks for `Tree` soon(ish) so you can wait until I have that set up if you want.

treeowl:
I have the rumbling feeling this is actually a bad idea. Trees tend not to work so great with fold/build deforestation, as I understand it, and there isn't much to win anyway.

PR #113, opened 2014-12-24
Write custom foldl' and foldr' methods

treeowl:
Additionally, use coercions to speed up folds.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -169,10 +169,7 @@ import Control.DeepSeq (NFData(rnf))
 import Control.Monad (MonadPlus(..), ap)
 import Data.Monoid (Monoid(..))
 import Data.Functor (Functor(..))
-import Data.Foldable (Foldable(foldl, foldl1, foldr, foldr1, foldMap), foldl', toList)
-#if MIN_VERSION_base(4,8,0)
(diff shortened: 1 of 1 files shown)


sjakobi:
@treeowl 

> Hang on with this for a bit. I just realized some things can safely be made stricter, which could help GHC make better code. For example, we can safely make `foldr'` and `foldl'` for `Node` strict in the initial value of the accumulator, and we may be able to play some similar tricks with `FingerTree`. I'm going to need to run some benchmarks and pore over some Core to figure out just how to do this right. Separating the first layer of the `FingerTree` from the rest (which will likely help with this) may also avoid the need for at least some of the coercions.

If this PR already is an improvement on the status quo, I'd advocate merging it and leaving the other ideas for future work. 

In general, I think we should try not letting the perfect stand in the way of good.

treeowl:
Fair.

sjakobi:
Good! :)

@treeowl I'd suggest that you fix the merge conflicts then and bring this to a state where you're happy with it, and then say when you want a review.

Maybe @Lysxia could help with the review then?! I think he knows `Data.Sequence` much better than I do!

Lysxia:
Sure, you can ping me to have a look at this.

PR #120, opened 2014-12-30
Add filterM

treeowl:
Thus far, this proposal has gotten good reception on the libraries
list.


m-renaud:
Appologies if this is too tangential: I haven't seen a proposal but is there any reason why `filterM` from base couldn't be changed to:

```
filterM :: (Applicative m, Foldable t) => (a -> m Bool) -> t a -> m [a]
```

Then not every data structure would have to provide its own implementation of these functions since it would just use the Foldable instance, which Sequence already has. This is obviously a much larger change, but was wondering if this has been considered.

PR #121, opened 2014-12-30
Add intersperse for sequences

treeowl:
`intersperse` is just like the one in `Data.List`. It is
implemented using `<**>` for near-optimal performance.


treeowl:
I've backed off adding `intercalate` for right now because it's not particularly efficient. I may change my mind later.

PR #126, opened 2015-01-01
Replace unfoldTreeM_BF and unfoldForestM_BF

treeowl:
Use radically simpler implementations that do not use
Data.Sequence.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -2363,8 +2363,7 @@ instance Functor PQL where
 instance Show e => Show (PQueue e) where
     show = unlines . draw . fmap show
 
--- borrowed wholesale from Data.Tree, as Data.Tree actually depends
--- on Data.Sequence
+-- borrowed wholesale from Data.Tree
 draw :: PQueue String -> [String]
 draw (PQueue x ts0) = x : drawSubTrees ts0
   where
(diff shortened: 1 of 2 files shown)


treeowl:
Do not merge (yet)! There are currently no QuickCheck properties for `Data.Tree`, and no Criterion benchmarks. I think this is probably right, and likely fast, but I *don't know*. Note that I used `splitAlong` to slightly enhance laziness. An alternative would be to fuse length calculations with the `concatMap`, which would likely perform differently, for better or worse, and would also have different space properties.

treeowl:
Actually, I think I can do better. Still pounding on it.

treeowl:
I though I could maybe structure it better than this, but maybe not. Trying to avoid the pointless-looking `concatMap` followed by `rebuild` keeps running me into things that would make this `O(n^2)`, which would not be cool. Maybe someone else can think of something, but this is the best I have for now.

treeowl:
The second commit is a stricter version. I think it matches the current semantics, and should perform better. All that assuming I got it right, which, considering how late it is, I may not have.

treeowl:
It's morning, so I have a new idea that might get the laziness and the efficiency: in the `concatMap` "stream", insert markers indicating breaks at different levels. That way, rebuilding won't need lengths.

treeowl:
Thinking aloud just a bit more, what I want `rebuild` to become is something akin to a parser for parenthesized expressions, but adjacent parentheses pointing the same direction get merged into each other and counted. The merging is required to avoid `O(n^2)` performance on trees in which each node has only one child.

PR #127, opened 2015-01-02
Alternative `unfoldTreeM_BF` and `unfoldForestM_BF` implementation

treeowl:
I think my initial optimism regarding laziness was misplaced, sadly. But I still think this is a pleasant implementation that we might want to consider.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -2363,8 +2363,7 @@ instance Functor PQL where
 instance Show e => Show (PQueue e) where
     show = unlines . draw . fmap show
 
--- borrowed wholesale from Data.Tree, as Data.Tree actually depends
--- on Data.Sequence
+-- borrowed wholesale from Data.Tree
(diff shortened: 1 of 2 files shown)


foxik:
If we want this merged, we need both tests (to be sure the behaviour is not changed) and benchmarks. Maybe we will need to use multiple monads (Identity/Maybe and List) in those?

treeowl:
Yes, we certainly need those. `Identity` gets a nice comparison to `unfoldTree` and `unfoldForest`; `[]` gets a lot of ways for things to go wrong. We also should include a "strict" monad like `Control.Monad.Trans.State.Strict` or `Control.Monad.ST` to make sure those don't break. One challenging thing will be comparing strictness properties of the different implementations; I think they're probably about the same, but I have no definite proof.

treeowl:
That's funny. I forgot I'd opened this one. I've since come up with one that seems to perform rather better.

PR #133, opened 2015-01-16
Improve MIN_VERSION_base fall-back

treeowl:
Guess the base library version based on `__GLASGOW_HASKELL__`
when compiling without Cabal.


foxik:
Thanks!

Issue #39, opened 2014-02-21
Data.Tree additional functions


jcristovao:
Hi,
The provided tree api seems a little small. After doing the usual workflow of defining my own version of, for example, `subtreeat :: Eq a => a -> Tree a -> Maybe (Tree a)`, only to later find it defined somewhere at hackage, my question is:

Does it make sense to include some of the functions defined in:
http://hackage.haskell.org/package/debian-3.81/docs/Debian-Apt-Dependencies.html#g:3
and 
http://hackage.haskell.org/package/hledger-lib-0.22.1/docs/Hledger-Utils.html#v:subtreeat
As functions of Data.Tree, or should that be left for a separate package?

My imediate suggestions would be:
treefilter :: (a -> Bool) -> Tree a -> Tree a
subtreeat :: Eq a => a -> Tree a -> Maybe (Tree a)
subtreeinforest :: Eq a => a -> [Tree a] -> Maybe (Tree a)

Or am I missing some easier way of avoiding these functions?

m-renaud:
It's been quite some time since you has been updated, was there any resolution on the libraries list for these? If not I was planning on going through the Tree api and putting together a proposal for functions to be added, I can include these in the list.

meooow25:
Some Tree functions were added in #1107.

I'm not convinced of the utility of `treefilter`, `subtreeat` and `subtreeinforest`. They have also been dropped at some point from `hledger-lib`, they are not present in the package today.

Please reopen if there is still a demand for these.

PR #140, opened 2015-03-06
A few more extractors for Data.Tree

athanclark:
I've found some use-cases for looking at all the "paths" that a tree has within it - both "complete" paths (all transitively connected nodes from the root to the end-points), and "partial" paths (all steps in-between on the paths).

The two functions I've added shouldn't clash with other namespaces easily, and they're simple enough to see they are correct. I haven't added any quickcheck tests or criterion benchmarks yet, please let me know if you'd like me to do this.

Thank you for maintaining this beast of a library! I'd also like to fulfill some of the wishes in #39 later if possible.


foxik:
Hello,

thanks for spending time improving `containers`.

Nevertheless, as noted in the README, please note that all API enhancements should be discussed on libraries@haskell mailing list. Please follow https://wiki.haskell.org/Library_submissions#Guide_to_proposers and start a discussion there.

athanclark:
I apologize! I'll present these ideas to the list and keep this thread up-to-date on any progress. Sorry about this.

foxik:
Do not apologize, it is perfectly fine. The process of public discussion is quite unexpected (at least it was for me when I started with this).

However, it allows people to be involved in API design of widely used packages (or at least monitor it) and it makes the API changes better (people can discuss proper names, parameters, consistency with other similar libraries [that is not the case with `Tree`, but is for `Map` etc. where we have `{,Int}{Set,Map}` here and also `Hash{Set,Map}` in unordered-containers).

treeowl:
@athanclark, how was your proposal received on the list?

treeowl:
I also have some doubts about the efficiency of these. Might it be better to form the paths backwards and then reverse them?

recursion-ninja:
Wouldn't forming the paths backwards and later reversing them greatly hinder laziness and make the operations on infinite trees impossible?

treeowl:
@recursion-ninja, that's a problem, for sure, but I think there's already a problem. Look at the definition of `paths`.

```haskell
paths (Node a as) = map (a:) $ concatMap paths as
```

The first element of the result is strict in the first element of `concatMap paths as`. You end up recursively descending the whole tree to get the first element of the result, if I'm not mistaken. There may be a way around this; I haven't dug into the details.

treeowl:
@athanclark Are you still hoping to get some of these in? I haven't seen a libraries proposal or any progress on efficient implementation.

meooow25:
Closing in favor of #1109, which adds `leaves` and `pathsFromRoot`.

PR #141, opened 2015-03-10
Clean up <*> some more

treeowl:
1. Remove all partial functions and all "impossible" errors.

2. Simplify the way the sequence pieces are put together at the
bottom.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -271,53 +271,67 @@ instance Monad Seq where
 
 instance Applicative Seq where
     pure = singleton
-
-    Seq Empty <*> xs = xs `seq` empty
-    fs <*> Seq Empty = fs `seq` empty
-    fs <*> Seq (Single (Elem x)) = fmap ($ x) fs
-    fs <*> xs
-      | length fs < 4 = foldl' add empty fs
-      where add ys f = ys >< fmap f xs
-    fs <*> xs | length xs < 4 = apShort fs xs
-    fs <*> xs = apty fs xs
-
     xs *> ys = replicateSeq (length xs) ys
 
--- <*> when the length of the first argument is at least two and
--- the length of the second is two or three.
-apShort :: Seq (a -> b) -> Seq a -> Seq b
-apShort (Seq fs) xs = Seq $ case toList xs of
-            [a,b] -> ap2FT fs (a,b)
-            [a,b,c] -> ap3FT fs (a,b,c)
-            _ -> error "apShort: not 2-3"
+    fs <*> xs = case viewl fs of
+      EmptyL -> empty
+      firstf :< fs' -> case viewr fs' of
+        EmptyR -> fmap firstf xs
+        Seq fs''FT :> lastf -> case (rigidify . (\(Seq a) -> a)) xs of
+             RigidEmpty -> empty
+             RigidOne (Elem x) -> fmap ($x) fs
+             RigidTwo (Elem x1) (Elem x2) ->
+                Seq $ ap2FT firstf fs''FT lastf (x1, x2)
+             RigidThree (Elem x1) (Elem x2) (Elem x3) ->
+                Seq $ ap3FT firstf fs''FT lastf (x1, x2, x3)
+             RigidFull r@(Rigid s pr _m sf) -> Seq $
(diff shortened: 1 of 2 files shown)


foxik:
Could you also add <*> benchmark changes to the commit message, please?

treeowl:
Yes. I haven't actually run the benchmarks yet. I would personally be willing to tolerate a modest performance decline to get cleaner code, but my guess is that performance won't be affected one way or the other.

foxik:
Yeah, I am definitely not saying the performance cannot deteriorate. But I think benchmark information should be available in the commit message for a reference.

treeowl:
We got a modest improvement in the benchmarks (yay!). Unfortunately, there's a build failure for GHC 7.10 and HEAD, and I can't figure out why. Do you think you could look into that?

treeowl:
FYI, the only further change I'd like to make to the `<*>` code is to make `rigidify` non-recursive. I just have to see if there's a way to do it without making that already-long function too much longer.

treeowl:
It looks like something is funky with dependencies or something:

```
Building regex-base-0.93.2...
Installed regex-base-0.93.2
Downloading regex-posix-0.95.2...
Failed to install regex-posix-0.95.2
Last 10 lines of the build log ( /home/travis/.cabal/logs/regex-posix-0.95.2.log ):
cabal: /home/travis/.cabal/logs/regex-posix-0.95.2.log: does not exist
The command "cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit'" failed and exited with 1 during .
```

foxik:
The build failure is caused by `<$>` being now exported from Prelude (since yesterday) and conflicting with [ansi-wl-pprint-0.6.7.1](https://hackage.haskell.org/package/ansi-wl-pprint-0.6.7.1/docs/Text-PrettyPrint-ANSI-Leijen.html), which has its own `<$>` operator. This will probably take some time to get fixed.

Another note, when adding benchmark results, please attach _difference_ instead of absolute numbers. You can get those by running (with criterion version 1.*; if you have older criterion you have to use different options that `Makefile` currently uses) the following in the benchmarks subdirectory:
- with the original repository state (without the changes being added) run `make BENCHMARK='<*>' bench-Sequence.csv` and then `mv bench-Sequence.csv bench-Sequence.csv.ori`
- with the new repository state (with the changes being commited) run `make BENCHMARK='<*>' bench-Sequence.csv`
- run `./bench-cmp.sh bench-Sequence.csv.ori bench-Sequence.csv` to get the difference results (I just commited a patch which adds column description to the output)

foxik:
The problem in ansi-wl-pprint-0.6.7.1 is already reported, see batterseapower/ansi-wl-pprint#11.

treeowl:
I got the benchmarks in. I had to edit `bench-cmp.sh` to remove the `-n` option from `column`—my local `column` does not support that option, and I couldn't even find documentation online that includes it, so I'll open a ticket to deal with that.

foxik:
On all my Linux machines `man column` contains
```
     -n      By default, the column command will merge multiple adjacent
             delimiters into a single delimiter when using the -t option; this
             option disables that behavior. This option is a Debian GNU/Linux
             extension.
```
I am using it automatically whenever using column, but we do not need it in `bench-cmp.sh`, so I will remove it.

treeowl:
Interesting. My `column` comes from `util-linux-2.24.2` on Fedora 20, which does not offer it. If it's actually important anywhere in `containers`, it would probably pay to come up with a more portable approach.

treeowl:
Is there anything more you need here, or are we ready to merge?

foxik:
As for the `-n` option, as mentioned in the man page, it is Debian specific (I have Debian on my machines and Ubuntu at work). I already commited a patch removing it.

foxik:
We are ready to merge it. I did not test it on 7.10, but we will deal with that once the ansi-wl-pprint is fixed.

foxik:
I forgot -- thanks :-)

treeowl:
Woohoo!

PR #146, opened 2015-03-16
Add chunksOf to Data.Sequence

treeowl:
Break up a sequence into pieces of a given size. Based on
`Data.List.Split.chunksOf` and implemented using `splitMap`.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -126,6 +126,7 @@ module Data.Sequence (
     -- * Sublists
     tails,          -- :: Seq a -> Seq (Seq a)
     inits,          -- :: Seq a -> Seq (Seq a)
+    chunksOf,       -- :: Int -> Seq a -> Seq (Seq a)
     -- ** Sequential searches
     takeWhileL,     -- :: (a -> Bool) -> Seq a -> Seq a
     takeWhileR,     -- :: (a -> Bool) -> Seq a -> Seq a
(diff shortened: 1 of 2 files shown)


treeowl:
I keep wondering if there's some cleverer way to do this, but I haven't found one yet. In particular, this implementation completely ignores the structure of the original sequence. It's not clear to me whether it's possible to exploit that structure somehow.

treeowl:
The build failure is that same old `ansi-wl-pprint-0.6.7.1` issue.

foxik:
We'll wait some more if anyone replies on libraries@...

treeowl:
The only real API question in my mind is whether we should bother to make a version that leaves the excess at the front as well as one that leaves it at the end. If someone really wants the excess at the front, they can always do their own arithmetic and split before calling this version. The fact that the `Seq` version gives an error for non-positive arguments instead of doing something graceful like the list version is unfortunate, but I don't think there's anything we can do about it. If we had a type for strictly positive numbers, we could use that, but we don't.

foxik:
I think we should not add a version with excess at front.

From my point, the interesting API question is if anyone (probably except for you) is interested in such a function and wants it :-)

treeowl:
That's a valid question, but from my perspective the following argue in its favor:

1. `Data.List.Split` offers such a function.
2. The public `Data.Sequence` API does not offer a way to implement it as efficiently.

foxik:
I understand that the method is potentially useful, but I would also like to hear from some folks that they need such a function. Personally I do not know `Data.List.Split` and do not have it installed.

But let's see what happens on libraries.

treeowl:
We can wait another week or two, sure. Data.List.Split is in the Haskell
Platform.

PR #147, opened 2015-03-16
Add IsString instance

treeowl:
Add `instance IsString (Seq Char)` when compiling with GHC.


treeowl:
Here too, the failure is not our fault.

foxik:
This is fairly noncontroversial, and the new PVP allows only minor version bump on non-orphaned instances, so merging without further discussion.

PR #148, opened 2015-03-16
Export cycleN

treeowl:
Export `cycleN` and document its performance.


treeowl:
As I mentioned in my message to the libraries list, the name is based on that of `iterateN`, which similarly adapts a function producing an infinite list to a finite context. The only other sensible API would be to make `cycleN` take the length of the result rather than the number of repetitions, but I think the latter is more intuitive.

treeowl:
I've decided to make `cycleN` take the desired length of the result, rather than the number of repetitions. This should work better for the common-for-lists combination of `cycle` and `zipWith`.

PR #156, opened 2015-05-19
Removed unnecessary strictness in IntSet.foldl accumulator.

### LINKED ISSUE #155: IntSet.foldl is unnecessary strict in accumulator
`IntSet.foldl` is unnecessary strict in accumulator apparently due to copy-paste error from `IntSet.foldl'` (note `STRICT_1_OF_2(go)` macro):

```haskell
foldl :: (a -> Key -> a) -> a -> IntSet -> a
foldl f z = \t ->      -- Use lambda t to be inlinable with two arguments only.
  case t of Bin _ m l r | m < 0 -> go (go z r) l -- put negative numbers before
                        | otherwise -> go (go z l) r
            _ -> go z t
  where
    STRICT_1_OF_2(go)
    go z' Nil           = z'
    go z' (Tip kx bm)   = foldlBits kx f z' bm
    go z' (Bin _ _ l r) = go (go z' l) r
{-# INLINE foldl #-}
```

This bug reveals itself in two kinds of errors (as it usually happen with unnecessary strict functions):

1. It returns bottom when it isn't expected:

  The following program (being compiled with `-O2` and `base>=4.7.0.0`) will fail with run-time error `Prelude.head: empty list`:

  ```haskell
  import qualified Data.IntSet as IS
  main = print $ head $ IS.toDescList $ IS.fromAscList [0]
  ```

  GHC uses rewrite rules:

  ```haskell
  {-# RULES
  "head/build"    forall (g::forall b.(a->b->b)->b->b) .
                  head (build g) = g (\x _ -> x) badHead
  #-}

  {-# RULES "IntSet.toDescList" [~1] forall s . toDescList s = build (\c n -> foldlFB (\xs x -> c x xs) n s) #-}
  ```

  so that the program turns into this:

  ```haskell
  import qualified Data.IntSet as IS
  main = print $ foldlFB (\_ x -> x) badHead $ IS.fromAscList [0]
    where
      badHead = error "Prelude.head: empty list"
      foldlFB = IS.foldl
  ```

  which eventually results in `badHead` error because `IntSet.foldl` is strict in accumulator.

2. It performs unnecessary computations:

  The following program witness that evaluation of `head . IntSet.toDescList` for a `IntSet` of size `n` takes `O(n)` instead of `O(W)`:

  ```haskell
  import qualified Data.IntSet as IntSet
  import System.TimeIt
  main = do
    let a = IntSet.fromAscList [0..10000000]
    timeIt $ print $ a == a
    timeIt $ print $ head $ IntSet.toDescList a
  ```

  Timings are:

  ```
  True
  CPU time:   0.22s
  10000000
  CPU time:   0.97s
  ```

The easy fix for both of the issues is just removal of `STRICT_1_OF_2` marco in the `IntSet.foldl` definition.

adubovik:
Fixes #155

### DIFF
--- a/Data/IntSet/Base.hs
+++ b/Data/IntSet/Base.hs
@@ -882,7 +882,6 @@ foldl f z = \t ->      -- Use lambda t to be inlinable with two arguments only.
                         | otherwise -> go (go z l) r
             _ -> go z t
   where
-    STRICT_1_OF_2(go)
     go z' Nil           = z'
     go z' (Tip kx bm)   = foldlBits kx f z' bm
     go z' (Bin _ _ l r) = go (go z' l) r
--- a/containers.cabal
+++ b/containers.cabal
@@ -252,3 +252,21 @@ test-suite intmap-strictness-properties
 
   ghc-options: -Wall
   include-dirs: include
+
+test-suite intset-strictness-properties
+  hs-source-dirs: tests, .
+  main-is: intset-strictness.hs
+  type: exitcode-stdio-1.0
+
+  build-depends:
+    array,
+    base >= 4.2 && < 5,
+    ChasingBottoms,
+    deepseq >= 1.2 && < 1.5,
+    QuickCheck >= 2.4.0.1,
+    ghc-prim,
+    test-framework >= 0.3.3,
+    test-framework-quickcheck2 >= 0.2.9
+
+  ghc-options: -Wall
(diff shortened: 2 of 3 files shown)


foxik:
Nice catch. Thanks for reporting and fixing it.

It is present since 3351ee0, nearly four years ago.

PR #160, opened 2015-06-29
Introduce Map.adjustWithDefault

greydot:
This is particularly useful when initially populating a map.

### DIFF
--- a/Data/Map/Base.hs
+++ b/Data/Map/Base.hs
@@ -131,6 +131,7 @@ module Data.Map.Base (
     -- ** Delete\/Update
     , delete
     , adjust
+    , adjustWithDefault
     , adjustWithKey
     , update
     , updateWithKey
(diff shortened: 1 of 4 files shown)


foxik:
As this is API change, please follow the https://wiki.haskell.org/Library_submissions#Guide_to_proposers, notably please start discussion about `adjustWithDefault` on libraries@haskell.org.

Personally I find this function not worth adding -- there is already a broad API which makes some users uncomfortable, and `adjustWithDefault` can be easily implemented without any loss in efficiency. To me, `adjustWithDefault` would not be used widely enough to be worth the cost of adding it.

treeowl:
This seems to be essentially the same as `insertWith`, but with a more confusing interface. You're welcome to argue your case on the libraries list, but I'm closing this for now.

PR #163, opened 2015-07-28
Add rewrite rule mapKeysMonotonic/coerce

nomeata:
to avoid traversing the tree just to do nothing.

This also adds a test case that checks this, and also checks the already
existing map/coerce rule.


nomeata:
Ah, ditch that. Won’t work: https://ghc.haskell.org/trac/ghc/ticket/10698#comment:1

PR #164, opened 2015-07-31
Make the 'draw' function works with Show instances

Zigazou:
The original draw functions set works only on Tree/Forest String. This patch allows them to work with any instance of the Show class.


foxik:
Thanks for the patch!

We already have it in develop-0.6 branch, where incompatible changes requiring major version bump are being kept. We will merge it when we do a major version bump, which we try doing not often, as it causes some work for lot of people.

PR #171, opened 2015-11-28
Semigroup instances

hvr:
Admittedly, the CPP isn't pretty but writing the code in this way makes the code future proof

See ekmett/semigroups#56 for more details

### DIFF
--- a/Data/IntMap/Base.hs
+++ b/Data/IntMap/Base.hs
@@ -222,6 +222,9 @@ import Data.Monoid (Monoid(..))
 import Data.Traversable (Traversable(traverse))
 import Data.Word (Word)
 #endif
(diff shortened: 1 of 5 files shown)


foxik:
We are using a lot of CPP anyway, so I do not mind using more. It is not pretty, but it is a burden only to us, not the users.

I did not really check the changes, so I hope you did the right thing :-)

hvr:
Yeah, and `containers` is a bit special here as it's a package bundled with GHC so it can't simply depend on the `semigroups` package and needs to cope with a `Data.Semigroup` module that is only available sometimes.

Packages which can simply depend on `semigroups` uncondintionally don`t need any CPP

PR #178, opened 2016-01-19
Make reverse helpers polymorphic

treeowl:
Lean a bit harder on the type system to ensure that the
reversing function is called everywhere it's supposed
to be. Yes, this should all be covered by the test suite
anyway, but it can't hurt.


treeowl:
This does nothing much, but if we wanted to, we could extend this idea to implement

```haskell
reverseTraverse f = fmap reverse . traverse f
traverseReverse f = traverse f . reverse
```

in one pass. I don't yet know how much we'd gain from those.

foxik:
Thanks!

PR #179, opened 2016-01-19
Add traverseWithIndex

treeowl:
Hack Milan's `mapWithIndex` into `traverseWithIndex`.

Add `default ()` to make sure we never accidentally use defaulting.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -139,6 +139,7 @@ module Data.Sequence (
     foldrWithIndex, -- :: (Int -> a -> b -> b) -> b -> Seq a -> b
     -- * Transformations
     mapWithIndex,   -- :: (Int -> a -> b) -> Seq a -> Seq b
+    traverseWithIndex, -- :: Applicative f => (Int -> a -> f b) -> Seq a -> f (Seq b)
     reverse,        -- :: Seq a -> Seq a
     -- ** Zips
     zip,            -- :: Seq a -> Seq b -> Seq (a, b)
@@ -209,6 +210,7 @@ import qualified GHC.Exts
 import Data.Functor.Identity (Identity(..))
 #endif
 
+default ()
 
 infixr 5 `consTree`
 infixl 5 `snocTree`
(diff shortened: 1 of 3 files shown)


treeowl:
We might want to consider specializing `traverseWithIndex` to the `Identity` applicative, and then redefining `mapWithIndex` in terms of `traverseWithIndex`:

```haskell
#if __GLASGOW_HASKELL__ >= 708
mapWithIndex f = runIdentity . traverseWithIndex (coerce f)
#else
mapWithIndex f xs = runIdentity (traverseWithIndex ((Identity .) . f) xs)
#endif
```

Stealing bits of the `Profunctor (->)` instance, we can even write

```haskell
mapWithIndex = (runIdentity .) #. traverseWithIndex .# coerce
```

which is essentially the same as `mapWithIndex = traverseWithIndex` except it satisfies the type system.

Some benchmarking is in order, I think. I'm not sure when I'll have a chance to do such.

foxik:
Could you please also add a test to tests/seq-properties for `traverseWithIndex`?

In this pull request I would only add the `traverseWithIndex`; changing implementation of `mapWithIndex` will require some benchmarking, so if you have time to perform them, start another pull request. But I am reasonably happy with the current situation, even if we have some very similar duplicated code.

treeowl:
Yes, a test is definitely in order. I would have added one last night, but
I ran out of time. I think a simple State thing will probably do, or maybe
[]. I'll try to get it done today.

treeowl:
I've added a test. It's not the most carefully crafted thing in the world, but it should be good enough under the circumstances.

foxik:
Thanks

treeowl:
@foxik, you're most welcome, of course. The only things I see remaining in this space are:

#### Building applicatively with an index
This would basically mix code from `fromFunction` with code from `applicativeTree`. It could be seen as an applicative version of `fromFunction` or a generalization of `replicateA`.

```haskell
replicateAWithIndex :: Applicative f => Int -> (Int -> f a) -> f (Seq a)
replicateAWithIndex i f = sequenceA (fromFunction i f)
```
This obeys
```haskell
replicateAWithIndex i (f . g) = traverse f (fromFunction i g)
```

#### `foldMapWithIndex`

I asked about this before, and it never went anywhere.

```haskell
foldMapWithIndex f = fold . mapWithIndex f
```

but it skips the intermediate structure.

foxik:
Maybe the `replicateAWithIndex` should be called `fromFunctionA`?

Both these functions seem to be just faster versions of compositions of two existing functions -- so from my point of view, either the new functions should be much faster or should be regularly used API calls. Nevertheless, just go through the libraries@, I think you will get enough +1s :-)

PR #181, opened 2016-02-10
repair broken URL in IntMap documentation

jwaldmann:
the URL in the current version is broken (or paywalled)


foxik:
Thanks for noticing!

PR #182, opened 2016-02-10
fix typo in documentation

jwaldmann:
was: IntMap, should be: Map


foxik:
Thanks for noticing :-)

PR #183, opened 2016-02-16
Delete tests-ghc

thomie:
`containers` received its own testsuite in 38743e394dd19fcd7983a1f726482d04e64111ec (2010). That commit mentions:

    "It includes unit tests for known past bugs (balancing)."

I think it's safe to delete the old tests in the `tests-ghc` directory. 

I'm planning to do some changes to the ghc testsuite. Deleting the `tests-ghc` directory means I don't have to apply those changes to `containers`.

Edit: I'm not sure about the unreliable tests. They were added only recently, but the ghc testsuite doesn't run them.


foxik:
We are keeping the tests-ghc just because GHC uses these; ourselves we only use the tests in the `tests` subdirectory.

Nevertheless, even if the tests in `tests-ghc` are trivial, I thought that it is good that some simple containers testing happens during `make test` in GHC. But containers are now used inside GHC, so if something was wrong, the whole GHC would collapse, I presume.

Still -- could you get someone other from GHC (for example hvr) to ACK here that it is fine removing `tests-ghc`?

thomie:
@hvr, @bgamari: are you ok with deleting the `libraries/containers/tests-ghc` directory? It contains just 4 tests that we still run on each validate.

The ghc tests in `filepath` and `directory` were deleted recently as well. They now have their own testsuite, and GHC basically expects these libraries to "just work".

bgamari:
Sure, this sounds fine to me. Thanks for handling this @thomie and @foxik!

PR #184, opened 2016-02-18
add Generics instance for Map, Set, IntMap, and IntSet

kuk0:
we want Generics for everything (otherwise we can't derive Generics for any data types which contain these without creating orphan instances)

the end-goal is to have all of these: https://gist.github.com/kuk0/92fff1c91c526ffb0a9e
and be able to derive instances for e.g. TextShow or Text.PrettyPrint.Generic.Pretty

### DIFF
--- a/Data/IntMap/Base.hs
+++ b/Data/IntMap/Base.hs
@@ -8,6 +8,8 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 #if __GLASGOW_HASKELL__ >= 708
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE EmptyDataDecls #-}
 #endif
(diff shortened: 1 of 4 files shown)


treeowl:
I'm a bit concerned that `to` and `from` here are only one-sided inverses. In particular, multiple representations may map to the same map/set. This probably isn't a good enough reason to reject your proposal, but it's something to think about.

foxik:
@treeowl This is a bit out of my league. David, could I ask for your advice? Regarding your concern, is there any way to do it better?

treeowl:
I know only a little about this so far, but it mostly looks about the same as other generics instances I've seen. As for my concern, no, there is no way to do better without full dependent types and lots of pain.

foxik:
@treeowl Thanks.

@kuk0 Thanks for the patch.

Issue #195, opened 2016-04-21
Fix Generics for 8.0


treeowl:
I'm seeing

```
Data/Set/Base.hs:352:41: error:
    Not in scope: type constructor or class ‘NoSelector’
    Perhaps you meant ‘Selector’ (imported from GHC.Generics)
```

We need to fix this ASAP!

kosmikus:
These new hand-written non-standard `Generic` instances seem to have been introduced comparatively recently. Is it clear that we really want them? I'm generally in favour of treating abstract types as abstract, and not having `Generic` instances for them.

The internal representation in GHC-8 has changed. Fixing them is relatively straight-forward, but will require CPP trickery to make work with both GHC-8 and older.

treeowl:
They're non-standard to maintain the abstraction. Personally, I'm not really a big fan, but it's quite possible someone is relying on them, and I don't know that it's possible to define them better.

kosmikus:
No, they clearly should be defined in the non-standard way they are, if they're to be defined at all. But I just looked at the original PR that introduced them which states

>    we want Generics for everything (otherwise we can't derive Generics
>    for any data types which contain these without creating orphan
>    instances)

but I don't think this justification is valid. We can very well derive `Generic` instances for datatypes containing abstract types, because `Generic` only defines a top-level transformation and does not touch the contained types.

I think it's unlikely that very many people rely on these (yet), because they don't seem to be in any released version of `containers`.

Anyway, it's not my decision to make. I can only say that I'd be leaning towards not having them at all.

This is what I think should be the diff for e.g. `Data.Set.Base`:
```
diff --git a/Data/Set/Base.hs b/Data/Set/Base.hs
index 47efe85..79be739 100644
--- a/Data/Set/Base.hs
+++ b/Data/Set/Base.hs
@@ -11,6 +11,9 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE EmptyDataDecls #-}
 #endif
+#if __GLASGOW_HASKELL__ >= 800
+{-# LANGUAGE DataKinds #-}
+#endif
 
 #include "containers.h"
 
@@ -339,6 +342,8 @@ setDataType = mkDataType "Data.Set.Base.Set" [fromListConstr]
 {--------------------------------------------------------------------
   A Generic instance
 --------------------------------------------------------------------}
+#if __GLASGOW_HASKELL__ < 800
+
 data D1Set
 data C1Set
 
[...]

PR #197, opened 2016-04-21
Revert "add Generics instance for Map, Set, IntMap, and IntSet"

treeowl:
Reverts haskell/containers#184

In #195, Andres Löh argues that these instances should not have been added. I think we should, therefore, revert them unless @kuk0 provides a more convincing demonstration that they are really a good idea.

### DIFF
--- a/Data/IntMap/Base.hs
+++ b/Data/IntMap/Base.hs
@@ -8,8 +8,6 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 #if __GLASGOW_HASKELL__ >= 708
 {-# LANGUAGE TypeFamilies #-}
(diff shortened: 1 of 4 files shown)


treeowl:
I'd really like to hear from @wrengr about this.

wrengr:
I agree with Andres. If the instances aren't actually required for the stated purpose (other folks deriving generics) and they're not actually used anywhere, then we should drop them. The state of the art for generic stuff keeps moving, so having unnecessary instances just increases maintenance burden.

Of course, if someone comes up with another more convincing argument for why the instances are necessary/helpful, then it'd be worth revisiting.

PR #215, opened 2016-05-03
Add `at` for Data.Map

treeowl:
Use a bit queue to implement `alterF`, a flipped version of `at`,
for `Data.Map`. This is fairly competitive with the simple implementation in
`Control.Lens.At` even with `Int` keys. For keys that are more expensive to compare,
it should be substantially better.

Unit tests and benchmarks were written by Phil Ruffwind, who proposed
adding `at` to begin with. The implementation is by David Feuer, as are
the QuickCheck properties.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,8 @@ GNUmakefile
 dist-install
 ghc.mk
 .stack-work
+/benchmarks/bench-Map
+/benchmarks/bench-Set
+/benchmarks/bench-IntSet
+/benchmarks/bench-IntMap
+/benchmarks/bench-Sequence
--- a/Data/Map/Base.hs
+++ b/Data/Map/Base.hs
@@ -137,6 +137,7 @@ module Data.Map.Base (
     , updateWithKey
     , updateLookupWithKey
     , alter
+    , at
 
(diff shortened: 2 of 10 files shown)


treeowl:
This needs some (fairly thorough) QuickCheck properties for `at`. It also desperately needs QuickCheck properties for the bit queue, which I hope someone else will be able to speed up.

treeowl:
Information on why a more "natural" implementation is too slow can be found in @Rufflewind's original pull request #192.

Things to remember before merging this:

1. Consider making the name and API match the rest of the library better.

2. Write a stricter version for the strict module--this one can install bottoms.

3. There may be lots of better ways to implement the "bit queue", but the first thing to do is to try to use only one conditional branch based on whether the low word is zero, fusing the double word shift code into the rest of `buildQ`.

Rufflewind:
I'm no expert with bit manipulation either, so I don't really see a way it could be improved beyond simplifying `buildQ`. I'm also not sure it would buy more than a few nanoseconds since `buildQ` is run only once.

It might be a good idea to use `wordSize - 1` instead of the hard-coded `63` in the `Show` instance though.

PR #216, opened 2016-05-04
Use unsafeShift in BitUtil for base >= 4.8

Rufflewind:
Use `unsafeShiftL/R` in `BitUtil` for base >= 4.8 instead of explicitly using GHC primitives, and also avoid the warning about `Trustworthy`.

### DIFF
--- a/Data/Utils/BitUtil.hs
+++ b/Data/Utils/BitUtil.hs
@@ -27,7 +27,13 @@ module Data.Utils.BitUtil
 
 import Data.Bits ((.|.), xor)
 
-#if __GLASGOW_HASKELL__
+#if !MIN_VERSION_base(4,8,0)
+import Data.Word (Word)
+#endif
+
+#if MIN_VERSION_base(4,5,0)
+import Data.Bits (unsafeShiftL, unsafeShiftR)
+#elif __GLASGOW_HASKELL__
 import GHC.Exts (Word(..), Int(..))
 import GHC.Prim (uncheckedShiftL#, uncheckedShiftRL#)
 #else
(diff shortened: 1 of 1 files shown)


treeowl:
The comments accompanying those indicate the purpose is to ensure inlining.
The Bits Int instance does not INLINE those methods, so they won't be
inlined in unfoldings, I don't think.

treeowl:
Of course, it's possible that's not necessary, but it would be good to be
sure.

Rufflewind:
You're right.  It doesn't seem to inline those … I don't understand why `shift` has INLINE but `unsafeShift` don't :/

https://hackage.haskell.org/package/base-4.8.2.0/docs/src/GHC.Word.html

~~~hs
shiftRL2 :: Word -> Int -> Word
…
shiftRL2 =
  \ (ds_d1VQ :: Word) (ds1_d1VR :: Int) ->
    case ds_d1VQ of _ [Occ=Dead] { W# x_a1MJ ->
    case ds1_d1VR of _ [Occ=Dead] { I# i_a1MK ->
    GHC.Types.W# (uncheckedShiftRL# x_a1MJ i_a1MK)
    }
    }

shiftRL :: Word -> Int -> Word
…
shiftRL = Data.Bits.$fBitsWord_$cunsafeShiftR
~~~

treeowl:
The comment in `Data.Bits` is not terribly informative. I would think these should *all* be marked `INLINE` or `INLINE CONLIKE`. I suggest you open a ticket upstream to consider this and link from here. I'll leave this ticket open pending a decision upstream.

Rufflewind:
> I suggest you open a ticket upstream to consider this and link from here.

https://ghc.haskell.org/trac/ghc/ticket/12022

m-renaud:
There doesn't appear to be any updates on the upstream ticktet and the Bits instance for Int in base-4.10.0.0 still doesn't have unsafeShiftL and unsafeShiftR marked INLINE. I'll follow up on the trac ticket when I get a chance.

m-renaud:
See updated comments in https://ghc.haskell.org/trac/ghc/ticket/12022. It sounds like the advice is still to let the compiler determine when to INLINE. mpickering also points out that the `shiftLL` function has different semantics depending on if `__GLASGOW_HASKELL__` is set or not.

I don't know enough about optimization at this point to comment on what we should do here.

sjakobi:
I think this can be closed. Today we have

```
shiftRL = unsafeShiftR
shiftLL = unsafeShiftL
```

PR #232, opened 2016-05-17
add a catamorphism on Trees

dmwit:
Inspired by [this SO question](http://stackoverflow.com/a/37279548/791604), it seems there's no fold for trees! That's odd.

### DIFF
--- a/Data/Tree.hs
+++ b/Data/Tree.hs
@@ -30,7 +30,7 @@ module Data.Tree(
     -- * Two-dimensional drawing
     drawTree, drawForest,
     -- * Extraction
-    flatten, levels,
+    flatten, levels, foldTree,
     -- * Building trees
     unfoldTree, unfoldForest,
     unfoldTreeM, unfoldForestM,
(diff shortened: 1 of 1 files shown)


treeowl:
I'm very much in favor, but you should send a message to libraries@haskell.org to make sure there aren't any major quibbles about the name. It would also be helpful to benchmark your implementation against some variants; at least try these:

```haskell
foldTree = go
  where
    go f (Node x ts) = f x (mapgo f ts)
    mapgo _ [] = []
    mapgo f (t : ts) = go f t : mapgo f ts
```

```haskell
foldTree f (Node x ts) = f x (go f ts)
  where
    go _ [] = []
    go f ((Node x xts) : ts) = f x (go f xts) : go f ts
```

My reasoning is that the paper on the constant argument transformation suggests that lifting out just one constant argument usually reduces performance.

treeowl:
I guess that second one is stricter than expected. Try

```haskell
foldTree f (Node x ts) = f x (go f ts)
  where
    go _ [] = []
    go f (t : ts) = f x p : go f ts
      where p = case t of Node x xts -> go f xts
```

treeowl:
On the other hand, your way should get some nice list fusion if `f` is implemented using `foldr` .... Hrm...

Issue #230, opened 2016-05-16
Data.Map.Strict improperly re-exports traverseWithKey


treeowl:
`Data.Map.Strict` re-exports `traverseWithKey` from `Data.Map.Base`. It should instead export a stricter version:

```haskell
traverseWithKey :: Applicative t => (k -> a -> t b) -> Map k a -> t (Map k b)
traverseWithKey f Tip = pure Tip
traverseWithKey f (Bin _ k v Tip Tip) = (\ !v' -> Bin 1 k v' Tip Tip) <$> f k v
traverseWithKey f (Bin s k v Tip r) = (\ !v' r' -> Bin s k v' Tip r') <$> f k v <*> traverseWithKey f r
traverseWithKey f (Bin s k v l Tip) = (\l' !v' -> Bin s k v' l' Tip) <$> traverseWithKey f l <*> f k v
traverseWithKey f (Bin s k v l r) = (\l' !v' r' -> Bin s k v' l' r') <$> traverseWithKey f l <*> f k v <*> traverseWithKey f r
```

treeowl:
Fixed by 26624a2007f217140a45b4803d322dd3f3feb3fb.

PR #239, opened 2016-05-20
Make Data.Map.Strict.traverseWithKey strict enough

treeowl:
Previously, `Data.Map.Strict` re-exported `traverseWithKey`
from `Data.Map.Base`. That function could produce a map containing
bottoms.


treeowl:
This fixes #230

PR #243, opened 2016-05-22
Integrate benchmarks with `cabal`.  Fixes #188

### LINKED ISSUE #188: Integrate benchmarks with `cabal` and `stack`?
`cabal bench` and `stack bench` both support running benchmarks defined in `containers.cabal` as `Benchmark` suites.  For example, the benchmark suite for `benchmarks/Set.hs` module would look like this:

```
Benchmark set-benchmarks
    type: exitcode-stdio-1.0
    hs-source-dirs: benchmarks
    main-is: Set.hs
    ghc-options: -O2
    -- The final version would have dependency bounds
    build-depends: base >= 4.2 && < 5, containers, criterion, deepseq, mtl
```

This will not affect support for older `cabal` versions, because `cabal` ignores unrecognized sections for forwards compatibility.  Older versions of `cabal` will emit the following harmless warning when encountering these `Benchmark` sections:

```
Cabal file warning in path/to/containers/containers.cabal: Ignoring unknown section type: benchmark
```

I'd be happy to contribute this change if you are okay with it

Gabriel439:
Some comments about the changes that might seem unrelated:

* I changed to more explicit imports so that it was easier to figure out what the correct bounds should be for each benchmark's dependencies
* Several benchmarks were (unnecessarily) using an old version of `defaultMainWith` that only worked before `criterion-1.0.0.0`.  I replaced this with just evaluating the values in the top-level `main` function

### DIFF
--- a/benchmarks/IntMap.hs
+++ b/benchmarks/IntMap.hs
@@ -1,10 +1,9 @@
 {-# LANGUAGE BangPatterns #-}
 module Main where
 
-import Control.DeepSeq
+import Control.DeepSeq (rnf)
 import Control.Exception (evaluate)
-import Control.Monad.Trans (liftIO)
-import Criterion.Main
+import Criterion.Main (bench, defaultMain, whnf)
 import Data.List (foldl')
 import qualified Data.IntMap as M
 import Data.Maybe (fromMaybe)
--- a/benchmarks/IntSet.hs
+++ b/benchmarks/IntSet.hs
@@ -2,10 +2,9 @@
 
 module Main where
 
-import Control.DeepSeq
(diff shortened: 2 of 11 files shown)


Gabriel439:
Hmmm.  I'm not sure why it says that this has conflicts that must be resolved as I just forked off of the latest `master` for this branch.

treeowl:
I've been doing a lot of work lately. Sorry about that! Do you want to
merge, or shall I?

Gabriel439:
Oops!  For some reason I thought I had already branched off the most recent `master`.  I will open a separate pull request with the fixed version here: https://github.com/haskell/containers/pull/245

PR #245, opened 2016-05-22
Integrate benchmarks with `cabal`. Fixes #188

### LINKED ISSUE #188: Integrate benchmarks with `cabal` and `stack`?

Gabriel439:


treeowl:
Thanks for all the great work!

Gabriel439:
You're welcome!

Issue #290, opened 2016-06-27
Add lookup function to Data.Set


DrNico:
## WHAT?

It is proposed to add a `lookup` function on `Set` in the "containers" package. The proposal is logged here first for reference in a later discussion on the Haskell libraries mailing list.

The function

```
lookup :: Ord a => a -> Set a -> Maybe a
```

is almost indentical to the `member` function but, in addition, returns the value stored in the set.
## WHY?

The point of this proposal is to facilitate program-wide data sharing. The `lookup` function gives access to a pointer to an object already stored in a Set and equal to a given argument. The `lookup` function is a natural extension to the current `lookupLT`, `lookupGT`, `lookupLE` and `lookupGE` functions, with obvious semantics.

Example use case: In a parser, the memory footprint can be reduced by collapsing all equal strings to a single instance of each string. To achieve this, one needs a way to get a previously seen string (internally, a pointer) equal to a newly parsed string. Amazingly, this is very difficult with the current "containers" library interface. One current option is to use a Map instead, e.g., `Map String String` which stores twice as many pointers as necessary.
## HOW?

The Pull Request (to be added shortly) contains the straight-forward implementation of the `lookup` function on `Set`, with test cases.

treeowl:
I don't really like the name, since it doesn't really act like lookup for
maps. You should certainly send the proposal to the libraries list, but I
personally like the general idea.

m-renaud:
Hey @DrNico, did you get around to sending a proposal to the libraries list? I noticed in PR #291 that the current suggested name is `lookupEntry` which SGTM. Also, could you add a similar function to `Data.IntSet` and `Data.IntMap` (commented there as well)?

andreasabel:
Maybe `lookupKey`?

andreasabel:
> [@andreasabel](https://github.com/andreasabel) see [#587](https://github.com/haskell/containers/issues/587) for the broader issue here.

Yes that is where I came from...

meooow25:
Ah I didn't realize. Because what name we use is secondary to whether we should add this at all.

meooow25:
Closing as not planned (see #587)

PR #291, opened 2016-06-27
Add 'lookup' function to Data.Set 

DrNico:
This is an implementation of the proposal submitted as Issue #290, but not yet discussed on the Haskell libraries list.

### DIFF
--- a/Data/Set.hs
+++ b/Data/Set.hs
@@ -63,6 +63,7 @@ module Data.Set (
             , size
             , member
             , notMember
+            , lookup
             , lookupLT
             , lookupGT
             , lookupLE
@@ -146,6 +147,7 @@ module Data.Set (
             ) where
 
 import Data.Set.Base as S
+import Prelude ()
 
 -- $strictness
 --
--- a/Data/Set/Base.hs
+++ b/Data/Set/Base.hs
@@ -114,6 +114,7 @@ module Data.Set.Base (
             , size
             , member
             , notMember
+            , lookup
             , lookupLT
             , lookupGT
             , lookupLE
@@ -194,7 +195,7 @@ module Data.Set.Base (
             , merge
             ) where
 
-import Prelude hiding (filter,foldl,foldr,null,map)
+import Prelude hiding (filter,foldl,foldr,null,map,lookup)
 import qualified Data.List as List
 import Data.Bits (shiftL, shiftR)
 #if !MIN_VERSION_base(4,8,0)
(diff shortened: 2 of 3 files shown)


treeowl:
Unless someone comes up with a better name, or a good reason to dislike this one, I'd like to go with @andreasabel's suggestion of `lookupEntry`, and to add to `Data.Map` `lookupEntry :: Ord k => k -> Map k v -> Maybe (k, v)`.

m-renaud:
Could you also add a similar function to `Data.IntSet` and `Data.IntMap`?

treeowl:
@m-renaud In the library list discussion, several people thought this was a bad idea. I believe that included @wrengr. They thought this sort of interning operation would be best handled by another data structure. Nothing about this is at all relevant to `Data.IntMap` or `Data.IntSet`; they (operationally) store `Int#` values, not `Int` values, so there can be no duplication.

treeowl:
@m-renaud, sorry, I ascribed the wrong objection to @wrengr. She opposed calling this a name relating to *interning*. The main dispute on the libraries list had to do with the name. The name `lookup` seems very strange to me here.

treeowl:
The mailing list discussion thread begins with https://mail.haskell.org/pipermail/libraries/2016-June/027116.html

m-renaud:
> Nothing about this is at all relevant to Data.IntMap or Data.IntSet

Ahh, that makes sense.

> The main dispute on the libraries list had to do with the name

Gotcha. Reading through the discussion it seems like everyone was okay with `lookupEntry` as you mention in your comment above. The only hesitation was that for sets the term "entry" doesn't existing in the vocabulary. Should we start a new thread with a proposal with the functions to make sure everyone is on board still?

```
lookupEntry :: Ord a => a -> Set a -> Maybe a
lookupEntry :: Ord k => k -> Map k v -> Maybe (k, v)
```

treeowl:
Go ahead.

m-renaud:
Giving this some more thought, why add `lookupEntry` for Data.Map? Map already has `lookup :: k -> Map k v -> Maybe v`, so if you really wanted you could implement `Map.lookupEntry k m = (\v -> (k,v)) <$> lookup k m` without loss of performance.

Also, since "entry" is really parlance related to maps, removing this from Map means we could use the more appropriate term "elem(ent)": `lookupElement :: (Ord a) => Set a -> Maybe a`.

WDYT?

treeowl:
The purpose for maps would be to extract the exact stored key. The keys in
the map correspond to the ones in a set; values are extra.

m-renaud:
> The purpose for maps would be to extract the exact stored key. 

Thanks for the clarification; I retract my previous comment. I'll go ahead and send a proposal to libraries for the two `lookupEntry` methods in https://github.com/haskell/containers/pull/291#issuecomment-353995588.

sjakobi:
> Thanks for the clarification; I retract my previous comment. I'll go ahead and send a proposal to libraries for the two `lookupEntry` methods in [#291 (comment)](https://github.com/haskell/containers/pull/291#issuecomment-353995588).

Did that libraries proposal ever happen?

PR #292, opened 2016-06-28
Spitzner nines

treeowl:


treeowl:
@lspitzner, I've made a stab at documenting the `fromList` design, changed some formatting, and removed a bang pattern (for `n1` in `getNodes`) that I don't *think* we have any particular reason to have. Your input would be greatly appreciated.

lspitzner:
@treeowl I think that your explanation is clear and it matches my current understanding. The slight changes you added all make sense. I'd be fine with this being merged.

(It is unfortunate that the explanation for the difference of performance (also in contrast to the pre-strict-list version) is somewhat uncertain.)

PR #317, opened 2016-08-08
Add general merge functions for Data.Map

treeowl:
Add `generalMerge` and `generalMergeA`. `generalMerge` is a
safe version of `mergeWithKey` that offers nearly as much useful
power but prevents users from accidentally violating internal
invariants and preserves the abstraction barrier. `generalMergeA`
is an `Applicative` version.


treeowl:
I realized that I was too quick to dismiss the possibility of a Haskell 98 version of this function. I'll add that instead.

PR #321, opened 2016-08-25
Add more indexed and unsafe functions for Data.Map

treeowl:
* Offer `take`, `drop`, and `splitAt` by index.

* Offer 'takeWhileAntitone`, `dropWhileAntitone`, and `spanAntitone`.


treeowl:
I need to write at least a few tests before merging this!

cgibbard:
Are you sure that the monotonic splits are O(n)? The goal was to use the property of the function to obtain the split in O(log n) time, because we only have to cut along a path from the root to the element where the predicate is changing.

treeowl:
Of course they're not. The monotonicity condition is also documented wrong.
I'll fix both.

PR #324, opened 2016-08-31
Refactor internal modules

esoeylemez:
I have:

  * renamed the `.Base` modules to `.Internal`,
  * updated all references to them,
  * unhidden them from Haddock,
  * hidden utility modules (see note!).

**Note**: According to the changelog @ekmett requested internal modules to be exposed.  I don't know if the unfortunately named utility modules (`Data.Utils.*`) were part of that request, but since they mostly contain generic utility functions I have operated under the assumption that they weren't.

If they were, just ignore the last commit and merge 17d7896 instead.


treeowl:
Some of the Util things are silly. Some of them could be useful to others.
I suspect the bit queue should be exposed for monkeying with Data.Map; the
bit utils may be helpful for people digging into Data.IntMap. Thanks for
the work so far! Would you mind putting together such a compromise?

esoeylemez:
I have reintroduced `.BitQueue` together with `.BitUtil`, as it will most likely be useful for this purpose. It would be better to give those modules better/less generic names though, perhaps something like `Data.ContainerUtils.*`.

treeowl:
You're right that Utils is too generic. I don't want to push it further up
the alphabet than necessary though. Data.ContainerUtils would be the very
first thing anyone saw in Haddocks! I also think that it would have been
(and can be!) better to put the warning messages about internal modules in
one separate file referenced everywhere. Thanks for the work so far!

esoeylemez:
Glad to help. If you need more adjustments on this, feel free to ask!

PR #339, opened 2016-09-10
Make min and max views more eager

treeowl:
- Make min and max views in `Data.IntMap` more
  eager, and structure them to reveal `Nothing` or `Just` as soon as
  possible. This lets GHC erase `Maybe` constructors and unbox tuples, at
  least when it's in a good mood.
- Add `lookupMin` and `lookupMax` to `Data.IntMap`
  as total alternatives to `findMin` and `findMax`.
- Add `!?` to `Data.IntMap` as a total alternative to `!`.
- Set the fixity of `Data.IntMap.!` to match that of `Data.Map.!`.
- Make more update functions total. Historically, many have thrown
  errors when given missing keys. Now, we prefer to return the structures
  unchanged.

### DIFF
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -72,7 +72,7 @@ module Data.IntMap.Internal (
       IntMap(..), Key          -- instance Eq,Show
 
     -- * Operators
-    , (!), (\\)
+    , (!), (!?), (\\)
 
     -- * Query
     , null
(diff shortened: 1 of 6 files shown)


treeowl:
@wrengr, does this all look sensible to you?

wrengr:
Two comments (already discussed, but for posterity):

* in `{min,max}ViewWithKey` the case analysis on `{min,max}ViewWithKeyBin` should be pushed down under the `Just` constructor. Unlike the lookup functions, these two don't have any space leak concerns, since the view data types contain all the information stored in the original map.

* in `{min,max}View` it'd be a lot cleaner to use Maybe's `fmap`/ `(<$>)` rather than inlining it.

Other than that, LGTM

treeowl:
I think this has all been sorted now.

PR #348, opened 2016-10-17
Improve documentation of mapKeysWith

benjamin-hodgson:
The documentation for `mapKeysWith` was missing a description of which argument to the combining function is which.


treeowl:
Thanks! Sorry it's taken so long to merge this.

PR #352, opened 2016-10-23
Data.Sequence.Internal: Fix CPP usage

erikd:
There was a mixture of `#ifdef TESTING` and `#if TESTING`. The later
works, but is not really correct. GHC HEAD now has a `-Wcpp-undef`
warning that we would like to turn on and hence need this fixed.


treeowl:
I'm fine with this sort of change, but we also have a ton of `#if
__GLASGOW_HASKELL__ >= ...`. Do these all need to check whether the macro
is defined first?

erikd:
@treeowl They should be fine because GHC is the only Haskell compiler with the `-Wcpp-ndef` warning flag :)

erikd:
Being a bit less flippant, `#if __GLASGOW_HASKELL__` is always going to be OK with GHC, since GHC has defined that value since version 4 or something. That construct could potentially be a problem for other Haskell compilers, but only if they implement something like the `-Wcpp-udef` flag I just added to GHC git HEAD.

erikd:
Ping? I'd like to get this into containers so I can move forward with building GHC with `-Wcpp-undef` by default.

treeowl:
Sorry, I've been a bit busy. I'll try to merge it tonight.

treeowl:
Sorry that took so long.

erikd:
No problem. Thanks @treeowl !

PR #353, opened 2016-10-24
Add another way to build benchmarks

osa1:
Building benchmarks while using `containers` as a library never worked
for me, because some of the dependencies are usually always installed
system-wide (`binary`, an older version of `containers` etc.).

So this package instead uses the library files as source files (instead
of using the library as a library dependency). This means that we link
whatever `containers` library available with dependencies like `binary`
and `criterion`, but we use `containers` source files from the parent
directory when compiling benchmark programs. In my case this always
works.

One disadvantage is that we re-compile `containers` source files again
and again for every single benchmark, so compile times suffer.

### DIFF
--- a/benchmarks/benchmarks.cabal
+++ b/benchmarks/benchmarks.cabal
@@ -1,0 +1,181 @@
+-- This package doesn't use `containers` as a library, instead compiles used
+-- modules from the library for each executable.
+--
+-- Relative paths are OK here as we don't generate tarball for this package.
+
+name:           benchmarks
+version:        0.1
+license:        BSD3
+license-file:   ../LICENSE
+maintainer:     libraries@haskell.org
+build-type:     Simple
+cabal-version:  >=1.8
+
+executable intmap-benchmarks
+    hs-source-dirs:     . ../
+    main-is:            IntMap.hs
+    ghc-options:        -O2
+
+    build-depends:
+        base,
+        criterion,
+        deepseq
+
+    if impl(ghc)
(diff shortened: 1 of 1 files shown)


treeowl:
I have done it successfully the current way, and others have too. The trick
is to install the dependencies manually. If you can find a way to avoid
that for free, I'm all ears, but your proposal sounds very expensive.

osa1:
I don't understand how can it work, maybe I don't know some things about Cabal. Since `containers` upper bound for dependencies will be lower than the current (in HEAD) `containers`, I think it'll always fail. Can you tell me what exactly do you do to build it?

RyanGlScott:
@osa1: I was able to get it to work by first preinstalling all of the dependencies. I believe just:

```
cabal install criterion random
```

Followed by (within the `containers` directory):

```
cabal configure --enable-benchmarks
cabal bench
```

treeowl:
We clearly need to document the procedure. @RyanGlScott, would you like to put together a PR, or should one of the maintainers do it?

treeowl:
I'm going to close this. I think documenting how to run the benchmarks is preferable to slowing everything down and making invasive changes to the package structure. I've opened #371 to remind us to write the necessary documentation.

osa1:
> slowing everything down

> making invasive changes to the package structure

No big deal, but this just adds a new cabal file, which is not used unless user explicitly runs it. I don't understand how it can slow everything down or how it is an invasive change.

PR #358, opened 2016-11-16
IntMap general merge

wrengr:
I've implemented a version of `mergeA` (and sundry other applicative functions) for `IntMap`.

### DIFF
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -236,7 +236,9 @@ module Data.IntMap.Internal (
     , highestBitMask
     ) where
 
-#if !(MIN_VERSION_base(4,8,0))
+#if MIN_VERSION_base(4,8,0)
+import Data.Functor.Identity (Identity (..))
+#else
 import Control.Applicative (Applicative(pure, (<*>)), (<$>))
 import Data.Monoid (Monoid(..))
 import Data.Traversable (Traversable(traverse))
@@ -272,6 +274,7 @@ import qualified GHC.Exts as GHCExts
 #endif
 import Text.Read
 #endif
+import qualified Control.Category as Category
 #if __GLASGOW_HASKELL__ >= 709
 import Data.Coerce
 #endif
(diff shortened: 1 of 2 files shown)


Data/IntMap/Internal.hs
```diff
@@ -1253,9 +1934,11 @@ maxViewWithKey t =
...
 minViewWithKey t =
-  case t of Nil -> Nothing
-            Bin p m l r | m < 0 -> case go r of (result, r') -> Just (result, binCheckRight p m l r')
-            _ -> Just (go t)
+  case t of
+    Nil -> Nothing
+    Bin p m l r | m < 0 ->
+      case go r of (result, r') -> Just (result, binCheckRight p m l r')
+    _ -> Just (go t)
   where
```
treeowl: I suspect a custom triple type will be easier for GHC to unbox than a pair with a pair in it. Also, I seem to remember that you suggested I get lazier about producing the `Just` constructor for `Data.Map`. We should probably try to be consistent about that one way or the other.

```haskell
data MinViewWithKeyResult a = MVWKR {-# UNPACK #-} !Key a !(IntMap a)
minViewWithKey t =
  case t of
    Nil -> Nothing
    Bin p m l r | m < 0 ->
      Just $ case go r of MVWKR k a r' -> let !t' = binCheckRight p m l r' in ((k,a), t')
    _ -> Just $ case go t of MVWKR k a t' -> ((k,a), t')
```

Data/IntMap/Internal.hs
```diff
@@ -1193,6 +1227,651 @@ mergeWithKey' bin' f g1 g2 = go
...
+  -> WhenMissing f x b
+mapGentlyWhenMissing f t = WhenMissing
+  { missingSubtree = \m -> fmap f <$> missingSubtree t m
+  , missingKey     = \k x -> fmap f <$> missingKey t k x }
+{-# INLINE mapGentlyWhenMissing #-}
+
+
+-- | Map covariantly over a @'WhenMatched' f k x@, using only a
+-- 'Functor f' constraint.
+mapGentlyWhenMatched
```
treeowl: I don't think I've actually exported these gentle functions yet. We should talk about whether we want to.

Data/IntMap/Internal.hs
```diff
@@ -1740,25 +2433,38 @@ split k t =
 splitLookup :: Key -> IntMap a -> (IntMap a, Maybe a, IntMap a)
```
treeowl: Might it pay to hold GHC's hand using `StrictTriple` for the recursion here? That should make it super-easy for GHC (and us) to see that the three pieces are all in WHNF at each stage and at the end.

Data/IntMap/Internal.hs
```diff
@@ -2370,30 +3097,30 @@ showTreeWith hang wide t
   | otherwise = (showsTree wide [] [] t) ""
```
treeowl: Can we add some deprecation pragmas to eventually move these debugging functions out of `Data.IntMap` and into a debugging module?

Data/IntMap/Internal.hs
```diff
@@ -1193,6 +1227,651 @@ mergeWithKey' bin' f g1 g2 = go
...
+--
+-- * 'dropMissing' drops all the keys.
+-- * 'preserveMissing' leaves all the entries alone.
+-- * 'mapMaybeMissing' does not use the 'Applicative' context.
+--
+-- When 'mergeA' is given three arguments, it is inlined at the call
+-- site. To prevent excessive inlining, you should generally only use
+-- 'mergeA' to define custom combining functions.
+--
+-- @since 0.5.8
```
treeowl: At last, my love has come along.
My lonely days are over,
and life is like a song!

Thanks a whole lot for going through this horrible slog!

Data/IntMap/Internal.hs
wrengr: For the time being I'm fine with it either way. That they're there at all is a cut&paste job :)

Data/IntMap/Internal.hs
wrengr: Can do

Data/IntMap/Internal.hs
wrengr: Can do

Data/IntMap/Internal.hs
wrengr: can do

Data/IntMap/Internal.hs
wrengr: Everything looks like it's already whnf by construction, so it shouldn't change anything. But I can make the change so that it's more explicit

PR #360, opened 2016-12-02
Data.Map.Internal: Fix CPP usage

erikd:
Switch from `#if` to `#ifdef` on conditionally defined values.


erikd:
Thanks @treeowl !

treeowl:
Thank you!

PR #376, opened 2016-12-30
Kill getSingleton

treeowl:
It's partial, and was used in a silly way for no particularly
good reason. We still need a partial function, but we can write
it where we use it.


treeowl:
This needs to be benchmarked before it's merged!

treeowl:
I changed my mind. This is sufficiently obviously The Right Thing that I don't want to bother benchmarking first. Passing the `getSingleton` thunk to a user function is just too horrible.

PR #377, opened 2017-01-01
Add function to check if two `IntSet`s are `disjoint`.

vlopezj:
This function is equivalent to computing the intersection and checking
if it is empty. However, it is more efficient because the intersection
set does not need to be built in memory, and the computation can
be short-circuited as soon as two non-disjoint `Tip`s are found.

### DIFF
--- a/Data/IntSet/Internal.hs
+++ b/Data/IntSet/Internal.hs
@@ -118,6 +118,7 @@ module Data.IntSet.Internal (
     , lookupGE
     , isSubsetOf
     , isProperSubsetOf
+    , disjoint
 
     -- * Construction
     , empty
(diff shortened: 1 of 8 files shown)


treeowl:
I like this idea. I think you should probably run it by the libraries list.
However, I wonder if we should consider generalizing. Specifically,

IntSet a ~= IntMap a ()

so we could write a (greatly reduced) version of the mergeA interface
supporting IntSet. It would include

data WhenMissing f
data WhenMatched f
dropMissing :: Applicative f => WhenMissing f
preserveMissing :: Applicative f => WhenMissing f
filterAMissing :: Applicative f => (Int -> f Bool) -> WhenMissing f
filterAMatched :: Applicative f => (Int -> f Bool) -> WhenMatched f

Then we could use a Const-like Applicative wrapping Bool to check if
IntSets are disjoint, but also do various other things that may or may not
prove to actually be useful.

wrengr:
I'm all for having `disjoint` for all the {Int,}{Set,Map} data types. Haven't had a chance to look at the code just yet; will do as soon as I can

treeowl:
@vlopezj I still haven't seen a message to the libraries list. I'd like to hear what others have to say.

vlopezj:
@treeowl Thanks for the reminder! I will send a proposal tomorrow about having `disjoint` for `IntSet`. 
And I like the generalized version; if it does not affect performance too much it can be implemented that way instead.

By the way, here are some benchmarks comparing the specialized `disjoint` implementation with an implementation which only uses the pre-existing API (`(null.).intersection`). The specialized implementation is 23× faster for non-disjoint sets, and 3.6% faster for the disjoint case.

```
benchmarking disjoint:false
time                 176.1 ns   (174.2 ns .. 178.7 ns)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 178.2 ns   (176.7 ns .. 180.4 ns)
std dev              6.094 ns   (4.561 ns .. 9.385 ns)
variance introduced by outliers: 52% (severely inflated)
               
benchmarking disjoint:true
time                 2.465 μs   (2.449 μs .. 2.481 μs)
mean                 2.471 μs   (2.455 μs .. 2.496 μs)
std dev              68.39 ns   (51.73 ns .. 105.5 ns)
variance introduced by outliers: 35% (moderately inflated)
               
benchmarking null.intersection:false
time                 4.143 μs   (4.108 μs .. 4.182 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 4.168 μs   (4.127 μs .. 4.277 μs)
std dev              215.3 ns   (103.2 ns .. 404.0 ns)
variance introduced by outliers: 65% (severely inflated)
               
benchmarking null.intersection:true
time                 2.555 μs   (2.535 μs .. 2.583 μs)
mean                 2.591 μs   (2.566 μs .. 2.629 μs)
std dev              98.45 ns   (66.08 ns .. 150.5 ns)
variance introduced by outliers: 51% (severely inflated)
```

The sets used are [1,3,…,2¹²-1] and [2,4,6,…,2¹²] for the disjoint case, and [1,2,3…,2¹²] and [2,4,6,…,2¹²] for the non-disjoint case.

@wrengr if we have it for `IntMap` also, perhaps it would be better to do it with @treeowl's generalized framework, as the meaning of `disjoint` for maps is less obvious. For example is it about disjoint keys (i.e. empty intersection), or about keys and values (in the spirit of `isSubmapOf`).

m-renaud:
Random thought, instead of adding this to the public API could you instead add a RULES pragma which rewrites `empty . intersection` to the more efficient implementation?

treeowl:
I think it's better to offer a predictably fast function, and not rely on
rules that may fail to fire. Adding rules to use the new function would be
perfectly fine.

On Dec 24, 2017 8:11 PM, "Matt Renaud" <notifications@github.com> wrote:

Random thought, instead of adding this to the public API could you instead
add a RULES pragma which rewrites empty . intersection to the more
efficient implementation?

—

<https://github.com/haskell/containers/pull/377#issuecomment-353811506>,
or mute
the thread

m-renaud:
Agreed, that's what I meant to write originally. I reworded my comment to add RULES in addition to the function, not as a replacement for.

treeowl:
Ah, silly GitHub email interface.

vlopezj:
From the discussion on the mailing list, it seems that adding `disjoint` would be a good idea. A corresponding function should also be added to the `Data.Set` module for consistency.

http://haskell.1045720.n5.nabble.com/PROPOSAL-Add-disjoint-method-to-Data-IntSet-td5868974.html

-----

@treeowl How do we proceed from here?

Data/IntSet/Internal.hs
```diff
@@ -653,6 +654,47 @@ isSubsetOf (Tip _ _) Nil = False
 isSubsetOf Nil _         = True
 
 
+{--------------------------------------------------------------------
+  Disjoint
+--------------------------------------------------------------------}
+-- | /O(n+m)/. Check whether two sets are disjoint (i.e. their intersection
+--   is empty).
```
m-renaud: Can you add some example invocations of the function and an `@since 0.5.11` annotation?

m-renaud:
Can you include the benchmark results for `Set.disjoint` vs. the intersection == empty approach in the commit message?

vlopezj:
@treeowl @m-renaud All done.
One of the tests shows as failed because of an error during test setup. It should pass once restarted, but I cannot do that from here.

treeowl:
One more thing: can you squash the second commit onto the first one? In case you don't know how:

Run `git log`. Find the commit just before the one where you add `disjoint` for `IntSet` and copy its hash.

Run `git rebase -i` with the commit hash you just copied as an argument. This should open up an editor with a list of the three commits here. If you see something else, double check the commit hash you copied earlier and if that's right, stop and ask for help.

The three listed commits should all have the word "pick" next to them. Change the middle one (the docs and `@since` annotations) from `pick` to `squash`. Save the file.

Git will now open up a new editor letting you edit the merged commit messages. You can delete everything except the first commit message. Save the file.

Check `git log` to make sure the log looks reasonable, use `git show` to make sure each commit looks reasonable, and then push. You'll have to pass `--force` when you push.

treeowl:
Thanks, @vlopezj!

PR #379, opened 2017-01-03
Use countTrailingZeros for indexOfTheOnlyBit

alexbiehl:
This is an experiment for `indexOfTheOnlyBit`. I haven't benchmarked yet. But will do so tonight.

### DIFF
--- a/containers/src/Data/IntSet/Internal.hs
+++ b/containers/src/Data/IntSet/Internal.hs
@@ -1418,6 +1418,11 @@ indexOfTheOnlyBit :: Nat -> Int
 {-# INLINE indexOfTheOnlyBit #-}
 #if MIN_VERSION_base(4,8,0) && (WORD_SIZE_IN_BITS==64)
 indexOfTheOnlyBit bitmask = countTrailingZeros bitmask
+
+lowestBitSet x = countTrailingZeros x
+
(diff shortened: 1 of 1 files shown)


treeowl:
@alexbiehl, what's with the regressions in list conversion functions?

treeowl:
I'm sorry I lost track of this. The benchmark numbers you posted look very good, but unfortunately I haven't been able to reproduce your results. Part of the problem may be the general flakiness of the incredibly short benchmarks. Can you suggest a procedure for getting good results reliably?

alexbiehl:
Me neither! I went ahead and further "optimized" the folds by using Count{Leading, Trailing}Zeros but the effect was basically nil. I suggest we abandon this for now.

treeowl:
All right. Let me know. We could use more reliable benchmarks regardless of what happens to this PR.

treeowl:
Do you think you could rebase this? @Fuuzetsu did some benchmarking and found that it at least doesn't seem to hurt. It's definitely less messy and cuts code size in key places.

alexbiehl:
Yeah sure, will do this week.

alexbiehl:
@treeowl Done!

treeowl:
Thanks!

Issue #392, opened 2017-02-06
Data.IntMap.restrictKeys and Data.IntMap.withoutKeys not working


tomsmalley:
I'm using containers-0.5.9.1, this is observable on both strict and lazy versions of IntMap:

```haskell
λ GHCi, version 8.0.1: http://www.haskell.org/ghc/  :? for help
λ import qualified Data.IntMap as M
λ import qualified Data.IntSet as S
λ let m = M.fromList [(1,'a'), (3,'b')]
λ M.restrictKeys m (S.singleton 1)
fromList []
λ M.withoutKeys m (S.singleton 1)
fromList [(1,'a'),(3,'b')]
```

Using the example definitions from the documentation works:

```haskell
λ let restrictKeys m s = M.filterWithKey (\k _ -> k `S.member` s) m
λ let withoutKeys m s = M.filterWithKey (\k _ -> k `S.notMember` s) m
λ restrictKeys m (S.singleton 1)
fromList [(1,'a')]
λ withoutKeys m (S.singleton 1)
fromList [(3,'b')]
```

treeowl:
Oh great. This looks bad. I'll look into it today.

On Feb 6, 2017 10:30 AM, "Tom Smalley" <notifications@github.com> wrote:

I'm using containers-0.5.9.1

λ GHCi, version 8.0.1: http://www.haskell.org/ghc/  :? for help
λ import qualified Data.IntMap as M
λ import qualified Data.IntSet as S
λ let m = M.fromList [(1,'a'), (3,'b')]
λ M.restrictKeys m (S.singleton 1)
fromList []
λ M.withoutKeys m (S.singleton 1)
fromList [(1,'a'),(3,'b')]

Using the example definitions from the documentation works:

λ let restrictKeys m s = M.filterWithKey (\k _ -> k `S.member` s) m
λ let withoutKeys m s = M.filterWithKey (\k _ -> k `S.notMember` s) m
λ restrictKeys m (S.singleton 1)
fromList [(1,'a')]
λ withoutKeys m (S.singleton 1)
fromList [(3,'b')]

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/haskell/containers/issues/392>, or mute the thread

treeowl:
Crud. I totally goofed those implementations, and then I messed up when
adding the functions to the test suite so I didn't notice. Argh! This will
be a scramble.

[...]

PR #400, opened 2017-02-09
Optimize withoutKeys and restrictKeys for IntMap

### LINKED ISSUE #394: Optimize restrictKeys and withoutKeys for IntMap
My quick emergency fix for these functions isn't as fast as it could be when the `IntSet` reaches `Tip` first. We should really go subtree by subtree and bit by bit there.

wrengr:
Optimize IntMap's `withoutKeys` and `restrictKeys`.

Fixes #394

### DIFF
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -1127,6 +1127,8 @@ restrictKeys t1@(Bin p1 m1 l1 r1) t2@(IntSet.Bin p2 m2 l2 r2)
 restrictKeys t1@(Bin _ _ _ _) (IntSet.Tip p2 bm2) =
     -- TODO(wrengr): should we manually inline/unroll 'lookupPrefix'
     -- and 'restrictBM' here, in order to avoid redundant case analyses?
+    -- TODO(wrengr): mask out the too-small and too-large keys
+    -- before entering 'restrictBM', for better IH.
     restrictBM bm2 (lookupPrefix p2 t1)
 restrictKeys (Bin _ _ _ _) IntSet.Nil = Nil
 restrictKeys t1@(Tip k1 _) t2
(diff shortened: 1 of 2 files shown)


Data/IntMap/Internal.hs
```diff
@@ -1043,30 +1043,37 @@ differenceWithKey f m1 m2
...
+        | otherwise         = withoutKeys t1 r2
+-- TODO(wrengr): should we inline the top-level 'withoutBM' here?
+withoutKeys t1@(Bin _ _ _ _) (IntSet.Tip kx' bm') = withoutBM kx' bm' t1
+    where
+    withoutBM !kx !bm t@(Bin p m l r)
+        | nomatch kx p m = t
+        | zero kx m      = binCheckLeft p m (withoutBM kx bm l) r
+        | otherwise      = binCheckRight p m l (withoutBM kx bm r)
+    withoutBM kx bm t@(Tip ky _)
+        -- TODO(wrengr): should we inline 'IntSet.member' here?
```
treeowl: One option might be to break up `member`:

```haskell
member :: Key -> IntSet -> Bool
member !x = go
  where
    go (Bin p m l r)
      | nomatch x p m = False
      | zero x m      = go l
      | otherwise     = go r
    go (Tip y bm) = memberTip x y bm 
    go Nil = False

memberTip :: Key -> Prefix -> BitMap -> Bool
memberTip x y bm = prefixOf x == y && bitmapOf x .&. bm /= 0
{-# INLINE memberTip #-}
```

Then you could reuse `memberTip` here. It's kind of ugly though. Another possibility (with a lot of its own ugliness!) might be to redefine `IntMap` and `IntSet` as fixed points of functors (ouch!) which may allow more pieces to be defined non-recursively so they'll inline better.

wrengr: yeah, I was thinking about doing the `memberTip` thing

wrengr:
Don't merge this yet. Looks like I reintroduced the #392 bug somehow. There's some chicanery about how IntMap uses bitmasks vs how IntSet does.

Data/IntMap/Internal.hs
treeowl: The functorial fixed point may also have some potential applications. I don't know for sure, but I've suspected for some time that it may be possible to do something like that to add optional size annotations to the tree.

wrengr:
That new patch fixes the bugs. However, it's not as performant as we'd like. In particular it's linear in the number of elements in the `IntSet.Tip`, and performs one `lookup` or `delete` per element; so it's /O(m * min(n,W))/, when really we should be able to get something that's /O(n+m)/ just like for `intersect` and `difference`.

wrengr:
This latest version of `restrictKeys` is at least passible performance-wise. (And all tests pass)

**The algorithm:** Consider the case `restrictKeys t@(IntMap.Bin p m _ _) (IntSet.Tip q s)`. We will first use `q` as a partial key for doing a lookup in`t`, thus returning `t'`: the `IntMap` which is rooted at the same place as the set `s`. The `IntSet`-suffix of `p` corresponds to the smallest index in `s` which could possibly correspond to any key in `t'`. Thus, we mask out all bits in `s` which are smaller than that suffix, yielding a subset `s'`. Finally, we do the `IntMap.keysSet` thing to unfold `s'` into a tree structure and check each element against `t`.

This last step is still a bit too slow IMO, as it still takes /O(m * min(n,W))/ in the worst case. Though we do get /m/ as small as possible (by unfolding `s'` rather than `s`), and do make the lookups as fast as possible (by doing lookups in `t'` rather than `t`). Maybe over the weekend I'll have time to take a closer look and work out how to make this part of the algorithm faster.

Still need to perform the analogous transformation/optimization on `withoutKeys` before merging.

wrengr:
Ugh. Why does my local quickcheck say things're fine when Travis's easily finds counterexamples. Will have to work on it some more

treeowl:
You can crank up both the number of tests and the test size on the command
line. Often that will help catch problems. But for this job, you may also
want to ask for large maps with Tip sets.

wrengr:
Found the bug. For posterity: the issue was that the `IntMap` prefix stored in `Bin` isn't guaranteed to have the low-order bits cleared; so we had to call `mask` in order to clear them out.

The current version could be made a bit smarter. I only remove from the bitmap all the keys which are too small to possibly be in the `Bin`. We can/should also remove all the keys which are too big to possibly be in the `Bin`. It's easy enough to do; just takes doing. The current version costs 9 primops (6 logical; 3 arithmetical) to restrict the bitmap, so should almost always be a win.

wrengr:
New version masks out the too-large keys as well. Total cost is: 9 logic ops, 6 arith ops, and 2 temporary registers (assuming `natFromInt` is a no-op; else three temporaries). Still should be almost always worth it.

Another enhancement to add: instead of starting `restrictBM` with `(IntSet.suffixBitMask + 1)`, we should be able to start it with something based on `minkey`. That way we don't have to scan past all the known-zero bits for too-small keys. We could also potentially add an additional argument based on `maxbit`, so we can avoid scanning all the known-zero bits for too-large keys.

wrengr:
Bah! It only passes because I missed a prime mark, so it's using the unfiltered bitmap...

treeowl:
I'm not understanding some of your commentary; it sounds a bit complicated.
It *sounds* (if I'm reading you right) like you're trying to narrow down
the IntSet to keys that could possibly lie in the IntMap; I would have
expected the opposite approach.

We start out with an IntMap Bin and an IntSet (Tip). We first want to know
if any keys in that map could possibly lie in the IntSet. For this initial
configuration, how can we answer that question? How can we arrange to be
able to answer that question recursively as we explore the IntMap?

[Pardon me if these comments make too little sense].

PR #412, opened 2017-02-20
Fix lots of strictness bugs

treeowl:
* `Data.IntMap.Strict` previously re-exported the lazy `traverseWithKey`.
  Implement a strict one.

* `Data.IntMap.Strict` and `Data.Map.Strict` previously had a number of
  rewrite rules with strictness bugs. Remove `map/coerce` rules from
  each and fix the other rules.

### DIFF
--- a/Data/IntMap/Strict.hs
+++ b/Data/IntMap/Strict.hs
@@ -217,62 +217,94 @@ module Data.IntMap.Strict (
 import Prelude hiding (lookup,map,filter,foldr,foldl,null)
 
 import Data.Bits
-import Data.IntMap.Internal hiding
-    ( findWithDefault
-    , singleton
-    , insert
-    , insertWith
-    , insertWithKey
-    , insertLookupWithKey
-    , adjust
-    , adjustWithKey
-    , update
-    , updateWithKey
-    , updateLookupWithKey
-    , alter
-    , alterF
-    , unionsWith
-    , unionWith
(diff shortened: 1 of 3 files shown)


Data/IntMap/Strict.hs
```diff
@@ -799,17 +827,50 @@ mapWithKey f t
...
+-- lsmap _ _ Nil = Nil
+-- lsmap f g (Tip k x) = let !gx = g x in Tip k (f gx)
+-- lsmap f g (Bin p m l r) = Bin p m (lsmap f g l) (lsmap f g r)
 {-# NOINLINE [1] mapWithKey #-}
 {-# RULES
 "mapWithKey/mapWithKey" forall f g xs . mapWithKey f (mapWithKey g xs) =
+  mapWithKey (\k a -> f k $! g k a) xs
+"mapWithKey/mapWithKeyL" forall f g xs . mapWithKey f (L.mapWithKey g xs) =
   mapWithKey (\k a -> f k (g k a)) xs
 "mapWithKey/map" forall f g xs . mapWithKey f (map g xs) =
```
wrengr: Since it's cheap to get the key, would it make sense to rewrite `(L.)map` into `(L.)mapWithKey` early on (e.g., phase 2) and then rewrite it back at the very end? That would help avoid this quadratic explosion of rules in the middle

Issue #423, opened 2017-04-04
Speed up unstable sorting


treeowl:
I [recently figured out](https://www.reddit.com/r/haskell/comments/63a4ea/fast_total_sorting_of_arbitrary_traversable/) how to write a (reasonably) efficient unstable sort for arbitrary `Traversable` containers. Surprisingly, that seems to be competitive with `Data.Sequence.unstableSort`, at least for short to medium-length sequences. I think this means we have some room to improve in `Data.Sequence`. A few possibilities:

1. Check out what if anything should be made stricter (or lazier).

2. Use a different heap representation.

3. Reconsider how we rebuild a sequence from a heap. The fact that we first "unroll" the heap to a list looks suspicious to me.

oisdk:
I've been playing around with this for a while, and I'm getting some decent speedups with a couple of changes. 

### Specialise

The main thing is that adding a specialise pragma to `replicateA` seems to give a 15-30% speedup.

### Removing intermediate list

The list is only being constructed in order to be deconstructed in the `fromList2` function, which only uses unconsing, so it seems like you should be able to use a `minView` function directly:

```haskell
pqState = State minView
```

Or something. However, the heap type can't represent an empty heap:

```haskell
data PQueue e = PQueue e (PQL e)
data PQL e = Nil | {-# UNPACK #-} !(PQueue e) :& PQL e
```

Which means that if you rewrite the unroll function into the direct-state style, it turns out to be too strict. If you replace the implementation of `<*>` with a handwritten definition (instead of `ap`), it's not too strict:

```haskell
fs <*> xs =
  State $ \s -> case runState fs s of
    (s',f) -> case runState xs s' of
      (s'',x) -> (s'', f x)
```

But I don't know if that effects other code negatively. That lets you remove the intermediate list:

```haskell
pqState cmp = State unrollPQ' where
  {-# INLINE unrollPQ' #-}
  unrollPQ' (PQueue x ts) = (mergePQs ts,x)
  mergePQs (t :& Nil) = t
  mergePQs (t1 :& t2 :& Nil) = t1 <+> t2
  mergePQs (t1 :& t2 :& ts) = (t1 <+> t2) <+> mergePQs ts
  mergePQs Nil = undefined
  (<+>) = mergePQ cmp
```

This speedup is less significant, (possible slower on small inputs) but on larger inputs it looks like it's significant.

### A Different Heap

[...]

PR #425, opened 2017-04-30
Speeding up unstableSort

oisdk:
This pull request addresses #423 

The two main improvements to the original function are a specialize pragma on `replicateA` (this gives a 26.5% speedup) and removal of the intermediate list (a further 11.7% speedup). These numbers are all on purely random sequences of length 50000:

    Times (ms)            min    est    max  std dev   R²
    to/from list:        70.90  72.44  75.07  2.224  0.998
    7/20/09 heapsort:    59.84  61.44  63.08  1.554  0.998
    7/20/09 w/pragma:    44.22  45.14  46.25  1.631  0.998
    4/30/17 heapsort:    38.21  39.86  40.87  1.203  0.996

It should also be noted that `Data.List.sortBy` has become significantly quicker since the original `unstableSort` was written. `Data.List.sortBy` also now recognizes strictly increasing sequences, making it much quicker for that case:

    Times (ms)            min    est    max  std dev   R²
    to/from list:        7.140  7.351  7.634  0.335  0.993
    7/20/09 heapsort:    19.52  19.78  20.13  0.445  0.999
    7/20/09 w/pragma:    8.050  8.271  8.514  0.357  0.995
    4/30/17 heapsort:    7.240  7.612  7.925  0.389  0.991

Another happy result of the specialization of `replicateA` is that the stable sort seems to speed up by 10-20%, and `iterateN` looks like it's about three times as fast. Would maybe specializing `replicateA` to `Identity` give similar speedups for other functions?


treeowl:
Wow, those are some impressive results! Thanks!
We don't want quite so many commits. Should I squash them all, or would you like to do some gentler squashing? Also

oisdk:
Squash away!

treeowl:
Very nice results. Thanks a lot! We should definitely look into whether `replicateA` needs further specialization pragmas, or other such modifications.

PR #431, opened 2017-08-18
Add restictKeys example via intersect, also withoutKeys

phadej:
I think this examples would helped me to spot (I tried Ctrl+F with `intersect`).

EDIT: also fixed markup

---

For dual `(set, map) -> set` operations I'd use `union`, `intersection` and `difference` names with some prefix or suffix, as all three are possible. `unionWithMap` etc doesn't sound too bad, does it?

```haskell
unionWithMap        :: Set k -> Map k a -> Set k
intersectionWithMap :: Set k -> Map k a -> Set k
differenceWithMap   :: Set k -> Map k a -> Set k
```


phadej:
ping @treeowl

treeowl:
Sorry about the long delay, and thanks for the patch!

phadej:
no worries.

PR #432, opened 2017-08-22
Expose Size (and other Int aliases) from *.Internal modules

RyanGlScott:
Currently, these aliases are referenced in the Haddocks for the `*.Internal` modules, but because they're not exported, you can't easily see what they represent without digging into the source code. At the very least, we can also export them from the `*.Internal` modules as well.

### DIFF
--- a/Data/IntSet/Internal.hs
+++ b/Data/IntSet/Internal.hs
@@ -98,6 +98,7 @@
 module Data.IntSet.Internal (
     -- * Set type
       IntSet(..), Key -- instance Eq,Show
+    , Prefix, Mask, BitMap
 
     -- * Operators
     , (\\)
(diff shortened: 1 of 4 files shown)


treeowl:
What if we make them newtypes instead? That would actually prevent bugs.
Also, we have a ticket open to add Int32Map and Int64Map. Does that relate?
Probably not.

RyanGlScott:
To be clear, are you pondering a future design for `containers` in https://github.com/haskell/containers/pull/432#issuecomment-324027368? Are are you saying "I won't accept this PR unless this change is made"?

treeowl:
Just speculating.

treeowl:
Let's just do this for now. If we switch to `newtype`s then anyone actually using these will be annoyed, but that's not really our problem.

PR #437, opened 2017-09-10
IntMap lookupMin and lookupMax

bwroga:


Data/IntMap/Internal.hs
```diff
@@ -2091,27 +2094,39 @@ deleteFindMax = fromMaybe (error "deleteFindMax: empty map has no maximal elemen
...
+lookupMin Nil = Nothing
+lookupMin (Tip k v) = Just (k,v)
+lookupMin (Bin _ m l r)
   | m < 0     = go r
   | otherwise = go l
-    where go (Tip k v)      = (k,v)
+    where go (Tip k v)      = Just (k,v)
           go (Bin _ _ l' _) = go l'
-          go Nil            = error "findMax Nil"
+          go Nil            = Nothing
```
treeowl: We have to decide whether the non-`Nil` case should produce `Just` immediately, or wait till the descent is complete. I believe that @wrengr prefers the former and that we do so for `Map`. I have some reservations about that approach from the potential space leak angle, but whichever way we choose we should be consistent.

tests/intmap-properties.hs
```diff
@@ -966,6 +980,18 @@ prop_lookupLE = test_lookupSomething lookupLE (<=)
 prop_lookupGE :: [(Int, Int)] -> Bool
 prop_lookupGE = test_lookupSomething lookupGE (>=)
 
+prop_lookupMin :: [(Int, Int)] -> Property
+prop_lookupMin ys = length ys > 0 ==>
+  let xs = List.nubBy ((==) `on` fst) ys
+      m  = fromList xs
+  in  lookupMin m == Just (List.minimumBy (comparing fst) xs)
```
treeowl: I don't like these properties. They are confusing, and also inefficient. How about this?

```haskell
prop_lookupMin :: IntMap Int -> Property
prop_lookupMin im = lookupMin m === listToMaybe (toAscList m)
```

bwroga: Yes, that is much better.

Data/IntMap/Internal.hs
bwroga: Would it make sense to have both versions, one in Lazy and one in Strict?

treeowl: I don't think that's a very good fit, no. I'd like if @wrengr would remind me why the lazy way might ever be a good idea. I can't remember or think of a really good reason for it right now.

bwroga:
Any word back from @wrengr ?

treeowl:
No, I've heard nothing from @wrengr. I'm just going to merge this, and we can leave the laziness question for the future. Thanks for your work!

PR #451, opened 2017-12-19
Add MonadFix instance for Data.Tree

treeowl:
This is based on an instance for a somewhat more general type
by G. Allais.
See [MonadTree.hs](https://gist.github.com/gallais/4c59b949c743c0a85cab55dcb73aaf7c).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
  - cabal install --only-dependencies
  # we need to install the test-suite and benchmark deps manually as the cabal
  # solver would otherwise complain about cyclic deps
- - cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit'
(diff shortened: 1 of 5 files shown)


treeowl:
@gallais Your review would be greatly appreciated.

treeowl:
In particular, I have no idea how to go about checking the `MonadFix` laws for this instance. Anything beyond the purity law seems to get messy.

treeowl:
@ocharles, actually, I see you were also involved in this. Did you or @gallais figure it out, or do both deserve credit? And do you think it's law-abiding?

ocharles:
https://gist.github.com/ocharles/9b6fb71669de4533373a9c7f1f3ce8f9 appears to work, but I haven't checked the laws hold.

ocharles:
In regards to credit, @gallais deserves all the credit, for sure.

gallais:
As mentioned in [the original discussion](https://stackoverflow.com/questions/47833188/can-a-monadic-rose-tree-have-a-monadfix-instance) this is based off of the intuition I built up reading the `MonadFix` instance for `[]`.

Your implementation looks like what I would have written but I don't think I'm qualified to say whether this is a valid `MonadFix` instance. I'll try to see if I can prove (some of) the laws.

treeowl:
@gallais intuition from the list instance suggests that the purity, sliding, and nesting laws should hold for the same reason they do for lists (whatever that may be). The one that really smells like it needs careful proving is left  shrinking, since it gets into the details of the `mfix`/`>>=` interaction.

Issue #304, opened 2016-07-26
Add validity checks to IntMap and IntSet tests


treeowl:
The other containers include validity checks to ensure that all operations on valid inputs produce valid outputs. For some reason, these seem to be missing from `IntMap` and `IntSet` tests. That makes me nervous!

m-renaud:
I can take a stab at this if no one has started on it yet.

treeowl:
Thanks!

nomeata:
> The other containers include validity checks to ensure that all operations on valid inputs produce valid outputs. For some reason, these seem to be missing from IntMap and IntSet tests. That makes me nervous!

So far, I found no reason to worry: https://github.com/antalsz/hs-to-coq/blob/master/examples/containers/theories/IntSetProofs.v :-)

treeowl:
Very interesting! I wonder how hard it would be to ensure such proofs stay
up to date. I know proof maintenance tends to be challenging.

nomeata:
That is a very important and hard question, that I will avoid answering for now…

Issue #404, opened 2017-02-14
Improve the documentation on how IntSet is represented


wrengr:
The current comments/documentation on the internal representation of `IntSet` are poorly/confusingly phrased. This should be fixed to ease introducing future developers.

m-renaud:
Thoughts on this being a README in the Data/IntMap/ directory (in addition to better comments in code)? That may make it easier to format and add graphics to help with the explanation. I can put something together since I've been digging around in the code a little bit.

treeowl:
A README sounds okay, but there will need to be a strongly worded note in
the code reminding people to update it whenever the structure or invariants
change. Also, I think improving the comments remains a priority regardless.

PR #456, opened 2017-12-22
Add validity checks for IntSet and IntMap

m-renaud:
Codify the invariants outlined in comments and add property checks to verify that the internal structure of IntSet and IntMap are correct under initial construction, union, intersection, and difference.

This addresses https://github.com/haskell/containers/issues/304.


Data/IntMap/Internal.hs
```diff
@@ -353,6 +354,11 @@ data IntMap a = Bin {-# UNPACK #-} !Prefix
                     {-# UNPACK #-} !Mask
                     !(IntMap a)
                     !(IntMap a)
+-- Invariant: Nil is never found as a child of Bin.
+-- Invariant: Prefix is the common high-order bits that all elements share to
+--            the left of the Mask bit.
+-- Invariant: In Bin prefix mask left right, left consists of the elements that
+--            don't have the mask bit set; right is all the elements that do.
```
treeowl: Perhaps explain what a `mask` looks like?

Data/IntMap/Internal.hs
```diff
@@ -3396,3 +3402,50 @@ node = "+--"
 withBar, withEmpty :: [String] -> [String]
 withBar bars   = "|  ":bars
 withEmpty bars = "   ":bars
+
```
treeowl: Should these assertions be here or under `tests`? I'm asking because I don't know.

Data/IntSet/Internal.hs
```diff
@@ -250,8 +251,8 @@ data IntSet = Bin {-# UNPACK #-} !Prefix {-# UNPACK #-} !Mask !IntSet !IntSet
 -- Invariant: In Bin prefix mask left right, left consists of the elements that
 --            don't have the mask bit set; right is all the elements that do.
             | Tip {-# UNPACK #-} !Prefix {-# UNPACK #-} !BitMap
--- Invariant: The Prefix is zero for all but the last 5 (on 32 bit arches) or 6
---            bits (on 64 bit arches). The values of the map represented by a tip
+-- Invariant: The Prefix is zero for the last 5 (on 32 bit arches) or 6 bits
+--            (on 64 bit arches). The values of the set represented by a tip
```
treeowl: Was this comment just flat-out wrong before? That's confusing and disturbing!

m-renaud:
Note to reviewers:

1. I'm not sure if there's a style guide for the package that I've violated or not. I tried to format things like the surrounding code, feel free to nit pick it though :)

2. One thing that struck me as strange is that some of the invariants that held for the IntSet relating to properties of the mask and prefix did not hold for the IntMap implementation (or I may have made a mistake porting them over). Specifically the `maskPowerOfTwo` property did not hold for `Mask` in the `IntMap`s `Bin` constructor. Also, the `tipsValid` check didn't appear to have a passing analogue.

Data/IntSet/Internal.hs
m-renaud: It seems to have been. It took me a little bit to realize that I was implementing the incorrect check :P I went and read the implementation closer and figured that this is _I think_ what it meant to say.

m-renaud: And the test that codifies this interpretation passes, sooo...

Data/IntMap/Internal.hs
m-renaud: I was following structure that Data.Set used[1], but I'm open to moving them somewhere else; perhaps in a follow-up PR though.

[1] https://github.com/haskell/containers/blob/master/Data/Set/Internal.hs#L1758

m-renaud: Looking through the implementations I don't think there's anything that requires them to be defined here since the internals of the types are exposed by the `Internal` module. I think I would prefer to do a mass move across all the data structures in a follow-up though.

treeowl:
Sounds good.

m-renaud:
> Great work so far!

Thanks! It's been fun ;)

>  I'd really like a validity check added to the tests for each operation that produces a set/map

Added a bunch more validity checks, see commit for details. Let me know if there's any others you'd like me to add :)

> Perhaps explain what a mask looks like?

I've added a short description given what I understand about the implementation, it's a little hand wavey though. #404 would definitely have made this easier ;P

m-renaud:
The CI failures appear to be due to configuration failing and not due to this change.

treeowl:
Ugh. I don't want to hear that! Have you had a chance to look at what
invariants you're missing in IntMap?

m-renaud:
> Ugh. I don't want to hear that!

:(

```
The command "cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit'" failed and exited with 1 during .
```

> Have you had a chance to look at what invariants you're missing in IntMap?

Not closely, the implementation is slightly different (wrt keys) than the IntSet implementation. I'll have to take another pass through but I didn't _see_ anything that the code was trying to enforce, there also aren't any comments though so I could very well be missing some.

m-renaud:
Went through the IntMap code again and that looks like all the invariants. `Tip` is much simpler in the IntMap case since it only ever holds a single (key, value) pair. Let me know if there are any other checks you'd like me to add. If I see anything I missed when making further changes I'll send another PR.

m-renaud:
Ah gotcha, no worries, that makes more sense than what I did :)

m-renaud:
> We also almost certainly want to make valid return a Property

Absolutely, although having something return a `Property` living in the `Internal` module seems weird, which gives more credence to the idea that the `valid` function maybe shouldn't live where it is today.

`tests/` is one option, or we could add a new `Data.IntMap.TestUtil` in case we decide to expose the internals as a separate package at some point in the future (but its probably best to go with tests/ since there's no concrete plan to export the internals separately).

So, there's two ways to go right now:
1. Change `valid` to return a `Property` (introducing testing code into the actual implementation).
2. Change it to a `Property` in the follow-up PR when I move all the `valid` checks from the various containers to the `tests/` directory.

I'm personally leaning towards 2, WDYT?

treeowl:
I'm leaning towards introducing the validity checks for IntMap where we
actually want them, and adding a follow-up to do the same for the rest.

On Dec 23, 2017 2:15 PM, "Matt Renaud" <notifications@github.com> wrote:

We also almost certainly want to make valid return a Property

Absolutely, although having something return a Property living in the
Internal module seems weird, which gives more credence to the idea that the
valid function maybe shouldn't live where it is today.

tests/ is one option, or we could add a new Data.IntMap.TestUtil in case we
decide to expose the internals as a separate package at some point in the
future (but its probably best to go with tests/ since there's no concrete
plan to export the internals separately).

So, there's two ways to go right now:

   1. Change valid to return a Property (introducing testing code into the
   actual implementation).
   2. Change it to a Property in the follow-up PR when I move all the valid
   checks from the various containers to the tests/ directory.

I'm personally leaning towards 2, WDYT?

—
You are receiving this because you commented.

<https://github.com/haskell/containers/pull/456#issuecomment-353743350>,
or mute
the thread

m-renaud:
> I'm leaning towards introducing the validity checks for IntMap where we
actually want them

Giving it another thought I completely agree, I'll go with that.

Data/IntMap/Internal.hs
m-renaud: Done.

m-renaud:
Moved the `valid` function to a new module in the tests/ directory. Is this an OK way to structure it?

treeowl:
I think that's generally okay. I don't know that we really need a separate module, but it shouldn't hurt. However, it shouldn't be a top-level name like that. One option is to add a `Data.IntMap.Debug` or some such, to include both the validity tests and tree-showing functions.

treeowl:
Er... I guess you're keeping the thing non-exposed from the library? If so, I guess it doesn't matter.

m-renaud:
Yeah, my thought was that if we ever decide to move the internals into their own package to be used by other packages then we can export the validity checks, but since only this package should be messing around with the internals I just created a new module in the tests/ directory. Let me know if that's alright with you or if you'd like me to move them somewhere else.

treeowl:
Sounds fine for now.

m-renaud:
SGTM. Unless you have any other concerns I think this is ready to merge.

tests/IntMapValidity.hs
```diff
@@ -0,0 +1,52 @@
...
+
+{--------------------------------------------------------------------
+  Assertions
+--------------------------------------------------------------------}
+-- | Returns true iff the internal structure of the IntMap is valid.
+valid :: IntMap a -> Property
+valid t
+  = property (nilNeverChildOfBin t) .&&.
+    property (commonPrefix t) .&&.
+    property (maskRespected t)
```
treeowl: I think we probably want to use `counterexample` rather than `property` here, to label each type of failure. I wonder if we should push the `Property` business into `commonPrefix` and `maskRespected` so we can see where in the tree the failure lies. That probably shouldn't be necessary for `nilNeverChildOfBin`, since that will be visually obvious.

m-renaud: Switched to using counterexample for each of these. It would take more work to push `Property` into the other functions and be done in such a way that we would get actionable information from it, so if its alright with you I'd like to hold off on that for now.

tests/IntSetValidity.hs
```diff
@@ -0,0 +1,86 @@
...
+
+-- Invariant: The Mask is a power of 2.  It is the largest bit position at which
+--            two elements of the set differ.
+maskPowerOfTwo :: IntSet -> Bool
+maskPowerOfTwo t =
+  case t of
+    Nil -> True
+    Tip _ _ -> True
+    Bin _ m l r ->
+      (m `mod` 2 == 0) && maskPowerOfTwo l && maskPowerOfTwo r
```
nomeata: I am a bit late to the party, but

    m `mod` 2 == 0

is not a way to test  that a mask is a power of 2. Alternative: `popcount m == 1`.

treeowl:
Not late at all. Thanks!

On Jan 17, 2018 10:30 PM, "Joachim Breitner" <notifications@github.com>
wrote:

*@nomeata* commented on this pull request.
------------------------------

In tests/IntSetValidity.hs
<https://github.com/haskell/containers/pull/456#discussion_r162241903>:

> +  where
+    noNilInSet t' =
+      case t' of
+        Nil -> False
+        Tip _ _ -> True
+        Bin _ _ l' r' -> noNilInSet l' && noNilInSet r'
+
+-- Invariant: The Mask is a power of 2.  It is the largest bit
position at which
+--            two elements of the set differ.
+maskPowerOfTwo :: IntSet -> Bool
+maskPowerOfTwo t =
+  case t of
+    Nil -> True
+    Tip _ _ -> True
+    Bin _ m l r ->
+      (m `mod` 2 == 0) && maskPowerOfTwo l && maskPowerOfTwo r

I am a bit late to the party, but

m `mod` 2 == 0

is not a way to test that a mask is a power of 2. Alternative: popcount m
== 1.

—
You are receiving this because you commented.

<https://github.com/haskell/containers/pull/456#pullrequestreview-89664243>,
or mute the thread

Issue #347, opened 2016-10-16
Possible Bug


giovannirescia:
Hi there!

I am working in a code that's not mine, so I can't understand it very well and therefore explain what I am doing (or trying to do). But the error that I get is the next:

`"index out of bounds\nCallStack (from HasCallStack):\n  error, called at ./Data/Sequence/Base.hs:1791:
19 in containers-0.5.8.1-GB7XRscggFnIkqZLCK9AcT:Data.Sequence.Base"`

I do not know a lot of Haskell also, so I can't interpret that message. Does anybody can help me to fix it or understand it?

Thanks a lot!
Giovanni

treeowl:
This is not nearly enough information to diagnose any problem. It sounds like the code you're using is probably calling `Data.Sequence.index` with a negative index or one too large for the sequence being indexed. I'll leave this ticket open briefly, because there is a sort of bug--the error message should mention the `index` function. But for help debugging your project, you should probably head to the `#haskell` channel on FreeNode, the haskell-beginners mailing list on haskell.org, or StackOverflow.

giovannirescia:
Ok, thank you very much!

m-renaud:
Error message improved in #457 so I'm going to close this issue.

PR #457, opened 2017-12-22
Improve error message for `index`

m-renaud:
When the index is out of bounds mention which function is causing the error. We include the index `i` in the error message because it may be useful in debugging (-1 vs large number could narrow down where to look).

This was originally mentioned in https://github.com/haskell/containers/issues/347.


treeowl:
Please specify `Data.Sequence.index` for clarity. Also, do you think you could benchmark the impact of adding `HasCallStack =>` to `index`? I don't know if we can get away with that, but it would be awfully friendly.

m-renaud:
Updated with Data.Sequence prefix, will try out HasCallStack when I get home tonight if I have time. If you want you can merge this in now and I can start a new PR for the HasCallStack if the benchmarks look alright.

treeowl:
Thanks!

Issue #354, opened 2016-10-24
Document how to build benchmarks


osa1:
I'm a bit confused about benchmark targets in `containers.cabal` -- as far as I understand there's no way to build those targets, because some of the dependencies of dependencies (like `criterion`) actually depend on `containers`, and `containers` is one of the packages that are always system-wide installed because of `ghc`.

What am I missing here?

This may be worth documenting actually, because no matter what I tried (sandboxes, fresh ghc installations) I couldn't build any of the benchmark targets. (hence #353)

PR #461, opened 2017-12-26
Update CONTRIBUTING.md with benchmark instructions

m-renaud:
This includes instructions for building and running benchmarks with both cabal-install and stack.

This addresses #354 and #371.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,27 +4,34 @@ As mentioned in the [README](https://github.com/haskell/containers/blob/master/R
 
 For proposing API changes/enhancements, please follow the [guidelines outlined on the Haskell Wiki](https://wiki.haskell.org/Library_submissions#Guide_to_proposers). Especially note that all API changes/enhancements should be discussed on libraries@haskell.org mailing list.
 
-## Building and testing
+## Building, testing, and benchmarking
 
-Building and testing the containers package can be done using either `cabal-install` or `stack`.
+Building, testing, and benchmarking the containers package can be done using either `cabal-install` or `stack`.
 
 ### With cabal-install
+
+Minimum cabal version: 1.24
 
 ```
 cabal sandbox init
 cabal install --only-dependencies
-cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit'
-cabal configure -v2 --enable-tests
+cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit' 'criterion'
+cabal configure -v2 --enable-tests --enable-benchmarks
 cabal build
 cabal test
-```
+cabal bench
+``` 
+
 
 ### With [Stack](https://docs.haskellstack.org/en/stable/README/)
+
+Minimum stack version: 1.6.1
 
 ```
 stack init   # If you haven't previously initialized stack
 stack build
 stack test
+stack bench
 ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -4,27 +4,34 @@ As mentioned in the [README](https://github.com/haskell/containers/blob/master/R
 
 For proposing API changes/enhancements, please follow the [guidelines outlined on the Haskell Wiki](https://wiki.haskell.org/Library_submissions#Guide_to_proposers). Especially note that all API changes/enhancements should be discussed on libraries@haskell.org mailing list.
 
-## Building and testing
+## Building, testing, and benchmarking
 
-Building and testing the containers package can be done using either `cabal-install` or `stack`.
+Building, testing, and benchmarking the containers package can be done using either `cabal-install` or `stack`.
```
treeowl: I suggest you mention the fact that the procedures here need to be a bit weird. If you have a good understanding of why (I don't) that might be good to mention.

CONTRIBUTING.md
```diff
@@ -4,27 +4,34 @@ As mentioned in the [README](https://github.com/haskell/containers/blob/master/R
...
 
+Minimum cabal version: 1.24
+
 ```
 cabal sandbox init
 cabal install --only-dependencies
-cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit'
-cabal configure -v2 --enable-tests
+cabal install 'test-framework >= 0.3.3' 'test-framework-quickcheck2 >= 0.2.9' 'QuickCheck >= 2.4.0.1' 'ChasingBottoms' 'HUnit' 'test-framework-hunit' 'criterion'
+cabal configure -v2 --enable-tests --enable-benchmarks
```
treeowl: You mention several specific dependency versions here. These could easily go out of date. What should we do to make sure they stay in tune with the `.cabal` file?

m-renaud: It turns out these specific versions aren't necessary; I've removed them from the instructions.

CONTRIBUTING.md
m-renaud: TL;DR; `cabal configure` isn't able to solve the constraints (I haven't had time to investigate why yet). Here's the error message:

```
Resolving dependencies...
Warning: solver failed to find a solution:
Could not resolve dependencies:
rejecting: containers-0.5.10.2:!bench (constraint from config file, command
line flag, or user target requires opposite flag selection)
trying: containers-0.5.10.2:*bench
unknown package: random (dependency of containers-0.5.10.2:*bench)
Dependency tree exhaustively searched.
Trying configure anyway.
Configuring containers-0.5.10.2...
cabal: Encountered missing dependencies:
ChasingBottoms -any,
HUnit -any,
QuickCheck >=2.7.1,
criterion >=0.4.0 && <1.3,
random <1.2,
test-framework >=0.3.3,
test-framework-hunit -any,
test-framework-quickcheck2 >=0.2.9
```

I plan on investigating this at some point. I've updated the paragraph with a brief explanation of why we don't just use `cabal configure`.

treeowl:
I'm not convinced the versions are unnecessary. Is it grabbing the right
ones by accident?

CONTRIBUTING.md
m-renaud: Filed https://github.com/haskell/containers/issues/462.

m-renaud:
Alright, I think we both agree the instructions for tests are weird, but since this is a benchmark instruction PR I'd like to revert the change that removes the version number and punt on that issue. WDYT?

m-renaud:
@treeowl does this look good now?

treeowl:
Can you separate the dependencies needed for testing from the ones needed for benchmarking? Not everyone will always need both. Otherwise, looks good.

m-renaud:
Done.

PR #463, opened 2017-12-26
Add stack.yaml file

### LINKED ISSUE #458: Cannot use stack for lts > 7.24
The last Stackage LTS which included the ChasingBottoms package was 7.24, using any more recent LTS does not include this which causes an error on initialization.

```shell
$ stack init --resolver lts-10.0
...
Resolver 'lts-10.0' does not have all the packages to match your requirements.
    ChasingBottoms not found
        - containers requires -any
...
$ stack init --resolver lts-9.20
...
Selected resolver: lts-9.20
Resolver 'lts-9.20' does not have all the packages to match your requirements.
    ChasingBottoms not found
        - containers requires -any
...
$ stack init --resolver lts-7.24
...
Initialising configuration using resolver: lts-7.24
Total number of user packages considered: 1
Writing configuration to file: stack.yaml
All done.
```

I looked at the [hackage page for ChasingBottoms](https://hackage.haskell.org/package/ChasingBottoms) but there was no email address to contact the maintainer (Nils Anders Danielsson) and the code isn't hosted on GitHub (just a Darcs repo on a server) so there's no way to file an issue.

Getting in contact with the maintainer and seeing if they can get it incorporated into the Stackage LTS is probably the best way forward.

m-renaud:
Running `stack init` fails if you're using a resolver newer than lts-7.24 because ChasingBottoms has not been in Stackage LTS releases since then. Since Stack is a common workflow people use we should support the most recent Stackage LTS versions out of the box.

In the stack.yaml file we add ChasingBottoms-1.3.1.3 as an extra-dep and default the resolver to 9.20 because lts-10.0 uses GHC 8.2 which is not supported all of the tools yet; ghc-mod is lacking support, for example. Once all tooling supports lts-10.0 we'll change the default resolver.

This also updates CONTRIBUTING.md, removing the `stack init` step because it is no longer needed.

This resolves #458.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,6 @@ cabal test
 ### With [Stack](https://docs.haskellstack.org/en/stable/README/)
 
 ```
-stack init   # If you haven't previously initialized stack
 stack build
 stack test
 ```
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,11 @@
+flags: {}
+packages:
+- '.'
+
+### Uncoment the resolver you want to use and re-run `stack build/test/bench`.
+# resolver: lts-10.0
+resolver: lts-9.20
+
+### ChasingBottoms is only in Stackage snapshots lts-7.24 and below.
+extra-deps:
+- ChasingBottoms-1.3.1.3


m-renaud:
Confirmed this will have no impact on Stackage snapshot generation: https://groups.google.com/forum/m/#!topic/haskell-stack/peD4SU0-S5c

treeowl:
Excellent! Go right ahead then!

Issue #464, opened 2017-12-26
Create external docs and tutorial for containers


m-renaud:
Containers is one of the core packages in the Haskell ecosystem and the APIs for many of the data types have grown very large over the years which makes it really difficult for someone ~new to Haskell~ to know where to start. There are no existing tutorials to be found that give a high level introduction to what `containers` provides, how to add it to your project, or what the "common" operations are from each of the modules.

The Haddocks are great if you know what you're looking for, but the module structure and the format of Haddocks in general isn't the best format for a tutorial. I propose we create a [ReadTheDocs](http://docs.readthedocs.io/en/latest/getting_started.html) page for containers. For each container this would provide an introduction of what the container is and when you would use it (following best practice advice), a short example of how to use it, and then a more detailed breakdown of the different types of operations with full examples and output. There would also be a subsection for each of the types with notes on performance (likely just liking out to https://github.com/haskell-perf). This would also be a good place to talk about other operations such as serialization (addressing #405) or common operations which aren't in the API proper but can easily be achieved by combining a couple of function calls that are.

The structure would be:
/doc/README.md - info about `containers`
/doc/Graph.md - info about Data.Graph
/doc/Set.md - info about Set.Strict and Set.Lazy, and reference to IntSet
/doc/IntSet.md - info about IntSet (could maybe just be a subsection of Set.md)
/doc/Map.md -info about Map.Strict and Map.Lazy, and reference to IntMap
/doc/IntMap.md - info about IntMap (could maybe just be a subsection of Map.md)
/doc/Sequence.md - info about Data.Sequence and when to use it
/doc/Tree.md - info about Data.Tree

It would be great if there was also corresponding docs (ideally in a similar place) for Data.HashSet and Data.HashMap from `unordered-containers` as well, I can look into doing that as well.

PR #466, opened 2017-12-29
Add docs for containers package.

m-renaud:
These docs are a beginner friendly introduction to the `containers` package. They include:

- what the `containers` package offers
- how to install and use the package
- info on related packages
- API docs for the most common functions in Set, Map, and Sequence

This code is configured so it can be served by https://readthedocs.org. For a preview of the rendered docs while reviewing see https://m-renaud-haskell-containers.readthedocs.io/en/docs/.

This at least in part addresses https://github.com/haskell/containers/issues/464.


treeowl:
I appreciate the effort you've put into this, but I'm sorry to say I'm not yet convinced we should do it.

I don't really see why users would look to `containers` for information about how to use the module or package system. That just seems out of scope.

The rest seems to duplicate a lot of the info found in the Haddocks, admittedly with a different focus. I'd really like to see how far we can get by enhancing and (harder) re-organizing the Haddock documentation before we turn to something out of line and thus harder to maintain. There's an awful lot of room for explanatory material that we're not using effectively; the `lens` package might be a good model of how to do better. We should also give more careful thought to the way functions should be grouped into sections, the order they should appear in those sections, and the order in which the sections should appear. Fortunately, it's fairly easy to do this without touching much more than the export lists and included text blocks, so the hard part is figuring out what to do.

m-renaud:
I completely agree that the Haddocks need to be improved (I'll have a PR for that soon as well), but they serve an entirely different purpose than the docs that I've put together here. There are two main workflows that the docs should support:

1. Someone who's new to Haskell (or someone out of practice) wants to do something more substantial, so one of the first things they ask is "Where are the standard data structures?" Inevitably someone points them to the Hackage page for containers or unordered-containers where they're confronted with the internal module structure of the package and when they eventually find their way into Data.Map.Strict they're presented with links to academic papers that underlie the implementation and 125(!!!) API functions. At this point they give up, go to Google and search for "containers tutorial"; myself, as well as many friends I have that have learned (or attempted to learn) Haskell followed this exact same workflow, for many packages. This is key to growing the Haskell community.

2. A somewhat experienced Haskeller is looking for a specific API function (say a specialized traversal or fold). They already know they're using Data.Map.Strict and where the Haddocks are, so they go there and jump to the "traversals" section which lists all the functions in the API with traversal like operations.

So, 2) is already somewhat addressed with the existing Haddocks. Yes, they could use some improvements (better grouping/ordering, explanation of semantics, copy-pasteable examples), but the core of it is there.

On the other hand, the story for 1) is a mess. There are two ways of going about fixing it: a) restructure the haddocks so that all the "simple" functions are near the top, and the more "complex" or "specialized" functions are at the bottom. b) create another set of docs aimed at being an introduction to what the package provides and a walkthrough of the core functionality for each of the data structures.

I'm going to posit that a) is the wrong approach for the following reasons:

- It hurts use case 2. Now if you want to see what traversals are provided you need to look in two places: the "simple" traversals and the "complex" traversals.

- Haddocks are the wrong format for this type of material:
  - Poor browsability: no good sidebar or search funcitonality
  - Doesn't allow you to intersperse additional paragraphs not tied to a given function or section.
  - Poor support for calling out notes/tips/warnings
  - Hard to edit and publish revisions
    - Need to find the repo
    - Can't previous changes for how they'll be rendered on hackage/can't view locally
    - Need to create a new package revision for the changes to go live
  - It's not very pretty

By having another set of docs, b), that specifically targets the first use case we can have the best of both world. Also, ReadTheDocs is already used by a number of other Haskell projects: [Cabal](https://readthedocs.org/projects/cabal/), [Stack](https://readthedocs.org/projects/stack/), [Servant](https://readthedocs.org/projects/haskell-servant/), and [Haddock](https://readthedocs.org/projects/haskell-haddock/) to name a few.

To address some of your more specific comments:

> I don't really see why users would look to containers for information about how to use the module or package system.

Sure, I would agree with you that the Haddocks are the wrong place to talk about the module or package system. These docs are specifically tailored to someone who's looking for an overview of the containers package and how to use it. This includes "how do I use this in my project? (a common question for beginners), and "what do I import?". Having a paragraph saying "just put `containers` in you build-depends" makes it really easy for a newcomer to use it (if they aren't super familiar with Cabal). The "import everything qualified" is something already suggested in the Haddocks, this just makes it a bit more prominent.

> out of line and thus harder to maintain

I think this is a straw man. The docs are checked into source control alongside the code, are easier to publish changes (just commit the changes), and the docs even have a direct link to GitHub where you can edit them in place. Also, the core/simple API for these containers isn't going to change very much over the years, so I don't think keeping this up to date with the haddocks is an issue.

One of the biggest complaints that Haskell users have is [libraries are poorly documented](http://taylor.fausak.me/2017/11/15/2017-state-of-haskell-survey-results/#question-36), and I think this can help improve that, for `containers` at least.

I'm also open (and think it would be a good idea) to getting feedback from the community on whether or not they think this would be beneficial. A reddit and twitter post may be a good way to go.

WDYT?

treeowl:
I don't see any links from the new documentation to the Haddocks for individual functions. It seems to me that all of them should be linked!

treeowl:
By the way, Edward Yang tells me that you can put `[ci skip]` in the commit message to tell Travis not to bother.

m-renaud:
> I don't see any links from the new documentation to the Haddocks

Agreed, I was trying to find a nice way to do it in restructuredtext, and it seems like I've found one. You can define a custom role so you can do stuff like:
```
:seq:`empty`
```
and it will generate a link to the `empty` function in `Data.Sequence`.

You can see the latest docs at https://m-renaud-haskell-containers.readthedocs.io/en/docs/.

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+      [] -> "empty list"
+      (x:xs) -> "first:" ++ show x ++ " rest:" ++ show xs
+    > "first:1 rest:[2,3]"
+
+
+We can do the same thing with sequences, but first we'll need to enable the
+``PatternSynonyms`` language extension. At the top of your Haskell source file
+above the ``module`` declaration add::
+
+   {-# LANGUAGE PatternSynonyms #-}
```
treeowl: The language extension should not be required to *use* the patterns with recent GHC versions. Just use

```haskell
import Data.Sequence (Seq (..))
```
or
```haskell
import Data.Sequence (Seq (Empty, (:<|), (:|>)))
```

rather than the older (and `PatternSynonyms`-only) syntax

```haskell
import Data.Sequence (pattern (:<), pattern (:>), pattern Empty)
```

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+
+Now that we have that out of the way, lets pattern match on a sequence!
```
treeowl: s/lets/let's

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+
+::
+
+    case Seq.fromList [1, 2, 3] of
+      Empty -> "empty sequence"
+      x :<| xs -> "first:" ++ x ++ " rest:" ++ show xs
+    > "first:1 rest:fromList [2,3]"
+
+.. NOTE:: You can't copy/paste this into GHCi because its multiple lines.
```
treeowl: s/its/it's

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+
+You can also take an element off the end::
+
+    case Seq.fromList [1, 2, 3] of
+      Empty -> "empty sequence"
+      xs :|> x -> "last element:" ++ x
```
treeowl: s/`++x`/`++ show x`

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+    xs !? n = ...
+
+:seq:`lookup` returns the element at the position ``n``, ``Nothing`` if the
+index is out of bounds. `!?
+<https://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Sequence.html#v:-33--63->`_
+is simply a flipped version of ``lookup``.
+
+.. NOTE::
+   You may need to import ``!?`` qualified if you're using a ``Map`` in the same
+   file because they both export the operator.
```
treeowl: `Map`, `IntMap`, and `Vector`, at least. `containers` stole it from `vector`; I don't know if they had it first.

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+
+::
+
+    Seq.index :: Seq a -> Int -> a
+    Seq.index xs n = ...
+
+:seq:`index` returns the element at the given position, throws a runtime
```
treeowl: s/, throws/; it throws. And why show `error` as code? That looks weird.

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+``error`` if the index is out of bounds.
+
+.. TIP::
+   Use ``lookup``/``!?`` whenever you can and explicitly deal with the
+   ``Nothing`` case.
```
treeowl: Examples of actually using `!?` might be nice.

```haskell
do
  let res = xs !? i
  case res of
    Nothing -> print "Whoops! I couldn't find your root vegetable!"
    Just x -> print "The " ++ show x ++ " potato was the best!"
```
```haskell
fromMaybe (error "M.myFunction internal error: index out of range") (xs !? i)
```

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+
+    let myMapFunc index val = index * val
+
+    Seq.mapWithIndex myMapFunc (Seq.fromList [1, 2, 3])
+    -- = fromList [0*1, 1*2, 2*3]
+    > fromList [0,2,6]
+
+
+Delete an element
+"""""""""""""""""
```
treeowl: It seems to me that this should be adjacent to `insertAt`, but maybe that's wrong?

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+Inserting an element
+""""""""""""""""""""
+
+::
+
+    Seq.insertAt :: Int -> a -> Seq a -> Seq a
+    Seq.insertAt i x xs = ...
+
+:seq:`insertAt` inserts ``x`` into ``xs`` at the index ``i``, shifting the rest
+of the sequence over.
```
treeowl: Note that if the index is out of range, the element will be added at the beginning or end of the sequence as appropriate. `insertAt (-10) 'a' (fromList "hi") = fromList "ahi"`, and `insertAt 10 'p' (fromList "hi") = fromList "hip"`.

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+    Seq.chunksOf :: Int -> Seq a -> Seq (Seq a)
+    Seq.chunksOf k xs = ...
+
+:seq:`chunksOf` splits the sequence ``xs`` into chunks of size ``k``. If the
+length of the sequence is not evenly divisible by ``k`` then the last chunk will
+have less than ``k`` elements.
+
+.. WARNING::
+   ``k`` can only be ``0`` when the sequence is empty, otherwise a runtime
+   ``error`` is thrown.
```
treeowl: Again, `error` looks funny to me.

docs/sequence.rst
```diff
@@ -0,0 +1,647 @@
...
+
+::
+
+    foldr :: (a -> b -> b) -> b -> Seq a -> b
+    foldr f init xs = ...
+
+    Seq.foldrWithIndex :: (Int -> a -> b -> b) -> b -> Seq a -> b
+    Seq.foldrWithIndex f init xs = ...
+
+:seq:`foldr` collapses the sequence into a summary value by applying repeatedly
```
treeowl: "applying repeatedly applying". Pick one and only one applying.

docs/sequence.rst
m-renaud: TIL, thanks!

docs/sequence.rst
m-renaud: Done.

docs/sequence.rst
m-renaud: Done.

docs/sequence.rst
m-renaud: Done.

docs/sequence.rst
m-renaud: Added.

docs/sequence.rst
m-renaud: Done, and I have no idea.

docs/sequence.rst
m-renaud: Added an example and some words.

docs/sequence.rst
m-renaud: Added words and examples demonstrating that.

docs/sequence.rst
m-renaud: Agreed, moved.

docs/sequence.rst
m-renaud: Removed formatting.

docs/sequence.rst
m-renaud: :)

m-renaud:
I think I've addressed all your comments. I added a TIP near the top mentioning the OverloadedLists extension but chose to use `fromList` throughout for clarity.

docs/sequence.rst
```diff
@@ -0,0 +1,672 @@
...
+fold using ``Seq``'s `Foldable
+<https://wiki.haskell.org/Typeclassopedia#Foldable>`_ instance.
+
+::
+
+    foldr (:) [] (Seq.fromList ["base", "containers", "QuickCheck"])
+    > ["base","containers","QuickCheck"]
+
+
+Pattern Matching
```
treeowl: The patterns are bidirectional. So for appropriate `containers` *and GHC* versions, these patterns can completely replace `<|`, `|>`, `empty`, `null` (except where that's actually desired), `viewl`, and `viewr`. I tend to think that in a tutorial like this, the patterns should be presented as the primary way to perform these operations. Only people concerned about Haskell 2010 compliance or older versions of GHC need to deal with the rest. The only annoying bit is that `:<|` and `:|>` are a bit verbose for historical reasons. If we were building from scratch, we might've used `:<` and `:>`, but those were taken for inferior view constructors.

m-renaud: Is it reasonable to assume that everyone would be on the "appropriate `containers` and GHC versions"? I'm assuming they would need containers >= 0.5.10 and ghc >= 8.2 which not everyone is on at this point (especially if they're using a system GHC).

If they need something that new I would be hesitant to put these as the primary way of doing these operations. I don't even use 8.2 regularly right now because ghc-mod still doesn't support 8.2 so editor tooling is kinda borked.

How would you feel about revisiting this in a few months when all the 8.2 kinks are worked out?

treeowl: GHC 8.0 will do, unless you want to run `-Wall` or `-Wincomplete-patterns`. It's up to you; I just think the pattern synonyms make sequences easier to work with and reduce the mental clutter of silly view functions.

docs/sequence.rst
```diff
@@ -0,0 +1,672 @@
...
+    show <$> nums
+    > fromList ["0","1","2"]
+
+
+    -- Fold a sequence into a summary value.
+    foldr (+) 0 (Seq.fromList [0, 1, 2])
+    > 3
+
+.. TIP:: You can use the `OverloadedLists
+	 <https://ghc.haskell.org/trac/ghc/wiki/OverloadedLists>`_ extension so
```
treeowl: Why are you linking to the GHC developer's wiki rather than the documentation?

docs/sequence.rst
```diff
@@ -0,0 +1,672 @@
...
+------------------
+
+When using ``Sequence`` in a Haskell source file you should always use a
+``qualified`` import becasue it exports names that clash with the standard
+Prelude (you can import the type constructor and some operators on their own
+though!).
+
+::
+
+    import Data.Sequence (Seq(..), (<|), (|>), (><))
```
treeowl: `Seq (..)` is only a reasonable thing to do when using a version with pattern synonyms; otherwise it'll produce a warning. Not sure how you want to deal with that.

docs/sequence.rst
treeowl: The real documentation is here: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#overloaded-lists

docs/sequence.rst
```diff
@@ -0,0 +1,672 @@
...
+- ``xs |> x`` places the element ``x`` at the end of the sequence ``xs``.
+
+- ``l >< r`` combines the two sequences ``l`` and ``r`` together.
+
+
+Create a list from a sequence
+"""""""""""""""""""""""""""""
+
+::
+
```
treeowl: There is no need to write this out by hand using `foldr`. Just use `Data.Foldable.toList`!

docs/sequence.rst
```diff
@@ -0,0 +1,672 @@
...
+    foldr (:) [] :: Seq a -> [a]
```
treeowl: There is no need to write this out by hand using `foldr`. Just use `Data.Foldable.toList`!

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.difference (Map.fromList[(1,"one"), (2,"two")]) (Map.fromList [(1,"uno")])
+    > fromList [(2,"two")]
+
+Subset (submap)
+"""""""""""""""
+
+::
+
+    Map.isSubmapOf :: (Eq a, Ord k) => Map k v -> Map k v -> Bool
+    Map.isSubmapOf l r = ...
```
treeowl: I kind of doubt we should include this. It's a fairly unusual thing to want to do.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.empty `Map.isSubmapOf` (Map.fromList [(1,"one"), (2,"two")])
+    > True
+
+    (Map.singleton 1 "uno") `Map.isSubmapOf` (Map.fromList [(1,"one"), (2,"two")])
+    > True
+
+
+Typeclass Instances
+-------------------
+
```
treeowl: Listing these without explaining their importance, relevance, or specific meaning for maps seems kind of silly to me. I think it could be helpful to explain how to use the `Foldable` and `Traversable` instances, but most of the rest of this is pretty boring.

docs/set.rst
```diff
@@ -0,0 +1,487 @@
...
+Typeclass Instances
+-------------------
+
+``Set`` is an instance of a number of common typeclasses, for the full list see
+the `docs
+<https://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Set.html#t:Set>`_.
+
+.. NOTE::
+   Some constraints have been left out for brevity, and the types given below
+   are speciliazed to ``Set``; the true types are more general.
```
treeowl: specialized

docs/set.rst
```diff
@@ -0,0 +1,487 @@
...
+  equality check: ``(==) :: Eq a => Set a -> Set a -> Bool``
+- `Ord
+  <https://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#t:Ord>`_ -
+  comparison: ``(<) :: Ord a => Set a -> Set a -> Bool``
+- `Foldable <https://wiki.haskell.org/Typeclassopedia#Foldable>`_ - collapse
+  into summary value: ``foldr :: (a -> b -> b) -> b -> Set a -> b``
+- `Semigroup <https://wiki.haskell.org/Typeclassopedia#Semigroup>`_ - combine
+  two things together (`union`_): ``(<>) :: Set a -> Set a -> Seq a``
+- `Monoid <https://wiki.haskell.org/Typeclassopedia#Monoid>`_  - a semigroup
+  with an identity element: ``mempty :: Seq a``
```
treeowl: I don't think explaining what the classes are is too helpful. Explaining how they apply to `Set` and how they're used with sets might be useful.

docs/set.rst
```diff
@@ -0,0 +1,487 @@
...
+
+Filtering elements from a set
+"""""""""""""""""""""""""""""
+
+::
+
+    Set.filter :: (a -> Bool) -> Set a -> Set a
+    Set.filter predicate s = ...
+
+:set:`filter` removes elements from the set ``s`` that **do not match** the
```
treeowl: Be positive!   `filter p s` produces a set consisting of the elements of `s` for which `p` returns `True`.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.delete 1 (Map.fromList [(1,"one"),(2,"two"),(3,"three")])
+    > fromList [(2,"two"),(3,"three")]
+
+Filtering map entries
+"""""""""""""""""""""
+
+::
+
+    Map.filter :: (v -> Bool) -> Map k v -> Map k v
+    Map.filter predicate m = ...
```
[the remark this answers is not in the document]
treeowl: Yes, I'm quite convinced of that. Wanting to filter a map by just *values* is likely to be very rare. I wouldn't even mention `filter`.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+Union
+"""""
+
+::
+
+    Map.union :: Ord k => Map k v -> Map k v -> Map k v
+    Map.union l r = ...
+
+:map:`union` returns a map containing all entries that are keyed in either of
+the two map. If the same key appears in both maps, the value from the left map
```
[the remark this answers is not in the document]
treeowl: Also, "are keyed in" is confusing.

docs/sequence.rst
```diff
@@ -0,0 +1,672 @@
...
+    Seq.length xs = ...
+
+:seq:`length` returns the length of the sequence ``xs``.
+
+::
+
+    Seq.length Seq.empty
+    > 0
+
+    Seq.size (Seq.fromList [1, 2, 3])
```
treeowl: There is no such function!

docs/sequence.rst
m-renaud: Probably because its neither on the first or second page when googling for "overloadedlists haskell". I didn't even know that page existed :worried:.

*/me adds this to the list of examples of "impossible to find documentation".*

docs/sequence.rst
m-renaud: Oh, much better!

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+
+Create an empty map
+"""""""""""""""""""
+
+::
+
+    Map.empty :: Map k v
+    Map.empty = ...
+
+:map:`empty` creates a map with zero entries.
```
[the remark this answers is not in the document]
m-renaud: SGTM

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+
+    Map.elems :: Map k v -> [v]
+    Map.elems m = ...
+
+:map:`elems` returns a list of values held in the map ``m``.
+
+::
+
+    Map.toAscList, Map.toList, Map.assocs :: Map k v -> [(k, v)]
+    Map.toAscList m = ...
```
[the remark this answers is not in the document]
m-renaud: Done.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    > 3
+
+
+Check if a key is present in the map (member)
+"""""""""""""""""""""""""""""""""""""""""""""
+
+::
+
+    Map.member :: Ord k => k -> Map k v -> Bool
+    Map.member key m = ...
```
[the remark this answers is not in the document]
m-renaud: I don't think so, removed it.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.!? m key = ...
+
+:map:`lookup` the value corresponding to the given ``key``, returns ``Nothing``
+if the key is not present; the ``!?`` operator (*since 0.5.10*) is a flipped
+version of ``lookup`` and can often be imported unqualified.
+
+::
+
+    Map.findWithDefault :: Ord k => v -> k -> Map k v -> v
+    Map.findWithDefault defaultValue key m = ...
```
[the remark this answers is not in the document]
m-renaud: Rewrote section.

docs/set.rst
m-renaud: Much clearer :)

docs/set.rst
m-renaud: Done.

docs/sequence.rst
m-renaud: Oops, typo! I probably got it confused with the `size` in Set and Map. Speaking of which, why aren't they all called size?

docs/map.rst
m-renaud: Agreed, removed.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+Difference
+""""""""""
+
+::
+
+    Map.difference :: Ord k => Map k v -> Map k v -> Map k v
+    Map.difference l r = ...
+
+:map:`difference` returns a map containing all entries that are keyed in the
+``l`` map but not the ``r`` map (`set difference/relative compliment
```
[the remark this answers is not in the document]
m-renaud: Done.

docs/map.rst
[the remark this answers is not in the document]
m-renaud: Changed, and small rewording.

docs/map.rst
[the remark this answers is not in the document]
m-renaud: Done.

m-renaud:
Thanks for the careful review :) Addressed the easy comments, I'll get around to the others when I have some free time.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.union newNums nums
+    > fromList [(1,"one"),(2,"two"),(3,"new three"),(4,"new four")]
+
+
+Importing Map and IntMap
+------------------------
+
+When using ``Map`` or ``IntMap`` in a Haskell source file you should always use
+a ``qualified`` import because these modules export names that clash with the
+standard Prelude (you can import the type constructor on its own though!).
```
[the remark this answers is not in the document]
m-renaud: In the example below the `Map` type constructor is imported unqualified, `lookup` conflicts with the list version from Prelude so I'm not sure we should recommend that. That leaves `lookup`, which we could recommend importing unqualified, but there's also the `!?` operator which we recommend importing unqualified when we present it.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.fromList [(1,"uno"), (1,"one"), (2,"two"), (3,"three")]
+    > fromList [(1,"one"),(2,"two"),(3,"three")]
+
+There's another incredibly useful function for constructing a map from a list::
+
+    Map.fromListWith :: Ord k => (a -> a -> a) -> [(k, a)] -> Map.Map k a
+    Map.fromListWith f xs = ...
+
+:map:`fromListWith` allows you to build a map from a list ``xs`` with repeated
+keys, where ``f`` is used to "combine" (or "choose") values with the same key.
```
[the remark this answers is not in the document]
m-renaud: Done.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.adjust :: Ord k => (v -> v) -> k -> Map k v -> Map k v
+    Map.adjust f key m = ...
+
+:map:`abjust` applies the value transformation function ``f`` to the entry with
+given ``key``, if no entry for that key exists then the map is left unchanged.
+
+::
+
+    Map.update :: Ord k => (v -> Maybe v) -> k -> Map k v -> Map k v
+    Map.update f key m = ...
```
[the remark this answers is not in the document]
m-renaud: Good call.

docs/map.rst
m-renaud: I think you're right, I don't think these instances really add anything so I've removed the section from all the pages.

docs/set.rst
m-renaud: Removed the typeclasses section. I think it'll require some more thought to figure out how to present them best which I'd like to do as a follow-up.

docs/sequence.rst
m-renaud: Removed the `(..)` part, the pattern matching section has an example for how to import the patterns.

docs/map.rst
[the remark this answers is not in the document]
treeowl: Regardless of whether you import `lookup` qualified, you should *always* `import Prelude hiding (lookup)` if you're using `Map`. Otherwise you may find yourself getting very confusing type errors if you leave off the qualifier by mistake. It looks like you meant to be talking about two different functions in the comment above (not just `lookup`) but things got mangled.

docs/map.rst
treeowl: Also, you probably don't want to use the `Foldable` or `Traversable` instances for maps at all; if you want to fold or traverse them you probably should use `fold*WithKey` or `traverseWithKey`. The `WithKey` folds and traversals would probably be worth discussing.

docs/map.rst
```diff
@@ -0,0 +1,629 @@
...
+    Map.union (Map.fromList [(1, "uno")]) (Map.fromList [(1,"one"),(2,"two")])
+    > fromList [(1,"uno"),(2,"two")]
+
+Intersection
+""""""""""""
+
+::
+
+    Map.intersection :: Ord k => Map k v -> Map k v -> Map k v
+    Map.intersection l r = ...
```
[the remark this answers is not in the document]
m-renaud: Done.

m-renaud:
I think I've implemented all your suggestions, let me know if one slipped through the cracks. I've also added the `:haddock:` autolinker so there's no reference to hackage anywhere in the docs, the location of the haddocks to link to can be specified when the docs are being built (see CL comments for example invocations).

I plan to clean up the code more and add it as an extension eventually, but I'd prefer to do that as a follow up if you're alright with that.

docs/map.rst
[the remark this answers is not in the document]
m-renaud: Good call, updated section.

The fact that such an easy mistake is a bad type error is really unfortunate, I would really like if GHC had error messages like Elm where it looks at similar functions and known common errors and suggests fixes.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+
+
+
+Create a list from a map
+""""""""""""""""""""""""
+
+::
+
+    Map.elems :: Map k v -> [v]
+    Map.elems m = ...
```
treeowl: This function isn't important enough, IMO.

docs/map.rst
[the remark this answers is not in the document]
treeowl: I'd mention that `Foldable.toList` is rarely useful for maps, but that its functionality is also available from `elems`. Or something like that. The goal is to avoid confusion, not to steer people toward things that aren't too relevant.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+Short Example
+-------------
+
+The following GHCi session shows some of the basic map functionality::
+
+    import qualified Data.Map.Strict as Map
+
+    let nums = Map.fromList [(1,"one"), (2,"two"), (3,"three")]
+
+    -- Check if we know the english words for 1 and 4
```
treeowl: s/english/English

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+    Map.member nums 1
+    > True
```
treeowl: Since I don't think `member` is important enough to discuss, I don't really think you should use it in examples either. Just skip it.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+
+    Map.member nums 4
+    > False
+
+
+    -- Get the english word for the number 3 and 4.
```
treeowl: s/english/English

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+    -- Remove the entry for 1 from our original map.
+    let fewerNums = Map.delete 1 nums
+
+    Map.toAscList fewerNums
+    > [(2,"two"),(3,"three")]
+
+
+    -- Create a new map and combine it with our original map.
+    -- Note that these operations are left-biased.
+    let newNums = Map.fromList [(3,"new three"), (4,"new four")]
```
treeowl: Specify that `union` (not "these operations") is left-biased. The other operation here is `fromList`, which is basically *right-biased*. By the way, no, that biasing thing is *not my fault*. And it would probably be good to explain what the "left-biased" thing means.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+	 will be converted automatically! The code here will continue to use
+	 ``fromList`` for clarity though.
+
+
+Importing Map and IntMap
+------------------------
+
+When using ``Map`` or ``IntMap`` in a Haskell source file you should always use
+a ``qualified`` import because these modules export names that clash with the
+standard Prelude (you can import the type constructor on its own though!). You
```
treeowl: How about something like

> You can always import any specific identifiers you want unqualified. Most of the time, that will include the type constructor (`Map`).

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+should also import Prelude and hide ``lookup`` because if you accidentally leave
```
treeowl: Import `Prelude`, I imagine.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+    import qualified Data.IntMap.Strict as IntMap
+
+
+Common API Functions
+--------------------
+
+.. TIP::
+   All of these functions that work for ``Map`` will also work for ``IntMap``,
+   which has the key type ``k`` specialized to ``Int``. Anywhere that you
+   see ``Map Int v`` you can replace it with ``IntMap v``.
```
treeowl: Add that using `IntMap v` instead of `Map Int v` will speed up most operations tremendously. The major exception is `size`, which is very fast (O(1)) for `Map` and very very slow (O(n)) for `IntMap`.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+
+.. NOTE::
+   ``fromList [some,map,entries]`` is how a ``Map`` is printed.
```
treeowl: I would use an example demonstrating that the list is made of *pairs*. Maybe

> A `Map` is printed as an association list preceded by `fromList`. For example, it might look like
> ```haskell
> fromList [(Key1, True), (Key2, False)]
> ```

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+Check if a map is empty
+"""""""""""""""""""""""
+
+::
+
+    Map.null :: Map k v -> Bool
+    Map.null m = ...
+
+:haddock_short:`Data.Map.Strict#null` returns ``True`` if the map ``m`` is
+empty, ``False`` otherwise.
```
treeowl: I would say "is empty and `False` otherwise".

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+
+::
+
+    Map.size Map.empty
+    > 0
+
+    Map.size (Map.fromList [(1,"one"), (2,"two"), (3,"three")])
+    > 3
+
+Lookup an entry in the map (lookup)
```
treeowl: s/Lookup an entry/Look up an entry

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+    Map.lookup :: Ord k => k -> Map k v -> Maybe v
+    Map.lookup key m = ...
+
+    Map.!? :: Ord k => Map k v -> k -> Maybe v
+    Map.!? m key = ...
+
+:haddock_short:`Data.Map.Strict#lookup` the value corresponding to the given
+``key``, returns ``Nothing`` if the key is not present; the ``!?`` operator
+(*since 0.5.10*) is a flipped version of ``lookup`` and can often be imported
+unqualified.
```
treeowl: `lookup` and `!?` can equally well be imported unqualified (once you hide the silly `lookup` from `Prelude`). I wouldn't necessarily bother mentioning that.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+
+::
+
+    Map.delete :: Ord k => k -> Map k v -> Map k v
+    Map.delete key m = ...
+
+:haddock_short:`Data.Map.Strict#delete` removes the entry with the specified
+``key`` from the map ``m``.  If the key doesn't exist it leaves the map
+unchanged. Remember, maps are immutable so if you delete an entry from a map you
+need to assign the new map to a new variable.
```
treeowl: The same warning would also apply to `insert`, `alter`, `adjust`, etc. I think it would be better to condense all such discussion. Near the top of this thing you introduce `Map` as an immutable data structure. I suggest you explain a bit about what that means there, and then probably never mention it again.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+    let removeElement _ = Nothing
+    Map.alter removeElement "key" (Map.fromList [("key", 0)])
+    > fromList []
+
+    let setValueToOne _ = Just 1
+    Map.alter setValueToOne "key" (Map.fromList [("key", 0)])
+    > fromList [("key",1)]
+
+    Map.alter setValueToOne "key" Map.empty
+    > fromList [("key",1)]
```
treeowl: Neither of these examples actually explain what really makes `alter` useful &mdash;you've just written likely-slower versions of `insert` and `delete`! You could also rewrite `adjust`, but there's not much point. `alter` gets useful if

* You don't know whether you want to change (or retain) or delete a key until you've seen what value it's mapped to, and/or

* You only want to insert a value if it's not already present.

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+::
+
+    Map.map (*10) (Map.fromList [("haskell", 45), ("idris", 15)])
+    > fromList [("haskell",450),("idris",150)]
+
+    -- Use the Functor instance for Map.
+    (*10) <$> Map.fromList [("haskell", 45), ("idris", 15)]
+    > fromList [("haskell",450),("idris",150)]
+
+There are several other more complex mapping functions available that let you
```
treeowl: I think both `mapWithKey` and `traverseWithKey` probably deserve to be in this tutorial. For `traverseWithKey`, you could give a simple example traversing in `IO`:

```haskell
-- | Ask the user how they want to schedule a bunch of tasks
-- that the boss has assigned certain priorities.
makeSchedule :: Map Task Priority -> IO (Map Task DateTime)
makeSchedule = traverseWithKey $ \task priority ->
  do
    putStrLn $ "The boss thinks " ++ show task ++ " has priority " ++ show priority ++
                      ". When do you want to do it?"
    readLn
```

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+
+.. _union:
+
+Union
+"""""
+
+::
+
+    Map.unionWith :: Ord k => (v -> v -> v) -> Map k v -> Map k v -> Map k v
+    Map.union f l r = ...
```
treeowl: s/union/unionWith

docs/map.rst
```diff
@@ -0,0 +1,602 @@
...
+    > fromList [(1,"ax"),(2,"y")]
+
+
+Intersection
+""""""""""""
+
+::
+
+    Map.intersectionWith :: Ord k => (v -> v -> v) -> Map k v -> Map k v -> Map k v
+    Map.intersection f l r = ...
```
treeowl: Again, s/intersection/intersectionWith

docs/set.rst
```diff
@@ -0,0 +1,471 @@
+Sets
+====
+
+:haddock:`unordered-containers`
+
+:haddock:`unordered-containers/Data.HashSet`
+
+:haddock:`unordered-containers/Data.HashSet#empty`
```
treeowl: These links really don't seem to belong here.

docs/set.rst
```diff
@@ -0,0 +1,471 @@
...
+	 will be converted automatically! The code here will continue to use
+	 ``fromList`` for clarity though.
+
+
+Importing Set and IntSet
+------------------------
+
+When using ``Set`` or ``IntSet`` in a Haskell source file you should always use
+a ``qualified`` import because these modules export names that clash with the
+standard Prelude (you can import the type constructor on its own though!):
```
treeowl: Also here, I would expand from just talking about the type constructor to any additional things you desire.

docs/set.rst
```diff
@@ -0,0 +1,471 @@
...
+    import qualified Data.IntSet as IntSet
+
+
+Common API Functions
+--------------------
+
+.. TIP::
+   All of these functions that work for ``Set`` will also work for ``IntSet``,
+   which has the element type ``a`` specialized to ``Int``. Anywhere that you
+   see ``Set Int`` you can replace it with ``IntSet``.
```
treeowl: Note here that almost all functions are much faster for `IntSet` than for `Set`. The only major exception is `size`, which is much much slower for `IntSet`.

docs/set.rst
```diff
@@ -0,0 +1,471 @@
...
+
+::
+
+    Set.size Set.empty
+    > 0
+
+    Set.size (Set.fromList [0, 2, 4, 6])
+    > 4
+
+Check if an element is in a set (member)
```
treeowl: This should float to the top of the querying section; it's the most important query by far.

docs/map.rst
treeowl: Also, please float `lookup` to the top of the query section. It's the most important query by far.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Removed.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: I like it.

docs/map.rst
m-renaud: Yup, done.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Removed.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Acknowledged.

docs/map.rst
m-renaud: Done.

docs/map.rst
m-renaud: Done.

docs/set.rst
m-renaud: Oops, these were here when I was testing the custom role and I forgot to remove them.

docs/set.rst
m-renaud: Oops, these were added when I was testing the autolinker and I forgot to remove them.

docs/set.rst
m-renaud: Done.

docs/set.rst
m-renaud: Done.

docs/set.rst
m-renaud: Done.

m-renaud:
Addressed most of the comments, the slightly more in depth ones I'll hopefully get to tomorrow.

treeowl:
Once the rest of these issues are ironed out, I think I'll be ready to merge it. I'd prefer to to do sooner rather than later. One thing I've noticed is that there are a number of places where your language or examples are clearer, more complete, or otherwise better than in the Haddocks. I'd really like to see some of those improvements integrated into the haddocks.

m-renaud:
> Once the rest of these issues are ironed out, I think I'll be ready to merge it.

Sounds good, I should be able to finish them tomorrow so it should be merged in by tomorrow night.

> I'd really like to see some of those improvements integrated into the haddocks

For sure, I've started going through the docs so I'll send out a follow-up PR with changed to the haddocks. I'd also like to discuss formatting the docs something like the following:
```
-- | One sentence summary.
--
-- More elaborate description,caveats, etc. if necessary.
--
-- Complexity: O(n)
--
-- Examples:
-- <examples>

and using that format throughout the package. That's a discussion we can have elsewhere though.

File: docs/map.rst (hunk 47 above)
Comment (m-renaud) [reply docs/map.rst]: Good call, I just copied these from the Haddocks, which means we should make the haddock examples better as well :) I'm trying to make everything in the examples copy/pasteable into a ghci session which doesn't work too well with case statements. How would you feel about something like this:

```
let addValueIfMissing mv = if isJust mv then mv else (Just 1)
```

An example where we want to look at the value would require a case statement, I think I'll just make a separate block and mention that it must be in a .hs source file.

```
doubleIfPositive :: Maybe Int -> Maybe Int
doubleIfPositive mv = case mv of
  Just v -> if v > 0 then (Just v*2) else (Just v)
  Nothing -> Just 1

Map.alter doubleIfPositive "key" (Map.fromList [("key", 1)])
```

Thoughts?

File: docs/map.rst (hunk 46 above)
Comment (m-renaud) [reply docs/map.rst]: Done. Made a similar change for the Set docs.

File: docs/map.rst (hunk 48 above)
Comment (m-renaud) [reply docs/map.rst]: Good idea!

File: docs/map.rst (hunk 26 above)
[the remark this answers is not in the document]
Comment (m-renaud) [reply docs/map.rst]: Thinking about this more I'm leaning towards just mentioning toAscList and excluding the others (as well as the note about toList). WDYT?

File: docs/map.rst (hunk 18 above)
Comment (m-renaud) [reply docs/map.rst]: Removed entirely.

File: docs/map.rst (hunk 47 above)
Comment (m-renaud) [reply docs/map.rst]: Updated the examples.

**Comment by m-renaud** [comment]
PTAL. I think I've addressed all your comments (please let me know if I missed anything). Also ignore ci, I forgot to add [ci skip] to the merge commit and just cancelled the jobs manually.

File: docs/.gitignore [hunk 55]
```diff
@@ -0,0 +1,2 @@
+_build/*
+_extensions/*.pyc
```
treeowl: Add a newline.

docs/map.rst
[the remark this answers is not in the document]
treeowl: I think the note about `toList` should be included, because it's a potential gotcha.

docs/.gitignore
m-renaud: Done.

docs/map.rst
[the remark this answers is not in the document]
m-renaud: Done.

m-renaud:
I think this is ready to be merged :)

treeowl:
The way you merged a bunch of other things in here is going to make merging this a little complicated (as I said, my git-fu is not very strong). Let me see what I can do.

treeowl:
Well, I've got everything merged on my end, but there's another issue. You've added a 3462-byte PNG for the Haskell logo. I don't want that in git! Is there an SVG or something we can use instead?

m-renaud:
Haha, oh yeah, good call :P Switched to an svg and used an all black logo (looks better against the blue background imho).

treeowl:
Merged in b72408b0aaa705a7d7a4f7d5ff98c116002c6e78.

PR #467, opened 2017-12-29
Add RULES for Data.IntMap.alterF

m-renaud:
Adds afterF rewrite RULES for the `Identity` and `Const` functor.

## Testing

- Adds unit tests for alterF for list, identity, and const
- Property test comparing the rewritten versions to versions operating on isomorphic structures

## Benchmark

```
benchmarking alter
time                 698.5 μs   (676.3 μs .. 726.1 μs)
                     0.994 R²   (0.991 R² .. 0.999 R²)
mean                 674.8 μs   (668.6 μs .. 686.6 μs)
std dev              28.63 μs   (14.53 μs .. 45.31 μs)
variance introduced by outliers: 34% (moderately inflated)

benchmarking alterF
time                 969.5 μs   (962.1 μs .. 977.3 μs)
mean                 956.7 μs   (952.8 μs .. 961.1 μs)
std dev              13.98 μs   (11.21 μs .. 17.72 μs)

benchmarking alterFIdentity
time                 660.7 μs   (654.5 μs .. 673.4 μs)
                     0.994 R²   (0.990 R² .. 0.997 R²)
mean                 674.1 μs   (660.4 μs .. 691.3 μs)
std dev              49.44 μs   (38.40 μs .. 57.64 μs)
variance introduced by outliers: 61% (severely inflated)

benchmarking alterFConst
time                 14.98 μs   (13.94 μs .. 16.02 μs)
                     0.976 R²   (0.972 R² .. 0.990 R²)
mean                 15.12 μs   (14.57 μs .. 15.84 μs)
std dev              2.104 μs   (1.813 μs .. 2.236 μs)
variance introduced by outliers: 92% (severely inflated)
```

This addresses #325.

### DIFF
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -986,20 +986,18 @@ alterF f k m = (<$> f mv) $ \fres ->
 -- TODO(m-renaud): Figure out if this should be marked INLINE or NOINLINE.
 -- It needs to be one or the other or else the specialization rule may not fire.
 {-# NOINLINE [1] alterF #-}
+#if MIN_VERSION_base(4,8,0)
 {-# RULES
 "Identity specialize alterF" forall (f :: Maybe a -> Identity (Maybe a)) k m.
-  alterF f k m =
-    Identity $ alter (runIdentity . f) k m
+  alterF f k m = Identity $ alter (runIdentity . f) k m
   #-}
+#endif
 
 #if MIN_VERSION_base(4,9,0)
--- TODO(m-renaud): Figure out where to import Const from from pre-base-4.9 or
--- how to conditionally include pragrams (the #if around this doesn't work :/).
--- {-# RULES
--- "Const specialize alterF" forall (f :: Maybe a -> Const x (Maybe a)) k m.
---   alterF f k m =
---     Const $ alter (getConst . f) k m
---   #-}
+{-# RULES
+"Const specialize alterF" forall (f :: Maybe a -> Const x (Maybe a)) k m.
+  alterF f k m = Const $ alter (getConst . f) k m
+  #-}
 #endif
 
 {--------------------------------------------------------------------
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -286,12 +286,13 @@ module Data.IntMap.Internal (
 
 #if MIN_VERSION_base(4,8,0)
 import Data.Functor.Identity (Identity (..))
-import Control.Applicative (liftA2)
+import Control.Applicative (liftA2
+#if MIN_VERSION_base(4,9,0)
+                           , Const(..)
+#endif
+                           )
 #else
 import Control.Applicative (Applicative(pure, (<*>)), (<$>), liftA2)
-#if MIN_VERSION_base(4,9,0)
-import Control.Applicative (Const(..))
-#endif
 import Data.Monoid (Monoid(..))
 import Data.Traversable (Traversable(traverse))
 import Data.Word (Word)
@@ -996,7 +997,7 @@ alterF f k m = (<$> f mv) $ \fres ->
 #if MIN_VERSION_base(4,9,0)
 {-# RULES
 "Const specialize alterF" forall (f :: Maybe a -> Const x (Maybe a)) k m.
-  alterF f k m = Const $ alter (getConst . f) k m
+  alterF f k m = Const . getConst . f $ lookup k m
   #-}
 #endif
 
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -289,6 +289,9 @@ import Data.Functor.Identity (Identity (..))
 import Control.Applicative (liftA2)
 #else
 import Control.Applicative (Applicative(pure, (<*>)), (<$>), liftA2)
+#if MIN_VERSION_base(4,9,0)
+import Control.Applicative (Const(..))
+#endif
 import Data.Monoid (Monoid(..))
 import Data.Traversable (Traversable(traverse))
 import Data.Word (Word)
@@ -980,6 +983,24 @@ alterF f k m = (<$> f mv) $ \fres ->
     Nothing -> maybe m (const (delete k m)) mv
     Just v' -> insert k v' m
   where mv = lookup k m
+-- TODO(m-renaud): Figure out if this should be marked INLINE or NOINLINE.
+-- It needs to be one or the other or else the specialization rule may not fire.
+{-# NOINLINE [1] alterF #-}
+{-# RULES
+"Identity specialize alterF" forall (f :: Maybe a -> Identity (Maybe a)) k m.
+  alterF f k m =
+    Identity $ alter (runIdentity . f) k m
+  #-}
+
+#if MIN_VERSION_base(4,9,0)
+-- TODO(m-renaud): Figure out where to import Const from from pre-base-4.9 or
+-- how to conditionally include pragrams (the #if around this doesn't work :/).
+-- {-# RULES
+-- "Const specialize alterF" forall (f :: Maybe a -> Const x (Maybe a)) k m.
+--   alterF f k m =
+--     Const $ alter (getConst . f) k m
+--   #-}
+#endif
 
 {--------------------------------------------------------------------
   Union
--- a/benchmarks/IntMap.hs
+++ b/benchmarks/IntMap.hs
@@ -1,9 +1,16 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
 module Main where
 
+#if MIN_VERSION_base(4,9,0)
+import Control.Applicative (Const(..))
+#endif
 import Control.DeepSeq (rnf)
 import Control.Exception (evaluate)
 import Criterion.Main (bench, defaultMain, whnf)
+#if MIN_VERSION_base(4,8,0)
+import Data.Functor.Identity (Identity (..))
+#endif
 import Data.List (foldl')
 import qualified Data.IntMap as M
 import qualified Data.IntMap.Strict as MS
@@ -35,6 +42,13 @@ main = do
         , bench "update" $ whnf (upd keys) m
         , bench "updateLookupWithKey" $ whnf (upd' keys) m
         , bench "alter"  $ whnf (alt keys) m
+        , bench "alterF" $ whnf (altFList keys) m
+#if MIN_VERSION_base(4,8,0)
+        , bench "alterFIdentity" $ whnf (altFIdentity keys) m
+#endif
+#if MIN_VERSION_base(4,9,0)
+        , bench "alterFConst" $ whnf (altFConst keys) m
+#endif
         , bench "mapMaybe" $ whnf (M.mapMaybe maybeDel) m
         , bench "mapMaybeWithKey" $ whnf (M.mapMaybeWithKey (const maybeDel)) m
         , bench "fromList" $ whnf M.fromList elems
@@ -89,6 +103,22 @@ upd' xs m = foldl' (\m k -> snd $ M.updateLookupWithKey (\_ a -> Just a) k m) m 
 
 alt :: [Int] -> M.IntMap Int -> M.IntMap Int
 alt xs m = foldl' (\m k -> M.alter id k m) m xs
+
+altFList :: [Int] -> M.IntMap Int -> M.IntMap Int
+altFList xs m = foldl' (\m k -> head $ M.alterF (pure . id) k m) m xs
+
+#if MIN_VERSION_base(4,8,0)
+altFIdentity :: [Int] -> M.IntMap Int -> M.IntMap Int
+altFIdentity xs m = foldl' (\m k -> runIdentity $ M.alterF (pure . id) k m) m xs
+#endif
+
+#if MIN_VERSION_base(4,9,0)
+altFConst :: [Int] -> M.IntMap Int -> M.IntMap Int
+altFConst xs m =
+    foldl' (\m k -> getConst $ M.alterF (const (Const m) . id) k m) m xs
+#endif
+
+
 
 maybeDel :: Int -> Maybe Int
 maybeDel n | n `mod` 3 == 0 = Nothing
--- a/tests/intmap-properties.hs
+++ b/tests/intmap-properties.hs
@@ -7,6 +7,12 @@ import Data.IntMap.Lazy as Data.IntMap hiding (showTree)
 #endif
 import Data.IntMap.Internal.Debug (showTree)
 
+#if MIN_VERSION_base(4,9,0)
+import Control.Applicative (Const(..))
+#endif
+#if MIN_VERSION_base(4,8,0)
+import Data.Functor.Identity (Identity (..))
+#endif
 import Data.Monoid
 import Data.Maybe hiding (mapMaybe)
 import qualified Data.Maybe as Maybe (mapMaybe)
@@ -56,6 +62,7 @@ main = defaultMain
              , testCase "updateWithKey" test_updateWithKey
              , testCase "updateLookupWithKey" test_updateLookupWithKey
              , testCase "alter" test_alter
+             , testCase "alterF" test_alterF
              , testCase "union" test_union
              , testCase "mappend" test_mappend
              , testCase "unionWith" test_unionWith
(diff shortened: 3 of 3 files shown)


Data/IntMap/Internal.hs
```diff
@@ -286,7 +286,11 @@ module Data.IntMap.Internal (
 
 #if MIN_VERSION_base(4,8,0)
 import Data.Functor.Identity (Identity (..))
-import Control.Applicative (liftA2)
+import Control.Applicative (liftA2
+#if MIN_VERSION_base(4,9,0)
+                           , Const(..)
+#endif
+                           )
```
treeowl: Please don't split the import list like this for such a little thing. You can just use independent imports for the two versions.

Data/IntMap/Internal.hs
```diff
@@ -980,6 +984,22 @@ alterF f k m = (<$> f mv) $ \fres ->
     Nothing -> maybe m (const (delete k m)) mv
     Just v' -> insert k v' m
   where mv = lookup k m
+-- TODO(m-renaud): Figure out if this should be marked INLINE or NOINLINE.
+-- It needs to be one or the other or else the specialization rule may not fire.
+{-# NOINLINE [1] alterF #-}
+#if MIN_VERSION_base(4,8,0)
+{-# RULES
+"Identity specialize alterF" forall (f :: Maybe a -> Identity (Maybe a)) k m.
+  alterF f k m = Identity $ alter (runIdentity . f) k m
```
treeowl: Use `coerce f` instead.

Data/IntMap/Internal.hs
```diff
@@ -980,6 +984,22 @@ alterF f k m = (<$> f mv) $ \fres ->
...
+  #-}
+#endif
+
+#if MIN_VERSION_base(4,9,0)
+{-# RULES
+"Const specialize alterF" forall (f :: Maybe a -> Const x (Maybe a)) k m.
+  alterF f k m = Const . getConst . f $ lookup k m
```
treeowl: Is the surprising `Const . getConst` needed to change a phantom?

tests/intmap-properties.hs
```diff
@@ -922,6 +976,52 @@ prop_alter t k = case lookup k t of
     f Nothing   = Just ()
     f (Just ()) = Nothing
 
+#if MIN_VERSION_base(4,8,0)
+-- Verify that the rewrite rules for Identity give the same result as the
+-- non-rewritten version. We use our own TestIdentity functor to compare
+-- against.
+
+data TestIdentity a = TestIdentity { runTestIdentity :: a }
```
treeowl: Use `newtype` instead of `data` to keep things as similar as we can.

tests/intmap-properties.hs
```diff
@@ -922,6 +976,52 @@ prop_alter t k = case lookup k t of
...
+    fTest Nothing   = TestIdentity (Just ())
+    fTest (Just ()) = TestIdentity (Nothing)
+#endif
+
+#if MIN_VERSION_base(4,9,0)
+-- Verify that the rewrite rules for Const give the same result
+-- as the non-rewritten version. We use a custom TestConst that
+-- will not fire the rewrite rules to compare against.
+
+data TestConst a b = TestConst { getTestConst :: a }
```
treeowl: Make this `newtype` too please.

m-renaud:
Latest failure is due to CI issue, failed to install ghc for one of the versions.

treeowl:
@m-renaud, you can click the restart button (circle with arrow) next to the test that failed in the Travis interface. Or at least I think you can. If not, let me know and I'll see about getting permissions fixed.

tests/intmap-properties.hs
m-renaud: Done.

tests/intmap-properties.hs
m-renaud: Done.

Data/IntMap/Internal.hs
m-renaud: Used my new-found knowledge of `coerce` :)

Data/IntMap/Internal.hs
m-renaud: TIL about coerce, neat!

Data/IntMap/Internal.hs
m-renaud: Done, yeah I thought that looked a little weird.

m-renaud:
Found it, thanks! I originally wasn't logged in and was on my phone so it wasn't very apparent. Addressed your comments, PTAL.

Data/IntMap/Internal.hs
```diff
@@ -980,6 +985,22 @@ alterF f k m = (<$> f mv) $ \fres ->
...
+{-# RULES
+"Identity specialize alterF" forall (f :: Maybe a -> Identity (Maybe a)) k m.
+  alterF f k m = Identity $ alter (coerce f) k m
+  #-}
+#endif
+
+#if MIN_VERSION_base(4,9,0)
+{-# RULES
+"Const specialize alterF" forall (f :: Maybe a -> Const x (Maybe a)) k m.
+  alterF f k m = coerce . f $ lookup k m
```
treeowl: No need for `coerce` here! If you really need `Const . getConst` here, then use it, but maybe give it a comment or a type signature. We want to use `coerce` only when needed to prevent silly eta expansion from adding closures and indirection at runtime.

treeowl: Or, of course, to re-type a whole structure from the outside.

m-renaud: Yeah, either `Const . getConst` or `coerce` is needed because `f :: Maybe a -> Const c (Maybe a)` but the function returns `Const c (IntMap a)`. Since the second type parameter for `Const` is phantom I would have thought that would be an appropriate use of `coerce`, but I also just learned about Coercible this morning so I may very well be wrong :stuck_out_tongue:.

I've changed it back to `Const . getConst` but I was wondering,  would `Const . getConst` have runtime overhead or should the optimizer figure out that this is just changing a phantom parameter? It seems to me that `Const . getConst` is equivalent to `coerce id` for `Const`.

treeowl:
`Const`, `getConst`, `Const . getConst`, and `runIdentity` should all be very reliably equivalent to `coerce`, at least when optimizations are enabled. The issue that comes with the `Identity` rule is that

```haskell
runIdentity . f = coerce . f =?= coerce f
```

Why is that last equality questionable?

```haskell
coerce . undefined = \x -> coerce (undefined x) = \ _ -> undefined  /=   undefined
```

Yuck! Because of this, GHC will (generally at least) only turn `g (runIdentity . f)` into `coerce g f` if it either

1. Knows that `f` has an arity of at least 1, or
2. Knows that `g` only uses `f` by applying it to an argument, and doesn't do something shady like using `seq` on it. I believe this currently requires that `g` be inlined, but I could be wrong.

In practice, GHC is pretty good at this stuff, but sometimes it goofs. In those circumstances where there's a risk, it's probably best to help it out. In first-order circumstances, there's generally no reason to bother (and make the code less clear).

m-renaud:
Thanks for the explanation! Do you have any more comments, or is this ready to be merged?

treeowl:
I don't understand the benchmark results. What does each of them mean? Please clarify in the commit message and PR text. How does the rewrite rule change performance for `Identity` and for `Const`? Be careful; if you define your own versions of `Identity` and `Const` to test with, be sure to copy the `Functor` and `Applicative` instances from `base`; there's some coercion magic about, as I recall.

treeowl:
For `Identity`:

```haskell
instance Functor Identity where
  fmap = coerce
instance Applicative Identity where
  pure = Identity
  (<*>) = coerce
  liftA2 = coerce
```

For `Const`,

```haskell
instance Monoid m => Applicative (Const m) where
    pure _ = Const mempty
    liftA2 _ (Const x) (Const y) = Const (x `mappend` y)
    (<*>) = coerce (mappend :: m -> m -> m)
```

m-renaud:
Added a section to the PR text, I can add it to the squashed commit when the branch is merged as well.

> be sure to copy the Functor and Applicative instances from base

Good catch, thanks for pointing that out! The `TestConst` functor instance was the same, but I've updated `TestIdentity`to use `coerce`. There's no need to define an applicative instance because alterF only requires `Functor`.

treeowl:
No, I mean what does each line in the benchmark results actually represent? Most especially, what does the `alterF` line mean?

m-renaud:
Ohhh, sorry, I misunderstood. I'll rename (and rework the benchmarks) as soon as I get a chance. I was thinking:

- `alterF_IdentityNoRewrite` (using TestIdentity)
- `alterF_IdentityRewrite` (using Identity)
- `alterF_ConstNoRewrite` (using TestConst)
- `alterF_ConstRewrite` (using Const)

Does that sound better?

treeowl:
Yes, much.

On Jan 4, 2018 11:16 AM, "Matt Renaud" <notifications@github.com> wrote:

Ohhh, sorry, I misunderstood. I'll rename (and rework the benchmarks) as
soon as I get a chance. I was thinking:

   - alterF_IdentityNoRewrite (using TestIdentity)
   - alterF_IdentityRewrite (using Identity)
   - alterF_ConstNoRewrite (using TestConst)
   - alterF_ConstRewrite (using Const)

Does that sound better?

—
You are receiving this because your review was requested.

<https://github.com/haskell/containers/pull/467#issuecomment-355325182>,
or mute
the thread

treeowl:
You have bona fide CI failures. It looks like you forgot to import `Data.Coerce (coerce)` some places.

treeowl:
By the way, there are buttons to restart or cancel CI runs on the Travis page. Do you see those? Please restart your own when there are silly problems, and please try to cancel your own when they're out of date or doomed to fail. Canceling useless builds saves time for the whole Haskell organization. Speaking of which, as I think I said, you should use `[ci skip]` where appropriate. In particular, your whole tutorial project should be `[ci skip]` unless and until we make Travis build those docs. Since we don't even make Travis build the Haddocks, I don't think that's going to happen anytime soon.

treeowl:
Can you fix up the tiny problem so this can pass CI? I want it merged!

m-renaud:
Sorry for the delay, I only had a few minutes here and there over the weekend to look at this stuff. I've re-organized the benchmarks as discussed and fixed the import.

It also appears that the import for `Data.Coerce` in IntMap/Internal.hs was redundant since it's been included in `__GLASGOW_HASKELL__ >= 709`, do I need to add another conditional to import if base >= 4.8 and GLASGOW_HASKELL > 709?

treeowl:
Something is really weird here, and it makes me nervous. Why on Earth do we need to specialize `Const` like that? GHC's automatic specialization *should* be taking care of that one for us! What happens if you change the `NOINLINE [1]` to `INLINABLE [1]`? Does the difference for `Const` go away?

tests/intmap-properties.hs
```diff
@@ -922,6 +977,52 @@ prop_alter t k = case lookup k t of
...
+-- against.
+
+newtype TestIdentity a = TestIdentity { runTestIdentity :: a }
+
+instance Functor TestIdentity where
+    fmap = coerce
+
+prop_alterF_IdentityRules :: UMap -> Int -> Bool
+prop_alterF_IdentityRules t k =
+    runIdentity tIdentity == runTestIdentity tTestIdentity
```
treeowl: This should definitely check that the result is valid and should produce a `Property`.

treeowl:
Um.... I wasn't telling you to just change that. I was suggesting that you check what that does to the benchmarks. I want to understand why there's a difference for `Const`, which I don't think there should be.

tests/intmap-properties.hs
m-renaud: Done.

m-renaud:
I originally had a todo to see if `alterF` should be marked `INLINABLE [1]` or `NOINLINE [1]` but it somehow got clobbered (I went with `NOINLINE` to be safe). The difference in `Const` _mostly_ goes away (see benchmark results in commit description), I think the remaining difference may just be noise (but I wanted to see what you thought). With such a small difference I'm not sure the custom rules for Const are necessary now, wdyt?

treeowl:
Oh, I missed the new benchmark results in your commit. I think you should run more tests. See what happens if you reverse the order the benchmarks run in, for example. If the difference remains, it would be nice to understand why, but we should probably keep the rule. If the difference is really just noise, scrap the rule.

m-renaud:
I ran it several more times and changed the order of benchmarks but the results remain about the same.

treeowl:
Interesting. Let's keep the rule for now but maybe try to dig into Core and
such some other time. We should get this merged.

m-renaud:
Woah woah woah, I ran it a few more times for good measure and got the below results which indicate that the rewrite is now slower than the non-rewritten version :/

```
benchmarking alterF_ConstRewrite
time                 23.26 μs   (22.48 μs .. 23.70 μs)
                     0.993 R²   (0.986 R² .. 0.997 R²)
mean                 22.07 μs   (21.30 μs .. 22.63 μs)
std dev              2.207 μs   (1.692 μs .. 2.816 μs)
variance introduced by outliers: 85% (severely inflated)
             
benchmarking alterF_ConstNoRewrite
time                 18.75 μs   (17.94 μs .. 19.86 μs)
                     0.978 R²   (0.968 R² .. 0.992 R²)
mean                 20.02 μs   (19.11 μs .. 21.07 μs)
std dev              3.218 μs   (2.715 μs .. 3.618 μs)
variance introduced by outliers: 94% (severely inflated)
```

treeowl:
This sounds like some sort of silliness. Maybe some code alignment nonsense
or something. It sounds like they're probably pretty much the same in
reality, so get rid of the Const rule and we can merge.

m-renaud:
Done. I've left a few of the tests and benchmarks in case we want to revisit this in the future.

benchmarks/IntMap.hs
```diff
@@ -90,6 +105,28 @@ upd' xs m = foldl' (\m k -> snd $ M.updateLookupWithKey (\_ a -> Just a) k m) m
...
+
+instance Functor TestIdentity where
+    fmap = coerce
+
+altFTestIdentity :: [Int] -> M.IntMap Int -> M.IntMap Int
+altFTestIdentity xs m =
+    foldl' (\m k -> runTestIdentity $ M.alterF (TestIdentity . id) k m) m xs
+
+altFIdentity :: [Int] -> M.IntMap Int -> M.IntMap Int
+altFIdentity xs m = foldl' (\m k -> runIdentity $ M.alterF (pure . id) k m) m xs
```
treeowl: Waait a minute. You're only testing *one* way of using `alterF`! Please copy over the relevant `Data.Map` benchmarks and test properly! Also, `pure . id` is the same as `pure`.....

treeowl:
Also, please indicate in the commit what GHC version you used for benchmarking, and use GHC >= 8.2. I don't know if the new join point system affects this stuff, but I would not be at all surprised. NB: when you squash this thing, make sure the final commit message is the one with the latest/correctest benchmarks.

m-renaud:
Updated PR description with new benchmark results and note on what GHC version used. Is rewording an already pushed commit allowed? I thought it may mess up history or something.

treeowl:
Speaking of GHC changes, I should probably rerun the benchmarks that led to my choice of `alterF` `Identity` specialization for `Data.Map`.

treeowl:
You've pushed these commits to a branch. You're absolutely free to amend
them, rebase them, etc., etc.
Once they're actually merged to master, neither you nor I has the authority
to touch them. BTW, my
git-fu is horrible (I made a very serious hash of the git history of
Data.Sequence last year by moving
a file the wrong way :-( ). But I can tell you that `git rebase -i` is
*very useful*. Just type `git rebase -i base`,
where base is the last commit before your branch, and you'll get a pretty
much totally self-explanatory way
to fix everything. If you only want one commit, you can set the first one
to "pick" and all the rest to "squash";
this will open an editor with all the texts of all your commits, and you
can delete/condense/etc. to get one
good message. If you're doing something where multiple commits make sense
(here, that might be
one commit for the benchmarks followed by another for the changes), you can
pick a couple, squash
others, etc. If you want to modify something and push that back in time,
you can add a new commit for
it, rebase -i, and move the new commit up higher and then (perhaps) squash
it onto a previous one.

treeowl:
You still need the additional benchmarks ported from `Data.Map`.

m-renaud:
Sooo, these benchmarks tell a different story than the other ones. TL;DR; rewrite rules perform
better in casses when the element being altered is present, but worse when the element is absent. See the commit message for more details.

treeowl:
Hrmmm. Yeah, this kind of weirdness complicated my choices for `Map` as well. I ended up rewriting the `Identity` case not to `alter` but to something that (for reasons I no longer remember) had performance characteristics somewhat more closely matched to the `alterF` version. Or something like that. Been a long time. Maybe you can come up with something? I think we should probably consider this longer and probably let it be till the next release. The benefit isn't nearly as great as for `Map` because `IntMap` lookups are really, really cheap.

m-renaud:
Sounds good, I'll keep poking at it and see if I can come up with something
better. And as you said, we shouldn't block the next release on this.

treeowl:
The biggest reason you're running into trouble is that if the passed function takes `Nothing` to `Nothing` and the key is absent, then

* `alterF` does nothing
* `alter` rebuilds a path in the map

One option is to use pointer equality in `alter` the way we do for many functions in `Data.Map`: if the new subtree is the same as the old, we return our argument unchanged. Another thing to experiment with is GHC's new unboxed sum feature.

```haskell
type Maybe# a = (# Void# | a #)

pattern Nothing# :: Maybe# a
pattern Nothing# <- (# _ | #) where
  Nothing# = (# void# | #)

pattern Just# :: a -> Maybe# a
pattern Just# a = (# | a #)
```

Now we can make recursive calls in `alter` return `Just#` a new map or `Nothing#`, without having to allocate a real `Just` constructor like we would for `Maybe`. We'll have to experiment to see how that performs.

treeowl:
To avoid confusion: `Void#` is not an unboxed version of `Void`. For hysterical raisins, it's an unboxed version of `()`.

m-renaud:
So just using `ptrEq` in `alter`'s zero and otherwise case makes the benchmarks worse (somewhat expected, since it's doing a `ptrEq` at every level). I'm going to try using unboxed sum approach and see if that helps.

treeowl:
One option is to implement alter the way it's done in containers--by hand.
I'm not sure if it's worth it.

m-renaud:
Well, if anything I was able to make `alter` twice as fast using unboxed sums :) I'll move that to its own PR.

PR #478, opened 2018-01-07
Rework Data.Sequence documentation

treeowl:
* Add extended introductory information to the module description.

* Experimentally switch to proper mathematics for big-O. Hopefully
  people won't complain about the loading speed.

* Tighten the performance bound for `chunksOf`.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -1,4 +1,7 @@
 {-# LANGUAGE CPP #-}
+#ifdef __HADDOCK_VERSION__
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+#endif
 
 #include "containers.h"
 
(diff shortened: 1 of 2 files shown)


treeowl:
I obviously need to split this commit, but aside from that, I'd really appreciate some feedback.

Data/Sequence.hs
```diff
@@ -13,34 +16,100 @@
...
+-- type @a@.
+--
+-- Sequences generally behave very much like lists.
+--
+-- * The class instances for sequences are all based very closely on those for
+-- lists.
+--
+-- * Many functions in this module have the same names as functions in
+-- the "Prelude" or in "Data.List". In almost all cases, these functions
+-- behave analogously. For example, 'filter' filters a list in exactly the
```
m-renaud: filters a _sequence_ in exactly the same way that 'Prelude.filter' filters a _list_.

Data/Sequence.hs
```diff
@@ -13,34 +16,100 @@
...
--- also differ from lists in supporting a wider variety of operations
--- efficiently.
+-- = Finite sequences
+--
+-- The @'Seq' a@ type represents a finite sequence of values of
+-- type @a@.
+--
+-- Sequences generally behave very much like lists.
+--
+-- * The class instances for sequences are all based very closely on those for
```
m-renaud: s/class/typeclass/ (?) Maybe its just me but its weird to see them being referred to as just "classes".

Data/Sequence.hs
```diff
@@ -13,34 +16,100 @@
...
+--  "Data.Map" rather than the one from "Data.List".
+--
+-- There are two major differences between sequences and lists:
+--
+-- * Sequences support a wider variety of efficient operations than
+-- do lists. Notably, they offer
+--
+--     * Constant-time access to both the front and the rear, with
+--     '<|', '|>', 'viewl', 'viewr', and the bidirectional pattern
+--     synonyms ':<|' and ':|>'.
```
m-renaud: Maybe mention that :<| and :|> are the preferred way?

Data/Sequence.hs
treeowl: I don't think it's weird.

Data/Sequence.hs
treeowl: Thanks.

Data/Sequence.hs
treeowl: OK.

treeowl:
I'm not moving the complexity info in this PR, but we can think about it later. We need to be sure that it's really easy to find, wherever we put it.

PR #485, opened 2018-01-10
Verify site ownership for Google analytics/webmaster tools.

m-renaud:
This will allow us to provide additional information to the crawler, improving search results when people are searching for containers documentation.

[ci skip]


treeowl:
I have no idea what that means.

m-renaud:
In order to access the [webmaster console](https://www.google.com/webmasters) you need to prove that you own the site. You do this by adding some meta attribute that is generated for you to the <head> of your page which is then checked. This has no effect on the rendering/functionality of the page.

treeowl:
So who will have access?

treeowl:
Sounds good then.

treeowl:
Please document somewhere all the steps that will be needed, when adding an
additional maintainer, to make sure they have access to both that and
ReadTheDocs. Also include instructions for your own replacement--no
maintainer lasts forever.

On Jan 11, 2018 2:08 PM, "David Feuer" <david.feuer@gmail.com> wrote:

Sounds good then.

PR #490, opened 2018-01-11
Added notes on skew-heap attempt

oisdk:
This adds notes on the performance of the pairing heap unstable sorting implementation, and on the performance of a skew heap implementation. The skew heap implementation is slower, so it's described here to save anyone else from duplicating work.


treeowl:
Please mention this note in a comment on `unstableSort` in `Data.Sequence.Internal`. And thanks!

treeowl:
Thanks a lot!

Issue #491, opened 2018-01-11
Speed up stable sorting


oisdk:
Since the unstable sort is significantly quicker than the stable variant, I think you can write a stable sort that uses it internally, and the overhead of "stabilising" it should be small enough that it's still faster than Data.List.sortBy.

I've [written an implementation](https://github.com/oisdk/containers/tree/faster-stable-sort) which uses a pairing heap which tags each element with its original position:

```haskell
data PQS e = PQS {-# UNPACK #-} !Int e (PQSL e)
data PQSL e = Nl | {-# UNPACK #-} !(PQS e) :&& PQSL e
```

It's quicker than the old implementation (which used Data.List.sort) in the general case (all times are on sequences of length 50000):

```
Times (ms)            min    est    max  std dev   r²
to/from list:        64.23  64.50  64.81  0.432  1.000
1/11/18 stable heap: 38.87  39.40  40.09  0.457  0.999
```

Slightly slower in the case of already sorted lists:

```
Times (ms)            min    est    max  std dev   r²
to/from list:        6.806  6.861  6.901  0.234  1.000
1/11/18 stable heap: 8.211  8.268  8.328  0.111  1.000
```

And quicker in the case of lists sorted in reverse:

```
Times (ms)            min    est    max  std dev   r²
to/from list:        26.79  28.34  30.55  1.219  0.988
1/11/18 stable heap: 9.405  10.13  10.91  0.670  0.977
```

Interestingly, the stable sort is now competitive with the unstable:

```
Times (ms)            min    est    max  std dev   r²
unstable:            34.71  35.10  35.38  0.845  1.000
stable:              38.84  39.22  39.59  0.564  0.999
```

And even beats it in the case of already-sorted lists:

```
Times (ms)            min    est    max  std dev   r²
unstable:            8.457  8.499  8.536  0.069  1.000
stable:              8.160  8.230  8.334  0.158  0.999
```

I couldn't think of an way to actually use `unstableSort`: the new stable sort tags elements with their position while building the heap (in a similar way to foldMapWithIndex). As a result, there's a lot of code duplication.

treeowl:
What happens if you actually *use* foldMapWithIndex?

On Jan 11, 2018 3:46 PM, "Donnacha Oisín Kidney" <notifications@github.com>
wrote:

Since the unstable sort is significantly quicker than the stable variant, I
think you can write a stable sort that uses it internally, and the overhead
of "stabilising" it should be small enough that it's still faster than
[...]

PR #492, opened 2018-01-11
Faster stable sort

oisdk:
As described in #491 

The stable sort now performs a heapsort, similar to the unstable sort, except it tags elements with their original positions to "stabilise" it. This is faster than the old implementation, which used Data.List.sortBy.

### DIFF
--- a/Data/Sequence/Internal.hs
+++ b/Data/Sequence/Internal.hs
@@ -4554,32 +4554,35 @@ popMin cmp = State unrollPQ'
 -- elements, converts a 'Seq' to a 'PQueue'.
 toPQ :: (e -> e -> Ordering) -> Seq e -> Maybe (PQueue e)
 toPQ cmp' (Seq xs') = toPQTree cmp' (\(Elem a) -> PQueue a Nil) xs'
+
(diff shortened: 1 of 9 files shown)


Data/Sequence/Internal.hs
```diff
@@ -4554,32 +4554,35 @@ popMin cmp = State unrollPQ'
...
-            Node2 _ a b -> f a <+> f b
-            Node3 _ a b c -> f a <+> f b <+> f c
-      where
-        (<+>) = mergePQ cmp
+    pr' = toPQDigit cmp f pr
+    sf' = toPQDigit cmp f sf
+    m' = toPQTree cmp (toPQNode cmp f) m
+    (<+>) = mergePQ cmp
+
+toPQDigit :: (b -> b -> Ordering) -> (a -> PQueue b) -> Digit a -> PQueue b
```
treeowl: Do you really need them top-level, or is asking a signature for `toPQNode` sufficient?

treeowl:
Also, please update `sorting.md`.

treeowl:
Your type signatures are missing contexts.

treeowl:
Thanks a lot, @oisdk!

Issue #489, opened 2018-01-11
Add HasCallStack to partial functions


dwincort:
Can we add `HasCallStack` to the constraints of partial functions like `Map.!`?  I'd be happy to put together a PR if it's a change that would get accepted.

treeowl:
Yes, please! I was already thinking about doing that. But please show
benchmark results to make sure the performance impact is tolerable.

sjakobi:
Related: https://github.com/haskell-unordered-containers/unordered-containers/pull/252

sjakobi:
Since the analogous work on `unordered-containers` was very warmly received, I'd love to see progress on this issue.

AFAICT not all changes in #493 are problematic. I think it might be a good first step to extract those into a separate PR that can be merged quickly.

(CC @bollu)

dwincort:
Oh wow, I totally forgot about this.  The project I was working on ended up finding a different way to track errors, and this went to the bottom of my priority list.

I don't have the bandwidth right now to work on this, so if someone else wants to take over, that would be great.  Otherwise, I may be able to find some time in a few weeks.  I'll make a reminder for myself.

sjakobi:
For reference: Similar plans for `base` are discussed in https://gitlab.haskell.org/ghc/ghc/-/issues/17040.

treeowl:
Yes, I tend to think it's acceptable if minimal. I think the theoretical justification is that it can help an end-user of software using the library write a useful bug report if said software has a bug.

PR #493, opened 2018-01-13
Added HasCallStack to partial functions

dwincort:
As discussed on Issue https://github.com/haskell/containers/issues/489, this PR adds `HasCallStack` to the constraints of partial functions.  I think I got all of the relevant partial functions, but I'd be happy for someone to double check that I didn't miss anything.

I ran the benchmarks (`stack bench`) before and after these changes and have attached them ([before_hascallstack.bench.txt](https://github.com/haskell/containers/files/1628500/before_hascallstack.bench.txt), [after_hascallstack.bench.txt](https://github.com/haskell/containers/files/1628499/after_hascallstack.bench.txt)).  I scanned through the results, trying to pay attention to tests on partial functions specifically, but I didn't notice any changes (outside of expected deviations within the margin of error).

Please let me know if there's more that I can do to help get this PR successfully merged.

### DIFF
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -331,6 +331,8 @@ import qualified Control.Category as Category
 import Data.Coerce
 #endif
 
+import GHC.Stack (HasCallStack)
+
 
 -- A "Nat" is a natural machine word (an unsigned Int)
 type Nat = Word
@@ -390,7 +392,7 @@ bitmapOf x = shiftLL 1 (x .&. IntSet.suffixBitMask)
 -- > fromList [(5,'a'), (3,'b')] ! 1    Error: element not in the map
 -- > fromList [(5,'a'), (3,'b')] ! 5 == 'a'
 
-(!) :: IntMap a -> Key -> a
+(!) :: HasCallStack => IntMap a -> Key -> a
 (!) m k = find k m
 
 -- | /O(min(n,W))/. Find the value at a key.
@@ -2169,11 +2171,11 @@ minView :: IntMap a -> Maybe (a, IntMap a)
 minView t = liftM (first snd) (minViewWithKey t)
 
 -- | /O(min(n,W))/. Delete and find the maximal element.
-deleteFindMax :: IntMap a -> ((Key, a), IntMap a)
+deleteFindMax :: HasCallStack => IntMap a -> ((Key, a), IntMap a)
 deleteFindMax = fromMaybe (error "deleteFindMax: empty map has no maximal element") . maxViewWithKey
 
 -- | /O(min(n,W))/. Delete and find the minimal element.
-deleteFindMin :: IntMap a -> ((Key, a), IntMap a)
+deleteFindMin :: HasCallStack => IntMap a -> ((Key, a), IntMap a)
 deleteFindMin = fromMaybe (error "deleteFindMin: empty map has no minimal element") . minViewWithKey
 
 -- | /O(min(n,W))/. The minimal key of the map. Returns 'Nothing' if the map is empty.
@@ -2188,7 +2190,7 @@ lookupMin (Bin _ m l r)
           go Nil            = Nothing
 
 -- | /O(min(n,W))/. The minimal key of the map. Calls 'error' if the map is empty.
-findMin :: IntMap a -> (Key, a)
+findMin :: HasCallStack => IntMap a -> (Key, a)
 findMin t
   | Just r <- lookupMin t = r
   | otherwise = error "findMin: empty map has no minimal element"
@@ -2205,7 +2207,7 @@ lookupMax (Bin _ m l r)
           go Nil            = Nothing
 
 -- | /O(min(n,W))/. The maximal key of the map. Calls 'error' if the map is empty.
-findMax :: IntMap a -> (Key, a)
+findMax :: HasCallStack => IntMap a -> (Key, a)
 findMax t
   | Just r <- lookupMax t = r
   | otherwise = error "findMax: empty map has no maximal element"
--- a/Data/IntSet/Internal.hs
+++ b/Data/IntSet/Internal.hs
@@ -216,6 +216,7 @@ import qualified GHC.Exts as GHCExts
 import GHC.Prim (indexInt8OffAddr#)
 #endif
 
+import GHC.Stack (HasCallStack)
 
 infixl 9 \\{-This comment teaches CPP correct behaviour -}
 
@@ -793,18 +794,18 @@ minView t =
 -- | /O(min(n,W))/. Delete and find the minimal element.
 --
 -- > deleteFindMin set = (findMin set, deleteMin set)
-deleteFindMin :: IntSet -> (Key, IntSet)
+deleteFindMin :: HasCallStack => IntSet -> (Key, IntSet)
 deleteFindMin = fromMaybe (error "deleteFindMin: empty set has no minimal element") . minView
 
 -- | /O(min(n,W))/. Delete and find the maximal element.
 --
 -- > deleteFindMax set = (findMax set, deleteMax set)
-deleteFindMax :: IntSet -> (Key, IntSet)
+deleteFindMax :: HasCallStack => IntSet -> (Key, IntSet)
 deleteFindMax = fromMaybe (error "deleteFindMax: empty set has no maximal element") . maxView
 
 
 -- | /O(min(n,W))/. The minimal element of the set.
-findMin :: IntSet -> Key
+findMin :: HasCallStack => IntSet -> Key
 findMin Nil = error "findMin: empty set has no minimal element"
 findMin (Tip kx bm) = kx + lowestBitSet bm
 findMin (Bin _ m l r)
@@ -815,7 +816,7 @@ findMin (Bin _ m l r)
           find Nil            = error "findMin Nil"
 
 -- | /O(min(n,W))/. The maximal element of a set.
-findMax :: IntSet -> Key
+findMax :: HasCallStack => IntSet -> Key
 findMax Nil = error "findMax: empty set has no maximal element"
 findMax (Tip kx bm) = kx + highestBitSet bm
 findMax (Bin _ m l r)
--- a/Data/Map/Internal.hs
+++ b/Data/Map/Internal.hs
@@ -411,6 +411,7 @@ import qualified Control.Category as Category
 import Data.Coerce
 #endif
 
+import GHC.Stack (HasCallStack)
 
 {--------------------------------------------------------------------
   Operators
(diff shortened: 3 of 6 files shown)


treeowl:
You're going to need to wrap all those pieces in CPP restricting in to appropriate `__GLASGOW_HASKELL__` versions.

dwincort:
Is it a `__GLASGOW_HASKELL__` version or is it `MIN_VERSION_base(4,9,0)`?

Thanks!

treeowl:
Glasgow Haskell, because it really relies on GHC-specific features.

Data/Map/Internal.hs
```diff
@@ -1566,7 +1581,11 @@ splitAt i0 m0
 -- > updateAt (\_ _  -> Nothing)  2    (fromList [(5,"a"), (3,"b")])    Error: index out of range
 -- > updateAt (\_ _  -> Nothing)  (-1) (fromList [(5,"a"), (3,"b")])    Error: index out of range
 
+#if __GLASGOW_HASKELL__ >= 800
+updateAt :: HasCallStack => (k -> a -> Maybe a) -> Int -> Map k a -> Map k a
+#else
 updateAt :: (k -> a -> Maybe a) -> Int -> Map k a -> Map k a
+#endif
```
treeowl: Would it be better to just make this function total by making an update at a bad index do nothing?

Data/Map/Internal.hs
```diff
@@ -1588,7 +1607,11 @@ updateAt f !i t =
 -- > deleteAt 2 (fromList [(5,"a"), (3,"b")])     Error: index out of range
 -- > deleteAt (-1) (fromList [(5,"a"), (3,"b")])  Error: index out of range
 
+#if __GLASGOW_HASKELL__ >= 800
+deleteAt :: HasCallStack => Int -> Map k a -> Map k a
+#else
```
treeowl: Same question here.

Data/Map/Internal.hs
```diff
@@ -2661,7 +2692,11 @@ mergeA
 -- @only2@ are 'id' and @'const' 'empty'@, but for example @'map' f@,
 -- @'filterWithKey' f@, or @'mapMaybeWithKey' f@ could be used for any @f@.
 
+#if __GLASGOW_HASKELL__ >= 800
+mergeWithKey :: (HasCallStack, Ord k)
```
treeowl: Ehhhhh... Not sure we should bother with this one. `mergeWithKey` is for users who don't mind the risk of being shot in the foot.

Data/Sequence/Internal.hs
```diff
@@ -4300,7 +4348,11 @@ zipWith f s1 s2 = zipWith' f s1' s2'
     s2' = take minLen s2
 
 -- | A version of zipWith that assumes the sequences have the same length.
+#if __GLASGOW_HASKELL__ >= 800
+zipWith' :: HasCallStack => (a -> b -> c) -> Seq a -> Seq b -> Seq c
+#else
```
treeowl: No no no. This is an internal function and is only used when we know it will succeed.

Data/Sequence/Internal.hs
```diff
@@ -4498,7 +4554,11 @@ draw (PQueue x ts0) = x : drawSubTrees ts0
 -- | 'popMin', given an ordering function, constructs a stateful action
 -- which pops the smallest elements from a queue. This action will fail
 -- on empty queues.
+#if __GLASGOW_HASKELL__ >= 800
+popMin :: HasCallStack => (e -> e -> Ordering) -> State (PQueue e) e
```
treeowl: No no no. This is an internal function.

Data/Map/Internal.hs
dwincort: This is definitely a more interesting question than just where to add `HasCallStack` and might make more sense for a different issue/PR.  Personally, I'm not sure I've ever used the `***At` functions, so I'm not really the target audience to be weighing in on this decision.  Is there a way you could get community feedback about such a change?  Maybe Haskell-cafe?

treeowl:
You're right; separate PR. I guess I can ask the libraries list.

Data/IntSet/Internal.hs
```diff
@@ -793,18 +796,30 @@ minView t =
...
 deleteFindMax = fromMaybe (error "deleteFindMax: empty set has no maximal element") . maxView
 
 
 -- | /O(min(n,W))/. The minimal element of the set.
+#if __GLASGOW_HASKELL__ >= 800
+findMin :: HasCallStack => IntSet -> Key
+#else
 findMin :: IntSet -> Key
+#endif
 findMin Nil = error "findMin: empty set has no minimal element"
```
treeowl: Does the call stack grow through the recursion here or elsewhere? If so, that's a big problem. You can check the Core to be sure. We know there are no `Nil`s except at the root, but GHC does not! If the stacks build, you'll need to restructure the functions to fix that. Watch out for performance. If the times for the current benchmarks exercising these functions are too short to trust, consider adding more benchmarks.

Data/Sequence/Internal.hs
```diff
@@ -4457,7 +4505,11 @@ unstableSortBy cmp (Seq xs) =
 -- | fromList2, given a list and its length, constructs a completely
 -- balanced Seq whose elements are that list using the replicateA
 -- generalization.
+#if __GLASGOW_HASKELL__ >= 800
+fromList2 :: HasCallStack => Int -> [a] -> Seq a
```
treeowl: This is an internal function and shouldn't get a call stack. It's used to implement the `IsList` instance. Arguably, the method it's used to implement should get a call stack, but that method isn't really intended to be called directly in user code anyway, and GHC will always call it correctly.

Data/IntSet/Internal.hs
dwincort: Wow, yes it does.  I didn't realize it would do that.  The fix is to use an internal `go` function.  I'll update the PR with that in a moment.

On a related note, I realized that my `HasCallStack` for `Map.!` was wrong because `Map.!` calls `find` which actually throws the error.  Is there a reason it uses `find` rather than `lookup`?  My inclination is to remove `find` altogether (it's only used by `!` and looks identical other than the `Maybe` wrapper) and replace with `lookup`---then `!` can call `error` itself, which should then make `HasCallStack` work the way we want.

dwincort: Oh, I just noticed the `[Note: Local 'go' functions and capturing]` -- I'll read that and try to make sure I'm not doing anything stupid.

treeowl:
You may be able to avoid capturing issues by giving the `go` functions full
signatures, making it clear they don't take call stacks. Check Core often,
maybe look at STG occasionally, and benchmark.

Data/IntMap/Internal.hs
```diff
@@ -399,7 +399,9 @@ bitmapOf x = shiftLL 1 (x .&. IntSet.suffixBitMask)
 #else
 (!) :: IntMap a -> Key -> a
 #endif
-(!) m k = find k m
+(!) m k
+  | Just a <- lookup k m = a
+  | otherwise = error ("IntMap.!: key " ++ show k ++ " is not an element of the map")
```
treeowl: Using `lookup` will incur a `Just` allocation. One possible solution (for very recent GHC versions only) would be to define a `lookup#` function producing an unboxed sum:

```haskell
type Maybe# a = (# Void# | a #)

pattern Nothing# :: Maybe# a
pattern Nothing# <- (# _ | #) where
  Nothing# = (# void# | #)

pattern Just# :: a -> Maybe# a
pattern Just# a = (# | a #)

lookup# :: Key -> IntMap a -> Maybe# a
lookup# k m = ...

lookup k m = case lookup# k m of
  Nothing# -> Nothing
  Just# a -> Just a
```

The (potentially) great thing about this is that the `Just` gets applied on the "outside", so inlining and the case-of-case transformation will end up making it go away altogether:

```haskell
m ! k = case
    (case lookup# k m of
       Nothing# -> Nothing
       Just# a -> Just a) of
       Nothing -> error ...
       Just v -> v

-- ==> case-of-case, case of known constructor

m ! k =
  case lookup# k m of
       Nothing# -> error ...
       Just# a -> a
```

Unlike a CPS version, this theoretically helps user-written code as well. However, I have not actually tried benchmarking anything like this. And the CPP is potentially troublesome too.

Data/Map/Strict/Internal.hs
```diff
@@ -887,9 +887,20 @@ atKeyIdentity k f t = Identity $ atKeyPlain Strict k (coerce f) t
 
 #if __GLASGOW_HASKELL__ >= 800
 updateAt :: HasCallStack => (k -> a -> Maybe a) -> Int -> Map k a -> Map k a
+updateAt = go where
```
treeowl: Does this actually give the right call stacks? Thinking about this more carefully, I suspect you want to check if the operation will succeed, and only then call a `go` function that takes no call stack. The alternative would be to call the `go` function using `withFrozenCallStack`, which I suspect will be more expensive. Benchmarks may tell.

treeowl:
Are you making any progress? I'd love to get this into the next release, but that's probably going to happen in the morning!

dwincort:
I'm sorry, but I got caught up in other work this past week and haven't worked on this at all.  I guess it will have to wait until the next release :/.

Data/IntMap/Internal.hs
dwincort: This seems cool -- it seems like it could be a fairly big change that would even prevent an allocation when one uses `lookup` (or related functions that return `Maybe` values) and immediately unwraps -- but it's a bit big for what I was hoping to contribute just now (also, a bit over my head).  I think I'll take the easy way out and go back to using a `find` function to get a result without the extra allocation.  Perhaps my next PR will be to overhaul the backend to use unboxed sums where possible :).

Data/Map/Strict/Internal.hs
dwincort: My tests show that, by default, functions like `go` do not have the `HasCallStack` constraint unless it is explicit in their type signatures.  Therefore, this seems to give the right call stack.  There's no need to check if the operation succeeds or not first, and because `go` doesn't have the constraint, there's no need for `withFrozenCallStack`.

dwincort:
More generally, I must confess that I really don't know how to check Core.  I think I know how to _look at_ Core (`-ddump-simpl`?), but I don't know what to do with it or how to tell if it's good or bad.

treeowl:
`-ddump-simpl` indeed. I suggest that for a while you should add `-dsuppress-all -dno-suppress-type-signatures`. You want to look for fishy things like recursively growing call stacks.

Separately, you should test each function with bad arguments of all relevant sorts and make sure the call stack indicates that the problem is with the bad code that called the partial function, not the partial function itself.

bollu:
What is the status on this? I wound up writing custom wrappers around some of the functions in `Data.Map` for stack traces, but it would be awesome if this were upstreamed.

treeowl:
No one has done the grunt work to check on whether this is optimized the
way we want and whether the performance regressions are tolerable.

bollu:
Is it necessary to pay a performance cost? Can we not expose a `Data.Map.{Strict, Lazy}.Debug` that contains the functions that expose debug info?/

meooow25:
Closing in favor of #1160

PR #494, opened 2018-01-15
Add unzips for sequences

treeowl:
* Add `unzip` and `unzipWith` to `Data.Sequence`.

* Make `unzipWith` *actually* avoid certain space leaks by
  ensuring that the result trees are constructed in lock-step.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -211,13 +211,15 @@ module Data.Sequence (
     traverseWithIndex, -- :: Applicative f => (Int -> a -> f b) -> Seq a -> f (Seq b)
     reverse,        -- :: Seq a -> Seq a
     intersperse,    -- :: a -> Seq a -> Seq a
-    -- ** Zips
+    -- ** Zips and unzip
     zip,            -- :: Seq a -> Seq b -> Seq (a, b)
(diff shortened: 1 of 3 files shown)


treeowl:
That was way more work than I expected! I should really add a regression test, but I'm not sure how to check residency for a test. I did something like this:

```haskell
data List a = Cons a !(List a) | Nil
replicateList :: Int -> a -> List a
replicateList = go Nil where
  go acc 0 _ = acc
  go acc n x = go (Cons x acc) (n - 1) x

smallSeq = fromFunction 500 id  -- smallSeq = fromList [0..499]
(xs, ys) = unzipWith (\x -> (replicateList (10^6) x, "Simple!")) smallSeq
```

Then I forced all the elements of `xs` (after which `xs` was dead and none of its elements were used). Finally, I performed a (trivial) calculation on `ys`. Without this patch, I ended up running out of memory, because `ys` was actually holding on to all the horribly long lists. With this patch, everything seems to work out fine. There's a cost, of course: we have to build up structure for each sequence as the *other* one is forced. I still think this is the way to go.

treeowl:
The first rule shouldn't be necessary. Once `unzip` inlines, the second rule should take care of everything. I need to add `@since` annotations and a changelog entry.

PR #496, opened 2018-01-18
IntSetValidity: Fix the maskPowerOfTwo invariant check

nomeata:
m `mod` 2 == 0

does not check that `m` is a power fo two. But every number with a
popCount of 1 is!

### DIFF
--- a/tests/IntSetValidity.hs
+++ b/tests/IntSetValidity.hs
@@ -1,7 +1,7 @@
 {-# LANGUAGE CPP #-}
 module IntSetValidity (valid) where
 
-import Data.Bits (xor, (.&.), popCount)
+import Data.Bits (xor, (.&.))
 import Data.IntSet.Internal
 import Test.QuickCheck (Property, counterexample, property, (.&&.))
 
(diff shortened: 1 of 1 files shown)


treeowl:
Thanks! Does `IntMap` have a similar problem?

nomeata:
Good question. The `IntMapValidity` does not even check that the mask is a power of two.

m-renaud:
Thanks for fixing this! I have no idea why I was thinking divisible by 2 :P Anyways, I can shed some light on why `IntMapValidity` doesn't have it: when I was writing the checks there wasn't a "masks are powers of 2" invariant comment for IntMap, and when adding the (broken) `powerOfTwo` check the tests failed. I was still learning about the internal representation and chalked the failure up to IntMap doing things slightly differently than IntSet.

The reason why the IntSet tests passed with the broken function is because the lowest 6 bits will never be in the IntSet mask because a bitmap is used in the Leaf to track the lowest significant bits. IntMap on the other hand only stores one entry per Leaf so its entirely possible that the switching bit is `1` (not divisible by 2, but is a power of 2).

I can send a PR updating the invariant comments in IntMap and adding the correct power of two check.

treeowl:
Let's get it done!

m-renaud:
`bitcount` doesn't appear to be exported from the `IntSet.Internal` module, I also need it in my PR for adding this check to `IntMapValidity`. Maybe we should consider moving the definition to `Utils.Containers.Internal.BitUtil`, wdyt?

treeowl:
Yes.

treeowl:
Actually, can we (or do we) conditionally depend on base-compat and scrap
that?

treeowl:
Er... Never mind.

m-renaud:
Created #498.

nomeata:
I’ll abandon this and let @m-renaud just fix both of them, that’s easier to coordinate.

treeowl:
Thanks for your help.

Issue #495, opened 2018-01-17
Documentation master task list


m-renaud:
- [x] Improve Data.Sequence module description
- [ ] Improve Data.Map (Lazy and Strict variants) module descriptions
- [ ] Improve Data.IntMap (Lazy and Strict variants) module descriptions
- [ ] Improve Data.Set module description
- [ ] Improve Data.IntSet module description
- [ ] Improve Data.Graph module description
- [ ] Improve Data.Tree module description
- [ ] Consistently format Haddock examples in expandable sections throughout package
- [ ] Gather consensus on Haddock format (#486) and use throughout package
- [ ] Add longer Set examples to bottom of [Set introduction](https://haskell-containers.readthedocs.io/en/latest/sets.html)
- [ ] Add longer Map examples to bottom of [Map introduction](https://haskell-containers.readthedocs.io/en/latest/maps.html)
- [ ] Add longer Sequence examples to bottom of [Sequence introduction](https://haskell-containers.readthedocs.io/en/latest/sequence.html)
- [ ] Add introduction for Data.Tree at https://haskell-containers.readthedocs.io
- [ ] Add examples to Data.Tree Haddocks 
- [ ] Add introduction for Data.Graph at https://haskell-containers.readthedocs.io
- [ ] Add examples to Data.Graph Haddocks

**Notes:**
- Module Haddock descriptions should:
  + link out to appropriate page at https://haskell-containers.readthedocs.io
  + follow the format of the [new Data.Sequence module description](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Sequence.html)
  + link to appropriate [performance benchmarks](https://github.com/haskell-perf)

PR #497, opened 2018-01-18
Improve Data.Map (strict and lazy) module docs

m-renaud:
This improves the module level docs for `Data.Map.Strict` and `Data.Map.Lazy` following the form of the recent improvements to `Data.Sequence`. This is for the most part a re-organization of the existing content with a few minor changes so should hopefully be uncontroversial.

Rendered docs can be found [here](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Map-Strict.html) and [here](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Map-Lazy.html).

Partially addresses #495.


m-renaud:
@treeowl any concerns with this change? If not I'd like to do something similar with the Set modules (as well as the Int variants of Set and Map).

treeowl:
Let me take a look. I have a cold right now and things aren't getting done
as quickly as usual.

m-renaud:
Oh no, hope you feel better soon! Take your time :)

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
 --
--- An efficient implementation of ordered maps from keys to values
--- (dictionaries).
 --
--- API of this module is strict in the keys, but lazy in the values.
--- If you need value-strict maps, use "Data.Map.Strict" instead.
--- The 'Map' type itself is shared between the lazy and strict modules,
--- meaning that the same 'Map' value can be passed to functions in
--- both modules (although that is rarely needed).
+-- = Lazy Maps (aka dictionaries)
```
treeowl: Do not put the "aka dictionaries" in the title. Also don't write it that way anywhere else. Also, we don't actually *have* strict and lazy `Map`s. We have a single `Map` type offering lazy and strict *operations*. The title should probably be "Finite Maps (lazy interface)" or something like that.

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
 --
--- These modules are intended to be imported qualified, to avoid name
--- clashes with Prelude functions, e.g.
+-- The @'Map' k v@ type represents an /ordered/ map indexed by keys of type @k@
+-- containing /lazy/ values of type @v@.
```
treeowl: What is an "ordered map"? Again, the `Map` *type* here is exactly the same as the one in `.Strict`. I would probably change the intro to something like this:

    -- The @'Map' k v@ type represents a finite map (sometimes called a dictionary)
    -- from keys of type @k@ to values of type @v@. A 'Map' is strict in its keys but lazy
    -- in its values.
    --
    -- The functions in "Data.Map.Strict" are careful to force values before installing
    -- them in a 'Map'. This is usually more efficient in cases where laziness is not essential.
    -- The functions in this module do not do so.

If you make the change this way, please remove the first bullet point below.

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
 --
--- >  import qualified Data.Map.Lazy as Map
+-- When deciding if this is the correct data structure to use, consider:
+--
+-- * The API of this module is strict in the keys but lazy in the values. If you
+-- don't need /value-lazy/ maps, use "Data.Map.Strict" instead.
+--
+-- * If your key type is @Int@, use "Data.IntMap.Lazy" instead.
```
treeowl: This is a bit harsh and also a bit wrong. `Data.IntMap.Lazy` is *usually* faster, but its implementation of `size` is very slow. That might matter to some people. Perhaps something like

    -- * If you are using 'Int' keys, you will get much better performance for most operations using
    -- "Data.IntMap.Lazy".

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
+--
+-- * If you don't care about ordering, use @Data.HashMap.Lazy@ from the
+-- <https://hackage.haskell.org/package/unordered-containers unordered-containers>
+-- package instead.
```
treeowl: This is often right, but not always. Size calculations for `HashMap` are very slow (although someone wants to try to fix that, I'm not sure it's going to happen). `Map` unions also have much tighter performance bounds than `HashMap` ones do. I would just say that people should consider using `unordered-containers` instead.

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
+--
+-- This module is intended to be imported qualified, to avoid name clashes with
+-- Prelude functions:
+--
+-- > import qualified Data.Map.Lazy as Map
+--
+-- Note that the implementation is /left-biased/, the elements of a first
+-- argument are always preferred to the second, for example in 'union' or
+-- 'insert'.
```
treeowl: This is generally true, but not strictly. In particular, `fromList` violates this expectation by preferring the elements that occur "to the right" in the list. Also, the first argument of `insert` is a key, and its second argument is a value, so what you wrote doesn't exactly make sense, although I can understand what you're *trying* to say. I think a narrower statement is almost certainly better, and special cases can be handled as they arise:

    -- Note that the implementation is generally /left-biased/. Functions that take
    -- two maps as arguments and combine them, such as `union` and `intersection`,
    -- prefer the values in the first argument to those in the second.

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
+-- == Detailed performance information
+--
+-- The amortized running time is given for each operation, with /n/ referring to
+-- the number of entries in the map.
+--
+-- Benchmarks comparing "Data.Map.Lazy" with other dictionary implementations
+-- can be found at https://github.com/haskell-perf/dictionaries.
+--
+--
+-- == Strictness propertios
```
treeowl: s/propertios/properties

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,63 @@
...
+--
+-- This module satisfies the following strictness property:
+--
+-- * Key arguments are evaluated to WHNF
```
treeowl: I don't think this section needs to be here. It's covered by the statement above that maps are strict in their keys. If someone doesn't understand what that means, they can check out your tutorial documentation or other Haskell documentation.

Data/Map/Strict.hs
```diff
@@ -15,19 +15,73 @@
...
--- An efficient implementation of ordered maps from keys to values
--- (dictionaries).
 --
--- API of this module is strict in both the keys and the values.
--- If you need value-lazy maps, use "Data.Map.Lazy" instead.
--- The 'Map' type is shared between the lazy and strict modules,
--- meaning that the same 'Map' value can be passed to functions in
--- both modules (although that is rarely needed).
+-- = Strict Maps (aka dictionaries)
+--
```
treeowl: Almost all my comments on `Data.Map.Lazy` apply to `Data.Map.Strict`.

Data/Map/Strict.hs
```diff
@@ -15,19 +15,73 @@
...
+--
+-- == Strictness propertios
+--
+-- This module satisfies the following strictness properties:
+--
+-- 1. Key arguments are evaluated to WHNF;
+--
+-- 2. Keys and values are evaluated to WHNF before they are stored in
+--    the map.
```
treeowl: Remove point 1 and all examples attempting to illustrate it. Instead, write something like

    -- As described, 'Map's are strict in their keys. In addition, functions in this module evaluate
    -- values before storing them in maps. More precisely:
    --
    --     If all values in all 'Map' arguments to a function are in WHNF, then
    --     all values in all 'Map' results will be in WHNF.

We need to be a bit careful because some operations (such as `union`) will reuse subtrees without touching them.

Data/Map/Lazy.hs
```diff
@@ -14,19 +14,66 @@
...
+--
+-- Here are some examples that illustrate the property:
+--
+-- > insertWith (\ new old -> old) undefined v m  ==  undefined
+-- > insertWith (\ new old -> old) k undefined m  ==  OK
+-- > delete undefined m  ==  undefined
```
treeowl: This `delete` example isn't even an example (Can you imagine *any* sort of finite map implementation with a lazy `delete`? What information could you actually obtain from a map from which an unknown and possibly-absent key has been deleted?) But I think all these examples should go away.

Data/Map/Strict.hs
```diff
@@ -46,29 +103,16 @@
...
---
--- /Warning/: The size of the map must not exceed @maxBound::Int@. Violation of
--- this condition is not detected and if the size limit is exceeded, its
--- behaviour is undefined.
---
--- Operation comments contain the operation time complexity in
--- the Big-O notation (<http://en.wikipedia.org/wiki/Big_O_notation>).
+-- The 'Map' type is shared between the lazy and strict modules,
+-- meaning that the same 'Map' value can be passed to functions in
+-- both modules (although that is rarely needed).
```
treeowl: I wouldn't make the claim that it's rarely needed. I don't actually know how often it's needed. This paragraph should be merged with the one above about `Functor` et al.

m-renaud:
Also, thoughts on changing the Data.Map [docs](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Map.html) to simply say that this is a compatibility layer and you should use either Data.Map.Strict or Data.Map.Lazy?

treeowl:
Let's think about Data.Map when we're not racing the clock to release.

m-renaud:
SG, let me know if there's any other changes you'd like. If there are none feel free to merge, or I can do it as well.

Data/Map/Strict.hs
```diff
@@ -15,19 +15,63 @@
...
--- The 'Map' type is shared between the lazy and strict modules,
--- meaning that the same 'Map' value can be passed to functions in
--- both modules (although that is rarely needed).
+-- = Finite Maps (strict interface)
 --
--- These modules are intended to be imported qualified, to avoid name
--- clashes with Prelude functions, e.g.
+-- The @'Map' k v@ type represents a finite map (sometimes called a dictionary)
+-- from keys of type @k@ to values of type @v@. A 'Map' is strict in its keys and
+-- values.
```
treeowl: No. This paragraph is just wrong. A `Map` is *lazy* in its values. Please don't say that it's strict in its values, although it's tempting to imagine that when importing only the `.Strict` module. The part below should say something like

```haskell
-- Each function in this module is careful to force values before installing them in a 'Map'. This
-- is usually more efficient when laziness is not necessary. When laziness *is* required, use
-- the functions in "Data.Map.Lazy".
--
-- In particular, the functions in this module obey the following law:
--
--     If all values stored in all maps in the arguments are in WHNF, then all values stored
--     in all maps in the results will be in WHNF once those maps are evaluated.
--
-- Examples:
--
-- @
-- 'insert' k undefined m == undefined
-- 'insert' 1 'a' $ "Data.Map.Lazy".'Data.Map.Lazy.fromList' ((2,undefined) : zip [3 .. 9] ['b'..])
--   == "Data.Map.Lazy".'Data.Map.Lazy.fromList' ((1,'a') : (2, undefined) : zip [3..9] ['b'..])
-- 'union' ('fromList' $ zip [1..10] ['a'..]) ("Data.Map.Lazy".'Data.Map.Lazy.fromList' $ zip [1..10] (repeat undefined)) == 'fromList' $ zip [1..10] ['a'..])
-- 'unionWith' (\_ _ -> undefined) m1 m2 == undefined
-- 'alterF' (\_ -> pure (Just undefined)) k m == pure undefined
-- @
```

Data/Map/Strict.hs
```diff
@@ -15,19 +15,63 @@
...
+--
+-- The size of a 'Map' must not exceed @maxBound::Int@. Violation of this
+-- condition is not detected and if the size limit is exceeded, its behaviour is
+-- undefined.
+--
+-- The 'Map' type is shared between the lazy and strict modules, meaning that
+-- the same 'Map' value can be passed to functions in both modules. This means
+-- that the 'Functor', 'Traversable' and 'Data' instances are the same as for
+-- the "Data.Map.Lazy" module, so if they are used on strict maps, the resulting
+-- maps will be lazy.
```
treeowl: I know you didn't write this, but replace "will be lazy" by something like "may contain suspended values (thunks)."

m-renaud:
The examples seem like a lot of noise for an introductory section imho. I'm wondering if moving the examples to the implementation section would be cleaner.

m-renaud:
[here](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Map-Strict.html) are the updated rendered docs.

treeowl:
It's not really about implementation. But we can leave them out for now if
you like, or hide them.

m-renaud:
I think its best to remove them for now until we figure out the best place to put them. If we weren't rushing to cut a release I'd spend some time now but I would like to get the IntSet and IntMap docs finished before the release which I think is higher priority.

treeowl:
Sure. Is this ready to be reviewed again?

m-renaud:
Yup, see [updated rendered docs](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Map-Strict.html) (may take a minute or two to rebuild).

Issue #499, opened 2018-01-20
Add sortOn for Data.Sequence


treeowl:
```haskell
sortOn, unstableSortOn :: Ord b => (a -> b) -> Seq a -> Seq a
sortOn f xs = fmap fst . sortBy (compare `on` snd) $ fmap (\x -> (x, f x)) xs
unstableSortOn f xs = fmap fst . unstableSortBy (compare `on` snd) $ fmap (\x -> (x, f x)) xs
```

But it should be quite a bit faster to build these directly over the heap sort instead of mapping twice. @oisdk, do you think you could write these up real quick, or should I give it a go?

oisdk:
I can try it! I should be able to use the standard functions for generating the heap

oisdk:
If we apply these functions while we build the heap, would that expose the opportunity for fusion in cases like:

```haskell
sort . fmap f
fmap f . sort
```

And so on? I could imagine a rule being something like:

```haskell
sortBy cmp (fmap f (Seq xs)) =
    maybe
        (Seq EmptyT)
        (execState (replicateA (size xs) (popMinS cmp)))
        (toPQS cmp (\s (Elem x) -> PQS s (f x) Nl) 0 xs)
```

And on the other side something like:

```haskell
fmap f (sortBy cmp (Seq xs)) =
    maybe
        (Seq EmptyT)
        (execState (replicateA (size xs) (fmap f (popMinS cmp))))
        (toPQS cmp (\s (Elem x) -> PQS s x Nl) 0 xs)
```

treeowl:
Sounds plausible. Be sure to use fmapSeq in your rules, and use a phased
INLINABLE for the sort. Make sure it doesn't hurt when it doesn't apply.

PR #500, opened 2018-01-21
sortOn functions using same heap as sort and unstableSort

oisdk:
As described in #499


Data/Sequence/Internal.hs
```diff
@@ -4604,7 +4607,22 @@ sortBy cmp (Seq xs) =
...
         (execState (replicateA (size xs) (popMinS cmp)))
-        (toPQS cmp (Seq xs))
+        (toPQS cmp (\s (Elem x) -> PQS s x Nl) 0 xs)
+
+-- | \( O(n \log n) \). `sortOn` sorts the specified `Seq` by comparing
+-- the results of a key function applied to each element. `sortOn f` is
+-- equivalent to `sortBy (comparing f)`, but has the performance
+-- advantage of only evaluating f once for each element in the input
+-- list. This is called the decorate-sort-undecorate paradigm, or
+-- Schwartzian transform.
```
m-renaud: Can you add some example invocations here and for the unstable version below?

Data/Sequence/Internal.hs
```diff
@@ -4624,7 +4642,22 @@ unstableSortBy cmp (Seq xs) =
...
+-- comparing the results of a key function applied to each element.
+-- `unstableSortOn f` is equivalent to `unstableSortBy (comparing f)`,
+-- but has the performance advantage of only evaluating f once for each
+-- element in the input list. This is called the
+-- decorate-sort-undecorate paradigm, or Schwartzian transform.
+unstableSortOn :: Ord b => (a -> b) -> Seq a -> Seq a
+unstableSortOn f (Seq xs) =
+    maybe
+       (Seq EmptyT)
+       (execState (replicateA (size xs) (fmap (\(_ :*: y) -> y) (popMin compareFst))))
```
m-renaud: Any reason you used y in the lambda here but x in the lambda for the stable version?

oisdk: Nope! I'll change it to be consistent.

treeowl:
For didactic reasons, I think you should mention compare `on` instead of
comparing, but others may disagree. It may be worth explaining that sortBy
(compare `on` f) will be faster than sortOn f when f is very cheap (e.g.,
fst or a record selector).

m-renaud:
@treeowl should we add either sortBy or sortOn to the introductory docs? Right now we only have the sort function in the sorting section.

treeowl:
I think that's up to you.

On Jan 21, 2018 11:58 AM, "Matt Renaud" <notifications@github.com> wrote:

@treeowl <https://github.com/treeowl> should we add either sortBy or sortOn
to the introductory docs? Right now we only have the sort function in the
sorting section.

—

<https://github.com/haskell/containers/pull/500#issuecomment-359262927>,
or mute
the thread

Data/Sequence/Internal.hs
```diff
@@ -4611,9 +4611,9 @@ sortBy cmp (Seq xs) =
 
 -- | \( O(n \log n) \). 'sortOn' sorts the specified 'Seq' by comparing
 -- the results of a key function applied to each element. @'sortOn' f@ is
--- equivalent to @'sortBy' ('Data.Ord.comparing' f)@, but has the performance
--- advantage of only evaluating f once for each element in the input
--- list. This is called the decorate-sort-undecorate paradigm, or
+-- equivalent to @'sortBy' ('compare' ``Data.function.on`` f)@, but has the
```
treeowl: `Data.Function` got capitalized wrong.

m-renaud:
I'll have to think about it a bit more, I don't think we need to add them right now.

Data/Sequence/Internal.hs
```diff
@@ -4604,7 +4607,35 @@ sortBy cmp (Seq xs) =
...
+-- If @f@ is very cheap (for example a record selector, or 'fst'),
+-- @'sortBy' ('compare' ``Data.Function.on`` f)@ will be faster than
+-- @'sortOn' f@.
+sortOn :: Ord b => (a -> b) -> Seq a -> Seq a
+sortOn f (Seq xs) =
+    maybe
+       (Seq EmptyT)
+       (execState (replicateA (size xs) (fmap (\(_ :*: x) -> x) (popMinS compareFst))))
+       (toPQS compareFst (\s (Elem x) -> PQS s (f x :*: x) Nl) 0 xs)
+  where
```
treeowl: I think we probably want to use yet another type instead of reusing `PQS`, to reduce the allocation and indirection of the strict pair constructor. Sorry; I know it's a bit annoying.

oisdk: No problem, for the normal PQueue as well?

oisdk: Actually, if that means there's going to be 4 different queues, it might begin to make sense to add functions like:

```haskell
foldDigit :: (b -> b -> b) -> (a -> b) -> Digit a -> b
foldDigit (<+>) = fm where
    fm f (One a) = f a
    fm f (Two a b) = f a <+> f b
    fm f (Three a b c) = f a <+> f b <+> f c
    fm f (Four a b c d) = f a <+> f b <+> f c <+> f d
{-# INLINE foldDigit #-}
```

and

```haskell
{-# SPECIALIZE foldWithIndexDigit :: (b -> b -> b) -> (Int -> Elem y -> b) -> Int -> Digit (Elem y) -> b #-}
{-# SPECIALIZE foldWithIndexDigit :: (b -> b -> b) -> (Int -> Node y -> b) -> Int -> Digit (Node y) -> b #-}
foldWithIndexDigit :: Sized a => (b -> b -> b) -> (Int -> a -> b) -> Int -> Digit a -> b
foldWithIndexDigit _ f !s (One a) = f s a
foldWithIndexDigit (<+>) f s (Two a b) = f s a <+> f sPsa b
  where
    !sPsa = s + size a
foldWithIndexDigit (<+>) f s (Three a b c) = f s a <+> f sPsa b <+> f sPsab c
  where
    !sPsa = s + size a
    !sPsab = sPsa + size b
foldWithIndexDigit (<+>) f s (Four a b c d) =
    f s a <+> f sPsa b <+> f sPsab c <+> f sPsabc d
  where
    !sPsa = s + size a
    !sPsab = sPsa + size b
    !sPsabc = sPsab + size c
```

And use them in the `toPQ` functions, as well as `foldMap` and `foldMapWithIndex`. What do you think?

treeowl:
I think you have good judgement and you're good at benchmarking. :-)

treeowl:
Another thing: sortOn and unstableSortOn probably should not themselves be
strict in the sequence elements. I doubt this matters much in practice, but
I think you should probably use a heap that holds the keys strictly and the
values lazily.

oisdk:
Not sure how to structure the new module, I've tried to document it so it shows up in the Haddocks properly

oisdk:
Nope, no `git mv`. On which file exactly? I'm not sure how I would do it for `Sorting.hs`, but the `sorting.md` file had its move properly recorded in the history.

Data/Sequence/Internal/sorting.md
```diff
@@ -0,0 +1,252 @@
...
+
+## sortOn Functions
+
+The `sortOn` and `unstableSortOn` functions perform the Schwartzian transform, however instead of the following implementation:
+
+```haskell
+sortOn f = fmap snd . sortBy (conparing fst) . fmap (\x -> (f x, x))
+```
+
+The `fmap`s are fused manually with the creation of the queue, avoiding the two extra traversals. It still suffers a slowdown of roughly 35%:
```
treeowl: Is it still that bad now that you've specialized the heap types?

tests/seq-properties.hs
```diff
@@ -540,6 +542,10 @@ prop_sortBy xs =
     toList' (sortBy f xs) ~= Data.List.sortBy f (toList xs)
   where f (x1, _) (x2, _) = compare x1 x2
 
+prop_sortOn :: Seq (OrdA, B) -> Bool
```
treeowl: Why not `Fun A OrdB -> Seq A -> Bool`? That seems to get to the point more obviously.

Data/Sequence/Internal/Sorting.hs
```diff
@@ -0,0 +1,417 @@
...
+--
+-- All of the functions fail on an empty queue.
+--
+-- Each of these functions contains an "unroll" function, which is
+-- called like this:
+--
+-- @unroll (Q x ts) = (mergeQs ts, x)@
+--
+-- The reason the call to @mergeQs@ is lazy is that it will be bottom
+-- the last element in the queue, preventing us from evaluating the
```
treeowl: Will be bottom *for*, or some such.

treeowl:
Oh, but you still need to split out `State`. Also, you need to add the new modules to the `.cabal` file!

oisdk:
You can go ahead and squash the whole thing, by the way! (I don't know how to squash from my end, not sure if I'm able to)

treeowl:
Thanks, @oisdk! Great work!

PR #502, opened 2018-01-22
Improve Data.Set module docs.

m-renaud:
This improves the module level docs for Data.Set following the form of the recent improvements to Data.Sequence. This is for the most part a re-organization of the existing content with a few minor changes so should hopefully be uncontroversial.

Rendered docs can be found [here](http://www.m-renaud.com/haddocks/containers-0.5.10.2/Data-Set.html).

Partially addresses #495.

### DIFF
--- a/Data/Set.hs
+++ b/Data/Set.hs
@@ -16,9 +16,9 @@
 --
 -- = Finite Sets
 --
--- The @'Set' e@ type represents a set of elements of type @e@. A 'Set' is
--- strict in the element type and most operations require that @e@ be an
--- instance of the 'Ord' class.
+-- The @'Set' e@ type represents a set of elements of type @e@. Most operations
+-- require that @e@ be an instance of the 'Ord' class. A 'Set' is strict in its
+-- elements.
 --
 -- For a walkthrough of the most commonly used functions see the
 -- <https://haskell-containers.readthedocs.io/en/latest/set.html sets introduction>.
--- a/Data/Set.hs
+++ b/Data/Set.hs
@@ -14,36 +14,26 @@
 -- Portability :  portable
 --
 --
--- = Sets
+-- = Finite Sets
 --
--- The @'Set' e@ type represents a set of /unique/, /ordered/ elements of type
--- @e@.
+-- The @'Set' e@ type represents a set of elements of type @e@. A 'Set' is
+-- strict in the element type and most operations require that @e@ be an
+-- instance of the 'Ord' class.
 --
 -- For a walkthrough of the most commonly used functions see the
 -- <https://haskell-containers.readthedocs.io/en/latest/set.html sets introduction>.
 --
--- Note that the implementation is /left-biased/ -- the elements of a
--- first argument are always preferred to the second, for example in
--- 'union' or 'insert'.  Of course, left-biasing can only be observed
--- when equality is an equivalence relation instead of structural
--- equality.
+-- Note that the implementation is generally /left-biased/. Functions that take
+-- two sets as arguments and combine them, such as `union` and `intersection`,
+-- prefer the entries in the first argument to those in the second. Of course,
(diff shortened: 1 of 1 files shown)


Data/Set.hs
```diff
@@ -13,14 +13,48 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
--- An efficient implementation of sets.
+--
+-- = Sets
+--
+-- The @'Set' e@ type represents a set of /unique/, /ordered/ elements of type
+-- @e@.
```
treeowl: What are ordered elements? What is a unique element? I think perhaps

    -- The @'Set' e@ type represents a set of elements of type @e@. Most operations
    -- require that @e@ be an instance of the 'Ord' class.

Data/Set.hs
```diff
@@ -13,14 +13,48 @@
...
+--
+-- For a walkthrough of the most commonly used functions see the
+-- <https://haskell-containers.readthedocs.io/en/latest/set.html sets introduction>.
+--
+-- Note that the implementation is /left-biased/ -- the elements of a
+-- first argument are always preferred to the second, for example in
+-- 'union' or 'insert'.  Of course, left-biasing can only be observed
+-- when equality is an equivalence relation instead of structural
+-- equality.
```
treeowl: See my note on `Map`. Also, rather than "left-biasing", how about "this bias"?

Data/Set.hs
```diff
@@ -13,14 +13,48 @@
...
+--
+-- == Strictness properties
+--
+-- This module satisfies the following strictness property:
+--
+-- * Key arguments are evaluated to WHNF
+--
+-- Here are some examples that illustrate the property:
+--
+-- > delete undefined s  ==  undefined
```
treeowl: This demonstrates nothing. How about

```haskell
-- > insert undefined xs == undefined
-- > singleton undefined = undefined
```

Data/Set.hs
m-renaud: Done.

Data/Set.hs
m-renaud: Done.

Data/Set.hs
m-renaud: Done.

treeowl:
I don't see the update.

m-renaud:
Oops, accidentally sent the review comments before pushing the commit. I thought I had "started a review" which I was going to finish after the commit went up. I'll have it up shortly, about to walk to work.

m-renaud:
Changes pushed, Haddocks haven't been pushed again because its a pain to get both changes built simultaneously because they're in different branches.

Data/Set.hs
```diff
@@ -13,14 +13,38 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
--- An efficient implementation of sets.
+--
+-- = Finite Sets
+--
+-- The @'Set' e@ type represents a set of elements of type @e@. A 'Set' is
+-- strict in the element type and most operations require that @e@ be an
```
treeowl: Surely you mean strict in its *elements*, not their *type*. I think you should move the strictness to a new sentence after the one about `Ord`.

m-renaud: Oops, yup that what I meant.

m-renaud: Maybe this should just be a bullet list, it may read better since both the sentences are so short. WDYT?

treeowl:
I don't think bullets will help.

m-renaud:
How about: "Most operations require that @e@ be an instance of the 'Ord' class, and all operations are strict in the element."

treeowl:
Technically, I suppose that's true. But it's much more intuitively helpful,
I believe, to think of Set as a strict data structure.

m-renaud:
Gotcha, I'll leave it as is. Are there any other changes you'd like made or can this be merged?

PR #505, opened 2018-01-22
Fix power of 2 validity check.

m-renaud:
Previously an incorrect check of divisible by 2 was being performed, this fixes
the existing behaviour for `IntSet` and adds the check to `IntMap` as wel.


treeowl:
We need this for `IntSet` too. Joachim withdrew his PR so you could do it all in one.

Issue #509, opened 2018-01-23
Test sort stability


treeowl:
`sort`, `sortBy`, and `sortOn` are all supposed to be stable sorts, but we only test stability for `sortBy`.

treeowl:
I'm pretty sure there is a type defined somewhere in the set/map tests with a sloppy `Ord` instance specifically designed for these sorts of tests.

jwaldmann:
bit late to the party (just saw the note in HWN), but .. I'd do it like this, where
* "hiding" is a stand-alone concept (not fused with "pairing")
* we check for monotonicity of the output only (assuming that it's checked elsewhere that the function returns the correct multiset)

```
{-# language RankNTypes #-}

import Test.LeanCheck

import qualified Data.Sequence as S
import Data.Foldable (toList)

newtype Hide t = Hide { reveal :: t } deriving Show

instance Eq (Hide t) where _ == _ = True
instance Ord (Hide t) where compare _ _  = EQ

sort_is_stable :: (forall a . Ord a => S.Seq a -> S.Seq a) -> [Int ] -> Bool
sort_is_stable  sort xs =
  let input = zip xs $ map Hide [ 0 :: Int .. ]
      output = toList $ sort $ S.fromList input
  in  monotone $ map ( \(x,y) -> (x, reveal y) ) output

monotone xs = and $ zipWith (<=) xs $ tail xs

main = do
  checkFor 10000 $ sort_is_stable S.sort
  checkFor 10000 $ sort_is_stable S.unstableSort
```

treeowl:
That's too coarse.

jwaldmann:
Out of curiosity - what is the exact definition of "stability" that these sort functions realize?

This definition could be put in the haddocks, so programmers know exactly what to expect. 

Here is one example of a formal definition of "stability w.r.t. a key function" http://dx.doi.org/10.1007/s10817-012-9260-7 (Def. 4)

treeowl:
Fixed in #514

treeowl:
@jwaldmann, I'm not sure we want to go into so much detail about a relatively well-known idea. Can you suggest something brief and understandable?

PR #514, opened 2018-01-25
test stability of sort and sortOn

oisdk:
as described in #509

### DIFF
--- a/tests/seq-properties.hs
+++ b/tests/seq-properties.hs
@@ -568,7 +568,7 @@ data UnstableOrd = UnstableOrd
     } deriving (Show)
 
 instance Eq UnstableOrd where
-    UnstableOrd x _ == UnstableOrd y _ = x == y
+    x == y = compare x y == EQ
 
 instance Ord UnstableOrd where
     compare (UnstableOrd x _) (UnstableOrd y _) = compare x y
--- a/tests/seq-properties.hs
+++ b/tests/seq-properties.hs
@@ -601,12 +601,10 @@ prop_sortOn (Fun _ f) xs =
     listSortOn k = Data.List.sortBy (compare `on` k)
 #endif
 
-prop_sortOnStable :: Seq UnstableOrd -> Bool
-prop_sortOnStable xs =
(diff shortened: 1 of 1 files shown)


tests/seq-properties.hs
```diff
@@ -560,6 +562,23 @@ prop_sort :: Seq OrdA -> Bool
 prop_sort xs =
     toList' (sort xs) ~= Data.List.sort (toList xs)
 
+data UnstableOrd = UnstableOrd
+    { ordKey :: OrdA
+    , _val :: A
```
treeowl: Maybe `_ignored`?

tests/seq-properties.hs
```diff
@@ -560,6 +562,23 @@ prop_sort :: Seq OrdA -> Bool
...
+    } deriving (Eq,Show)
```
treeowl: No, definitely don't derive `Eq`!!! It must match the `Ord` instance! You'll need to `fmap` an `unignore` function over each of the results in the tests.

treeowl:
My reasoning: your current implementation may not use the `Eq` instance, but that's by no means guaranteed. And relying on implementation details in testing makes it really annoying for someone reading the test to know if it's right.

oisdk:
Yup, makes sense!

tests/seq-properties.hs
```diff
@@ -596,7 +603,8 @@ prop_sortOn (Fun _ f) xs =
 
 prop_sortOnStable :: Seq UnstableOrd -> Bool
 prop_sortOnStable xs =
-    toList' (sortOn ordKey xs) ~= listSortOn ordKey (toList xs)
+    (fmap . fmap) unignore (toList' (sortOn ordKey xs)) ~=
+    fmap unignore (listSortOn ordKey (toList xs))
```
treeowl: No, this is wrong. You want

```haskell
Fun A UnstableOrd -> Seq A -> Bool
```

This means `UnstableOrd` will need an `Arbitrary` instance, which should be pretty easy to base on the one for pairs.

treeowl:
Thinking more carefully, we probably don't actually need a separate stability test for `sortOn` after all. But having it does very little harm.

PR #515, opened 2018-01-25
add nubOrd and friends

### LINKED ISSUE #439: add ordNub somewhere in containers
After dropping the standard ordNub (i.e. this: https://github.com/nh2/haskell-ordnub) code into a codebase for the zillionth time I went and checked the history of why it never made it into a standard lib. It can't go into base because it requires containers or reimplementing `Set` just for this purpose, which is silly. It doesn't necessarily belong in the `Data.Set` module. But perhaps it could fall into a new module, like `Data.SetUtils` or `Data.ContainerUtils` or the like? Anyway, this package is the most natural place for it to live, even if it doesn't quite fit, and everyone ends up using it everywhere, so it would be really nice to just have a standard for it...

gbaz:
resolves #439

I think I have the asymptotics for `nubInt` right, but a sanity check wouldn't hurt...

### DIFF
--- a/Data/Containers/ListUtils.hs
+++ b/Data/Containers/ListUtils.hs
@@ -1,0 +1,65 @@
+{-# LANGUAGE CPP #-}
+#if !defined(TESTING) && __GLASGOW_HASKELL__ >= 703
+{-# LANGUAGE Safe #-}
+#endif
+
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Data.Containers.ListUtils
+-- Copyright   :  (c) Gershom Bazerman 2018
+-- License     :  BSD-style
+-- Maintainer  :  libraries@haskell.org
+-- Portability :  portable
+--
+-- This module provides efficient containers-based functions on the list type.
+-----------------------------------------------------------------------------
+
+module Data.Containers.ListUtils (
+       nubOrd,
+       nubOrdOn,
+       nubInt,
+       nubIntOn
+       ) where
+
+import qualified Data.Set as Set
+import qualified Data.IntSet as IntSet
+
+-- | /O(n log n)/. The 'nubOrd' function removes duplicate elements from a list.
+-- In particular, it keeps only the first occurrence of each element. By using a 'Set' internally
+-- it has better asymptotics than the standard 'nub' function.
+nubOrd :: (Ord a) => [a] -> [a]
+nubOrd l = go Set.empty l
+  where
+    go _ [] = []
+    go s (x:xs) = if x `Set.member` s then go s xs
+                                      else x : go (Set.insert x s) xs
+
+-- | The `nubOrdOn` function behaves just like `nubOrd` except it preforms comparisons not on the
+-- original datatype, but a user-specified projection from that datatype.
+nubOrdOn :: (Ord b) => (a -> b) -> [a] -> [a]
+nubOrdOn f l = go Set.empty l
+  where
+    go _ [] = []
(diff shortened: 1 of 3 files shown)


treeowl:
Could you add QuickCheck properties?

Data/Containers/ListUtils.hs
```diff
@@ -0,0 +1,65 @@
...
+-- In particular, it keeps only the first occurrence of each element. By using a 'Set' internally
+-- it has better asymptotics than the standard 'nub' function.
+nubOrd :: (Ord a) => [a] -> [a]
+nubOrd l = go Set.empty l
+  where
+    go _ [] = []
+    go s (x:xs) = if x `Set.member` s then go s xs
+                                      else x : go (Set.insert x s) xs
+
+-- | The `nubOrdOn` function behaves just like `nubOrd` except it preforms comparisons not on the
```
treeowl: Performs, not preforms.

Data/Containers/ListUtils.hs
```diff
@@ -0,0 +1,65 @@
...
+-- it has better asymptotics than the standard 'nub' function.
+nubInt :: [Int] -> [Int]
+nubInt l = go IntSet.empty l
+  where
+    go _ [] = []
+    go s (x:xs) = if x `IntSet.member` s then go s xs
+                                         else x : go (IntSet.insert x s) xs
+
+-- | The `nubIntOn` function behaves just like 'nubInt' except it preforms comparisons not on the
+-- original datatype, but a user-specified projection from that datatype to 'Int'.
```
treeowl: Performs, not preforms.

Data/Containers/ListUtils.hs
```diff
@@ -0,0 +1,65 @@
...
+nubIntOn :: (a -> Int) -> [a] -> [a]
+nubIntOn f l = go IntSet.empty l
```
treeowl: Eta reduce. And it's probably slightly faster to make `f` an argument to `go`, here and above.

treeowl:
Actually, I changed my mind. We want to let `f` have the chance to inline, so keep it pulled out.

treeowl:
However however, inlining `f` is only useful if we make this participate in list fusion. So let's do that too.

gbaz:
Added tests, fixed typo, eta reduced.

I did not add fusion. I took a run at it and it wasn't straightforward. Doing a foldr doesn't really work because it reverses the direction in which things are dropped. So that means making it a good consumer is hard. Further, the way in which we end up building out of the fold (which involves taking a snd of the result, since we need to pass around a pair) is rather ugly and I think gets in the way of making it a good producer.

treeowl:
You can definitely make it a good consumer nicely.

nubOrd xs = foldr go stop xs S.empty

gbaz:
I don't know what that means. In any case, I tried, and didn't get it working. If you know how to get it working, I suggest we merge first, and then you make a further patch.

treeowl:
Sure.

On Jan 26, 2018 1:46 AM, "gbaz" <notifications@github.com> wrote:

I don't know what that means. In any case, I tried, and didn't get it
working. If you know how to get it working, I suggest we merge first, and
then you make a further patch.

—
You are receiving this because you commented.

<https://github.com/haskell/containers/pull/515#issuecomment-360697951>,
or mute
the thread

gbaz:
Ok, is this good to merge then?

treeowl:
Thanks.

PR #516, opened 2018-01-26
Faster traverse

oisdk:
So I applied the same techniques in the folds to the traverse, and I got similar speedups. On the multiplyup benchmark: 

```haskell
multiplyUp :: Seq.Seq Int -> Seq.Seq Int
multiplyUp = snd . flip runState 0 . traverse go where
  go x = do
    s <- State (\x -> (x + 1,x))
    return (s * x)
```

On a sequence of length 500000:

```
benchmarking 500000/multiplyUp/new
time                 137.6 ms   (129.7 ms .. 145.7 ms)
                     0.996 R²   (0.990 R² .. 1.000 R²)
mean                 138.8 ms   (136.4 ms .. 141.4 ms)
std dev              3.470 ms   (2.492 ms .. 4.484 ms)
variance introduced by outliers: 12% (moderately inflated)
             
benchmarking 500000/multiplyUp/old
time                 355.1 ms   (281.2 ms .. 407.7 ms)
                     0.995 R²   (0.983 R² .. 1.000 R²)
mean                 339.1 ms   (324.9 ms .. 346.0 ms)
std dev              12.30 ms   (0.0 s .. 12.63 ms)
variance introduced by outliers: 19% (moderately inflated)
```

And even old functions which use traverse internally get a speedup. `scanl (+) 0` for instance:

```
benchmarking 500000/scanl/new
time                 134.5 ms   (121.4 ms .. 156.0 ms)
                     0.979 R²   (0.945 R² .. 0.998 R²)
mean                 141.0 ms   (134.2 ms .. 148.6 ms)
std dev              9.440 ms   (6.655 ms .. 12.88 ms)
             
benchmarking 500000/scanl/old
time                 373.0 ms   (253.8 ms .. 471.4 ms)
                     0.988 R²   (0.957 R² .. 1.000 R²)
mean                 379.5 ms   (356.0 ms .. 396.6 ms)
std dev              25.81 ms   (0.0 s .. 29.52 ms)
```

The inlining of the inner functions also allows us to avoid coercion around `Elem`. The code is *much* longer, though.

### DIFF
--- a/Data/Sequence/Internal.hs
+++ b/Data/Sequence/Internal.hs
@@ -428,37 +428,7 @@ instance Traversable Seq where
 #if __GLASGOW_HASKELL__
     {-# INLINABLE traverse #-}
 #endif
-    traverse _ (Seq EmptyT) = pure (Seq EmptyT)
-    traverse f' (Seq (Single (Elem x'))) =
-        (\x'' -> Seq (Single (Elem x''))) <$> f' x'
-    traverse f' (Seq (Deep s' pr' m' sf')) =
-        liftA3
-            (\pr'' m'' sf'' -> Seq (Deep s' pr'' m'' sf''))
-            (traverseDigitE f' pr')
-            (traverseTree (traverseNodeE f') m')
-            (traverseDigitE f' sf')
-      where
-        traverseTree
-            :: Applicative f
-            => (Node a -> f (Node b))
-            -> FingerTree (Node a)
-            -> f (FingerTree (Node b))
-        traverseTree _ EmptyT = pure EmptyT
-        traverseTree f (Single x) = Single <$> f x
-        traverseTree f (Deep s pr m sf) =
-            liftA3
-                (Deep s)
-                (traverse f pr)
-                (traverseTree (traverse f) m)
-                (traverse f sf)
-        traverseDigitE
-            :: Applicative f
-            => (a -> f b) -> Digit (Elem a) -> f (Digit (Elem b))
-        traverseDigitE f t = traverse (traverse f) t
-        traverseNodeE
-            :: Applicative f
-            => (a -> f b) -> Node (Elem a) -> f (Node (Elem b))
-        traverseNodeE f t = traverse (traverse f) t
+    traverse f (Seq xs) = Seq <$> traverse (traverse f) xs
 
 instance NFData a => NFData (Seq a) where
     rnf (Seq xs) = rnf xs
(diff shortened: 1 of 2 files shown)


treeowl:
There seems to be some verbosity inherent in efficient traversal, thanks to the limitations of the type role system.

Data/Sequence/Internal.hs
```diff
@@ -424,29 +424,49 @@ instance Foldable Seq where
...
+        traverseTree f (Single x) = Single <$> f x
+        traverseTree f (Deep s pr m sf) =
+            liftA3
+                (Deep s)
+                (traverseDigitN f pr)
+                (traverseTree (traverseNodeN f) m)
+                (traverseDigitN f sf)
+        traverseDigitE
+            :: Applicative f
+            => (a -> f b) -> Digit (Elem a) -> f (Digit (Elem b))
```
treeowl: What happens if you generalize these `E` functions and pull them to the top? Do they get slow again? That is, something like

```haskell
traverseDigitE
     :: (Applicative f, Traversable t)
     => (a -> f b) -> Digit (t a) -> f (Digit (t b))
```

Just a thought to try to disentangle `FingerTree` from `Seq`.

oisdk: That slows it down again, but I may be able to figure out a FingerTree instance which keeps the speed isn't to do with `Elem`

treeowl:
Don't worry too much about it. traverse can be a weirdo. We should try to
cut down on it where we don't need it though.

oisdk:
Yeah, so this version delegates to an `Elem`-agnostic traverse on `FingerTree`, and it seems like the extra `fmap`s are elided, because it has the same performance on the benchmarks

treeowl:
I'm not sure I quite buy it. What if that final `fmap` is expensive?

treeowl:
I'm concerned about traversing all those `Elem`s, actually. Let's not do
that.

oisdk:
Yeah, I agree. The extra `fmap`s were making me nervous (although I couldn't find a test to make them slower

treeowl:
Hey, did you happen to check out whether `traverseWithIndex` needs any similar attention?

oisdk:
I'm not sure—I *think* I did. I'll check later today, but I'd imagine it's fine: [this](https://github.com/haskell/containers/blob/fd1a7d351ba528dfcada745a1b8167642d599602/Data/Sequence/Internal.hs#L2907) specialisation is effectively doing the same as in `traverse`.

PR #517, opened 2018-01-28
Add RULES for nub functions

treeowl:
* Add rewrite rules to allow the nub functions in `ListUtils`
  to participate in fold/build fusion.

* For the sake of simplicity, define `nubOrd` and `nubInt` in terms
  of `nubOrdOn` and `nubIntOn`.


treeowl:
@gbaz, you may find this interesting. Compare the rules here to the ones for `scanl` in `GHC.List`. I've changed the strictness a bit to make it more consistent.

treeowl:
@gbaz, you may also find my changes to the tests interesting. I always like to start with tests that obviously get to the point, and go for more carefully constructed/detailed ones if I have a specific reason.

gbaz:
Nice!

PR #519, opened 2018-01-30
Reorganize Map and IntMap Haddocks.

m-renaud:
**Changes:**

- Move "construction" section to the top
- Include fromList functions in "construction" section
- Remove "operators" section, you don't go "looking for operators", you go
  "looking for functionality" which may be available as an operator.
- Move most common query operations (lookup and friends) to the top
- Move partial functions to the bottom of sections.

**Rationale:**

The other approach I considered was that [employed by Data.HashMap](https://hackage.haskell.org/package/unordered-containers-0.2.8.0/docs/Data-HashMap-Strict.html#g:3) which groups the "basic interface" API functions together. I didn't use that approach because I feel it defeats the purpose of an API reference, specifically that all related functions are grouped together. By separating out the "basic interface" from the "advanced interface" you now need to look at two places if you're looking for a particular function. Instead a introduction and tutorial (a la https://haskell-containers.readthedocs.io) is a more appropriate place to present the basic interface in my opinion.

### DIFF
--- a/Data/Map/Lazy.hs
+++ b/Data/Map/Lazy.hs
@@ -91,11 +91,20 @@ module Data.Map.Lazy (
     -- * Construction
     , empty
     , singleton
+    , fromSet
+
+    -- ** From Lists
     , fromList
     , fromListWith
     , fromListWithKey
-
-    -- * Operators
+    , fromAscList
+    , fromAscListWith
+    , fromAscListWithKey
+    , fromDistinctAscList
+    , fromDescList
+    , fromDescListWith
+    , fromDescListWithKey
+    , fromDistinctDescList
 
     -- * Query
     , lookup
@@ -185,7 +194,6 @@ module Data.Map.Lazy (
     , keys
     , assocs
     , keysSet
-    , fromSet
 
     -- ** Lists
     , toList
@@ -193,14 +201,6 @@ module Data.Map.Lazy (
     -- ** Ordered lists
     , toAscList
     , toDescList
-    , fromAscList
-    , fromAscListWith
-    , fromAscListWithKey
-    , fromDistinctAscList
-    , fromDescList
-    , fromDescListWith
-    , fromDescListWithKey
-    , fromDistinctDescList
 
     -- * Filter
     , filter
--- a/Data/Map/Lazy.hs
+++ b/Data/Map/Lazy.hs
@@ -110,6 +110,22 @@ module Data.Map.Lazy (
     , fromDescListWithKey
     , fromDistinctDescList
 
+    -- * Insertion
+    , insert
+    , insertWith
+    , insertWithKey
+    , insertLookupWithKey
+
+    -- * Deletion\/Update
+    , delete
+    , adjust
+    , adjustWithKey
+    , update
+    , updateWithKey
+    , updateLookupWithKey
+    , alter
+    , alterF
+
     -- * Query
     , lookup
     , (!?)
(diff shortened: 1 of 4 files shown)


treeowl:
No need to request my review unless it's been a while. I look at every issue and PR.

m-renaud:
Habit from work :)

Data/Map/Lazy.hs
```diff
@@ -88,24 +88,28 @@ module Data.Map.Lazy (
...
     , lookupGT
     , lookupLE
     , lookupGE
-
-    -- * Construction
-    , empty
-    , singleton
+    , null
+    , size
+    , (!)
```
treeowl: I don't like partial functions, but `!`really belongs in the lookup zone.

Data/Map/Lazy.hs
```diff
@@ -184,9 +189,6 @@ module Data.Map.Lazy (
 
     -- ** Lists
     , toList
-    , fromList
-    , fromListWith
-    , fromListWithKey
```
treeowl: Hrmmmm.. Maybe.

Data/Map/Lazy.hs
m-renaud: Wait, isn't that where it is? The construction heading moved elsewhere so it's under the query heading. The interleaved diffs are confusing sometimes. Unless I've completely misinterpreted your comment :P

Data/Map/Lazy.hs
m-renaud: I thought these fit more with construction because they allow you to build a map from some other value. I think of conversions as converting a map to some other value, the opposite of construction.

m-renaud:
There's also the more specialized `fromAscList` and friends that are currently in the "Conversion > Ordered lists" section which feel like they should also be in the "Construction" section. WDYT?

m-renaud:
This is ready for another review, PTAL.

Data/Map/Lazy.hs
```diff
@@ -88,24 +88,37 @@ module Data.Map.Lazy (
...
     Map              -- instance Eq,Show,Read
 
-    -- * Operators
-    , (!), (!?), (\\)
+    -- * Construction
+    , empty
+    , singleton
+    , fromSet
+
+    -- ** From Lists
```
treeowl: Why are these in a separate section here but under construction for sets? One way or the other, I think they should be consistent.

Data/Map/Lazy.hs
```diff
@@ -88,24 +88,37 @@ module Data.Map.Lazy (
...
+    , fromListWith
+    , fromListWithKey
+    , fromAscList
+    , fromAscListWith
+    , fromAscListWithKey
+    , fromDistinctAscList
+    , fromDescList
+    , fromDescListWith
+    , fromDescListWithKey
+    , fromDistinctDescList
```
treeowl: Let's have three subsections for un-ordered, ascending, and descending.

Data/Map/Lazy.hs
treeowl: I want `(!)` adjacent to `(!?)`.

m-renaud: Done.

Data/Map/Lazy.hs
m-renaud: I went back and fourth with this a few times. The number of fromList functions for sets is relatively small but maps have A LOT! I would argue that some of these are so specialized (`fromDistinctAscKey` and friends) that they don't belong in the core containers package at all and would be better hosted in containers-extra, but that's a separate discussion :)

I've taken your suggestion below and added sections for the different fromList variants.

Data/Map/Lazy.hs
m-renaud: Done.

m-renaud:
Any last comments? If not I'd like to merge this in today, thanks!

treeowl:
I kind of want to split the Query section into two subsections. Maybe something like "Lookup" and "Size"? I'm not sure about the names, but `lookup` is *very* different from `null`.

m-renaud:
Done.

treeowl:
Cool. Should be good then!

PR #521, opened 2018-01-30
Remove "strict map" wording from docs

m-renaud:
There's no such things as a "strict map", only a Strict API which forces evaluation of values before inserting them into the map.

[ci skip]


m-renaud:
Likely non-controversial, wanted to put this up in case there were other small changes to make in the docs that could be included in the PR.

PR #527, opened 2018-02-06
Unbox the `SetM` array

treeowl:
`Data.Graph` uses a mutable array to track which vertices have
been visited. This was implemented using an `STArray s Vertex Bool`.
This is obviously bad, because the garbage collector will have to
trace the array if it's alive during a collection. Using an unboxed
array instead should always be slightly better.

It may be better to go further and use a bit array to save space,
but I'm not entirely comfortable making that change until we have
some tests for `Data.Graph`--any mistakes there will likely be
subtle.

### DIFF
--- a/Data/Graph.hs
+++ b/Data/Graph.hs
@@ -72,7 +72,7 @@ module Data.Graph(
 -- Extensions
 #if USE_ST_MONAD
 import Control.Monad.ST
-import Data.Array.ST (STArray, newArray, readArray, writeArray)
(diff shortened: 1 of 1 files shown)


treeowl:
Ah, actually, I see that we can just use an unboxed array of `Bool`, and that should do the right thing... What was I thinking.

treeowl:
Why won't GitHub let me reopen this?

treeowl:
Ah, okay.

PR #528, opened 2018-02-06
Unbox the `SetM` array

treeowl:
`Data.Graph` uses a mutable array to track which vertices have
been visited. This was implemented using an `STArray s Vertex Bool`.
This is obviously bad, because the garbage collector will have to
trace the array if it's alive during a collection. Using an unboxed
array instead should always be better. Indeed, doing so will also
dramatically reduce the size of the array, from the number of
vertices in the graph to that divided by the word size. Each array
operation is slightly more complex, but we'll surely win from
memory locality and such anyway.

### DIFF
--- a/Data/Graph.hs
+++ b/Data/Graph.hs
@@ -72,7 +72,7 @@ module Data.Graph(
 -- Extensions
 #if USE_ST_MONAD
 import Control.Monad.ST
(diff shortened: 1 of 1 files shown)


treeowl:
It's too bad we don't have any graph benchmarks whatsoever. I wonder if this will show anything on the gipeda GHC dashboard the next time GHC upgrades containers.

PR #529, opened 2018-02-06
Make Data.Graph use a bit vector

treeowl:
Use a bit vector in `Data.Graph` instead of an array of
`Bool`. This reduces the size of the array from the number
of words the graph has vertices to the number of vertices
divided by the word size.

Is this a good idea? I don't know for sure. If we *don't* do
this, we should *almost certainly* make it use an unboxed array
rather than a boxed one; we don't need the lazness, and we surely
don't need the garbage collector running through arrays of Bool!

### DIFF
--- a/Data/Graph.hs
+++ b/Data/Graph.hs
@@ -72,7 +72,7 @@ module Data.Graph(
 -- Extensions
 #if USE_ST_MONAD
 import Control.Monad.ST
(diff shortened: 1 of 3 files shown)


treeowl:
This turns out to be pretty much worthless because `Data.Array.ST` *already does this* for unboxed arrays of `Bool`. Duh.

PR #530, opened 2018-02-08
Fix IntSet and IntMap validity tests (resolves #522).

### LINKED ISSUE #522: IntSetValidity: commonPrefix too loose
This code:
```
-- Invariant: Prefix is the common high-order bits that all elements share to
--            the left of the Mask bit.
commonPrefix :: IntSet -> Bool
commonPrefix t =
  case t of
    Nil -> True
    Tip _ _ -> True
    b@(Bin p _ _ _) -> all (sharedPrefix p) (elems b)
  where
    sharedPrefix :: Prefix -> Int -> Bool
    sharedPrefix p a = 0 == (p `xor` (p .&. a))
```
is too weak, I believe: It checks only that the outermost prefix is correct, but not of the `Bin`’s inside the left and right tree.

Maybe 
```
-- Invariant: Prefix is the common high-order bits that all elements share to
--            the left of the Mask bit.
commonPrefix :: IntSet -> Bool
commonPrefix t =
  case t of
    Nil -> True
    Tip _ _ -> True
    b@(Bin p _ l r) -> all (sharedPrefix p) (elems b) && commonPrefix l && commonPrefix r
  where
    sharedPrefix :: Prefix -> Int -> Bool
    sharedPrefix p a = 0 == (p `xor` (p .&. a))
```
would be better?

m-renaud:
The previous implementations only checked the `commonPrefix` and `maskRespected`
invariant for the top level Bin constructor and didn't appropriately recurse
into subtrees.

This resolves #522.

### DIFF
--- a/tests/IntMapValidity.hs
+++ b/tests/IntMapValidity.hs
@@ -46,10 +46,10 @@ commonPrefix t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    b@(Bin p _ _ _) -> all (sharedPrefix p) (keys b)
+    b@(Bin p _ l r) -> all (sharedPrefix p) (keys b) && commonPrefix l && commonPrefix r
   where
     sharedPrefix :: Prefix -> Int -> Bool
-    sharedPrefix p a = 0 == (p `xor` (p .&. a))
+    sharedPrefix p a = p == p .&. a
 
 -- Invariant: In Bin prefix mask left right, left consists of the elements that
 --            don't have the mask bit set; right is all the elements that do.
@@ -60,4 +60,6 @@ maskRespected t =
     Tip _ _ -> True
     Bin _ binMask l r ->
       all (\x -> zero x binMask) (keys l) &&
-      all (\x -> not (zero x binMask)) (keys r)
+      all (\x -> not (zero x binMask)) (keys r) &&
+      maskRespected l &&
+      maskRespected r
--- a/tests/IntSetValidity.hs
+++ b/tests/IntSetValidity.hs
@@ -49,10 +49,10 @@ commonPrefix t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    b@(Bin p _ _ _) -> all (sharedPrefix p) (elems b)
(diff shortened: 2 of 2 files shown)


tests/IntMapValidity.hs
```diff
@@ -60,4 +60,6 @@ maskRespected t =
     Tip _ _ -> True
     Bin _ binMask l r ->
       all (\x -> zero x binMask) (keys l) &&
-      all (\x -> not (zero x binMask)) (keys r)
+      all (\x -> not (zero x binMask)) (keys r) &&
+      maskRespected l &&
+      maskRespected r
```
treeowl: This implementation of `maskRespected` is at least quadratic. Can't you get it down to linear?

m-renaud:
I had considered doing that but since this is test only code and I didn't
see any noticable slowdown when running the tests I wasn't sure it was
worth the complication. It may make it less obvious what is being tested if
it's optimized as well. If you'd still like me to optimize it let me know.

treeowl:
You're right that getting `O(n)` will be too complicated/too far removed from the basic operations. But I think a small `O(W*n)` should be pretty reasonable. Does this look right?

```haskell
 maskRespected :: IntMap a -> Bool
 maskRespected = go [] [] where
   go should_zero shouldnt_zero t =
     case t of
       Nil -> True
       Tip k _ -> all (zero k) should_zero && all (not . zero k) shouldnt_zero
       Bin _ binMask l r ->
         go (binMask : should_zero) shouldnt_zero l &&
         go should_zero (binMask : shouldnt_zero) r
```

treeowl:
Actually, I may have been wrong about your version being quadratic....

treeowl:
Yeah, your version is probably not actually as bad as I thought. Sorry.

m-renaud:
No worries, the version you posted is more efficient but not immediately obvious (to me at least) that it's testing the correct thing.

tests/IntMapValidity.hs
m-renaud: That's something we could do, but I'm not sure if it would improve or hurt the readability. It would require fewer traversals of the tree but may make it more difficult to see how each invariant is tested.

PR #531, opened 2018-02-08
Reorganize Set and IntSet Haddocks.

m-renaud:
This moves the Construction section to the top and puts all "construction like"
functions under that heading. Also, moves operators out of their own section and
into the appropriate section.

### DIFF
--- a/Data/IntSet.hs
+++ b/Data/IntSet.hs
@@ -80,24 +80,24 @@ module Data.IntSet (
             , fromAscList
             , fromDistinctAscList
 
-            -- * Queries
-            , IS.null
-            , size
+            -- * Insertion
+            , insert
+
+            -- * Deletion
+            , delete
+
+            -- * Query
             , member
             , notMember
             , lookupLT
             , lookupGT
             , lookupLE
             , lookupGE
+            , IS.null
+            , size
             , isSubsetOf
             , isProperSubsetOf
             , disjoint
-
-            -- ** Insertion
-            , insert
-
-            -- * Delete
-            , delete
 
             -- * Combine
             , union
@@ -140,8 +140,6 @@ module Data.IntSet (
             -- ** List
             , elems
             , toList
-
-            -- ** Ordered list
             , toAscList
             , toDescList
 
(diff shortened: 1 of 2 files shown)


Data/IntSet.hs
```diff
@@ -73,10 +73,14 @@ module Data.IntSet (
...
+            , singleton
+            , fromList
+            , fromAscList
+            , fromDistinctAscList
 
-            -- * Query
+            -- * Queries
             , IS.null
             , size
             , member
```
treeowl: `member` goes at the top of the section. I don't think the section should be renamed like that.

Data/IntSet.hs
```diff
@@ -89,16 +93,17 @@ module Data.IntSet (
             , isProperSubsetOf
             , disjoint
 
-            -- * Construction
-            , empty
-            , singleton
+            -- ** Insertion
             , insert
+
+            -- * Delete
```
treeowl: Deletion

Data/IntSet.hs
```diff
@@ -89,16 +93,17 @@ module Data.IntSet (
             , isProperSubsetOf
             , disjoint
 
-            -- * Construction
-            , empty
-            , singleton
+            -- ** Insertion
             , insert
```
treeowl: Insertion and deletion should go right after construction.

Data/Set.hs
```diff
@@ -152,15 +159,10 @@ module Data.Set (
             -- ** List
```
treeowl: The two list sections should probably be merged.

Data/IntSet.hs
m-renaud: Done and done.

Data/IntSet.hs
m-renaud: Done.

Data/IntSet.hs
m-renaud: Done.

Data/Set.hs
m-renaud: Done.

treeowl:
I think it's an improvement and can be merged. I'm still not entirely satisfied, but I can't think of anything better.

m-renaud:
Sounds good and thanks for the review! I plan to keep making improvements when I think of them, if I come up with something I'll send out a PR and we can discuss there.

PR #532, opened 2018-02-09
Improve Data.Graph documentation.

m-renaud:
- Improves module docs
- Re-orders function documentation order
- Adds examples

### DIFF
--- a/Data/Graph.hs
+++ b/Data/Graph.hs
@@ -23,45 +23,67 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
--- A version of the graph algorithms described in:
+-- = Finite Graphs and Strongly Connected Components
 --
---   /Structuring Depth-First Search Algorithms in Haskell/,
---   by David King and John Launchbury.
+-- The @'Graph'@ type is an adjacency list representation of a finite, directed
+-- graph with @Int@ vertices.
+--
+-- The @'SCC'@ type represents a strongly-connected component of a graph.
+--
+-- == Implementation
+--
+-- The implementation is based on
+--
+--   * /Structuring Depth-First Search Algorithms in Haskell/,
+--     by David King and John Launchbury.
 --
 -----------------------------------------------------------------------------
 
-module Data.Graph(
+module Data.Graph (
 
-        -- * External interface
+    -- * Graphs
+      Graph
+    , Table
+    , Bounds
+    , Edge
+    , Vertex
 
-        -- At present the only one with a "nice" external interface
-        stronglyConnComp, stronglyConnCompR, SCC(..), flattenSCC, flattenSCCs,
+    -- ** Graph Construction
+    , graphFromEdges
+    , graphFromEdges'
+    , buildG
+    , transposeG
 
-        -- * Graphs
+    -- ** Graph Properties
+    , vertices
+    , edges
+    , outdegree
+    , indegree
 
-        Graph, Table, Bounds, Edge, Vertex,
+    -- ** Graph Algorithms
+    , dfs
+    , dff
+    , topSort
+    , components
+    , scc
+    , bcc
+    , reachable
+    , path
 
-        -- ** Building graphs
 
-        graphFromEdges, graphFromEdges', buildG, transposeG,
-        -- reverseE,
+    -- * Strongly Connected Components
+    , SCC(..)
 
-        -- ** Graph properties
+    -- ** Construction
+    , stronglyConnComp
+    , stronglyConnCompR
 
-        vertices, edges,
-        outdegree, indegree,
+    -- ** Conversion
+    , flattenSCC
+    , flattenSCCs
 
-        -- * Algorithms
-
-        dfs, dff,
-        topSort,
-        components,
-        scc,
-        bcc,
-        -- tree, back, cross, forward,
-        reachable, path,
-
-        module Data.Tree
+    -- * Trees
+    , module Data.Tree
 
     ) where
 
@@ -69,7 +91,6 @@ module Data.Graph(
 # define USE_ST_MONAD 1
 #endif
 
--- Extensions
 #if USE_ST_MONAD
 import Control.Monad.ST
 import Data.Array.ST (STUArray, newArray, readArray, writeArray)
@@ -108,7 +129,7 @@ import Data.Typeable
 
 -------------------------------------------------------------------------
 --                                                                      -
---      External interface
+--      Strongly Connected Components
 --                                                                      -
 -------------------------------------------------------------------------
 
@@ -255,22 +276,44 @@ type Bounds  = (Vertex, Vertex)
 -- | An edge from the first vertex to the second.
 type Edge    = (Vertex, Vertex)
 
--- | All vertices of a graph.
+-- | Returns the list of vertices in the graph.
+--
+-- ==== __Examples__
+--
+-- > vertices (buildG (0,-1) []) == []
+--
+-- > vertices (buildG (0,2) [(0,1),(1,2)]) == [0,1]
 vertices :: Graph -> [Vertex]
 vertices  = indices
 
--- | All edges of a graph.
+-- | Returns the list of edges in the graph.
+--
+-- ==== __Examples__
+--
+-- > edges (buildG (0,-1) []) == []
+--
+-- > edges (buildG (0,2) [(0,1),(1,2)]) == [(0,1),(1,2)]
 edges    :: Graph -> [Edge]
 edges g   = [ (v, w) | v <- vertices g, w <- g!v ]
 
 mapT    :: (Vertex -> a -> b) -> Table a -> Table b
 mapT f t = array (bounds t) [ (,) v (f v (t!v)) | v <- indices t ]
 
--- | Build a graph from a list of edges.
+-- | Builds a graph from a list of edges.
+--
+-- ==== __Examples__
+--
+-- > buildG (0,-1) [] == array (0,-1) []
+-- > buildG (0,2) [(0,1), (1,2)] == array (0,1) [(0,[1]),(1,[2])]
+-- > buildG (0,2) [(0,1), (0,2), (1,2)] == array (0,2) [(0,[2,1]),(1,[2]),(2,[])]
 buildG :: Bounds -> [Edge] -> Graph
 buildG bounds0 edges0 = accumArray (flip (:)) [] bounds0 edges0
 
 -- | The graph obtained by reversing all edges.
+--
+-- ==== __Examples__
+--
+-- > transposeG (buildG (0,2) [(0,1), (1,2)]) == array (0,2) [(0,[]),(1,[0]),(2,[1])]
 transposeG  :: Graph -> Graph
 transposeG g = buildG (bounds g) (reverseE g)
 
@@ -278,11 +321,23 @@ reverseE    :: Graph -> [Edge]
 reverseE g   = [ (w, v) | (v, w) <- edges g ]
 
 -- | A table of the count of edges from each node.
+--
+-- ==== __Examples__
+--
+-- > outdegree (buildG (0,-1) []) == array (0,-1) []
+--
+-- > outdegree (buildG (0,2) [(0,1), (1,2)]) == array (0,2) [(0,1),(1,1),(2,0)]
 outdegree :: Graph -> Table Int
 outdegree  = mapT numEdges
              where numEdges _ ws = length ws
 
 -- | A table of the count of edges into each node.
+--
+-- ==== __Examples__
+--
+-- > indegree (buildG (0,-1) []) == array (0,-1) []
+--
+-- > indegree (buildG (0,2) [(0,1), (1,2)]) == array (0,2) [(0,0),(1,1),(2,1)]
 indegree :: Graph -> Table Int
 indegree  = outdegree . transposeG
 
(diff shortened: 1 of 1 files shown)


Data/Graph.hs
```diff
@@ -23,53 +23,74 @@
...
-module Data.Graph(
-
-        -- * External interface
+module Data.Graph (
 
-        -- At present the only one with a "nice" external interface
-        stronglyConnComp, stronglyConnCompR, SCC(..), flattenSCC, flattenSCCs,
+    -- * Graphs
+      Graph
+    , Table
```
treeowl: Side note: I've been wondering whether the `Table` type synonym is doing more harm than good. What happens to signatures if we stop using it (but export it for backwards compatibility)?

Data/Graph.hs
```diff
@@ -23,53 +23,74 @@
...
+    , bcc
+    , reachable
+    , path
 
-        graphFromEdges, graphFromEdges', buildG, transposeG,
-        -- reverseE,
 
-        -- ** Graph properties
+    -- * Strongly Connected Components
+    , SCC(..)
```
treeowl: Types are generally introduced before the operations that involve them.

Data/Graph.hs
```diff
@@ -255,34 +276,68 @@ type Bounds  = (Vertex, Vertex)
...
 type Edge    = (Vertex, Vertex)
 
--- | All vertices of a graph.
+-- | Returns the list of vertices in the graph.
+--
+-- ==== __Examples__
+--
+-- > vertices (buildG (0,-1) []) == []
+--
+-- > vertices (buildG (0,2) [(0,1),(1,2)]) == [0,1]
```
treeowl: Which vertices? If your second example is right, it's not all.

Data/Graph.hs
```diff
@@ -255,34 +276,68 @@ type Bounds  = (Vertex, Vertex)
...
+-- > edges (buildG (0,2) [(0,1),(1,2)]) == [(0,1),(1,2)]
 edges    :: Graph -> [Edge]
 edges g   = [ (v, w) | v <- vertices g, w <- g!v ]
 
 mapT    :: (Vertex -> a -> b) -> Table a -> Table b
 mapT f t = array (bounds t) [ (,) v (f v (t!v)) | v <- indices t ]
 
--- | Build a graph from a list of edges.
+-- | Builds a graph from a list of edges.
+--
```
treeowl: I think I prefer "build", with other things made to match grammatically.

Data/Graph.hs
```diff
@@ -255,34 +276,68 @@ type Bounds  = (Vertex, Vertex)
...
+-- ==== __Examples__
+--
+-- > buildG (0,-1) [] == array (0,-1) []
+-- > buildG (0,2) [(0,1), (1,2)] == array (0,1) [(0,[1]),(1,[2])]
+-- > buildG (0,2) [(0,1), (0,2), (1,2)] == array (0,2) [(0,[2,1]),(1,[2]),(2,[])]
```
treeowl: Mention how this is partial, perhaps?

Data/Graph.hs
m-renaud: Personally I don't think its particularly helpful, although maybe adding more comments on it would probably be an improvement.

Data/Graph.hs
m-renaud: Good point, moved this up into the section with the graph types.

Data/Graph.hs
m-renaud: Oops, I originally had the bounds incorrect `(0,1)` which just silently discarded the last vertex. I fixed it but forgot to update the output.

As a side note, this API is really hard to use properly.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
m-renaud: Much better after removing the `Table` type alias from the types.

Data/Graph.hs
```diff
@@ -270,8 +270,8 @@ type Vertex  = Int
 type Table a = Array Vertex a
```
treeowl: There should probably be a Haddock line explaining that this is included for backwards compatibility.

Data/Graph.hs
treeowl: Yes, the API is very strange. It kind of looks like it wasn't meant for public consumption. My *guess* is that it may have started out GHC-internal, and someone else wanted to use it, and it ended up being exposed to the world before it really had a proper API. Those were fairly early days in Haskell history.

Data/Graph.hs
```diff
@@ -298,8 +356,35 @@ graphFromEdges' x = (a,b) where
 
 -- | Build a graph from a list of nodes uniquely identified by keys,
 -- with a list of keys of nodes this node should have edges to.
--- The out-list may contain keys that don't correspond to
--- nodes of the graph; they are ignored.
+--
+-- - @node@ is the type of data held at each vertex
+-- - @key@ is uniquely identifies each node in the graph
```
treeowl: s/is//

Data/Graph.hs
```diff
@@ -298,8 +356,35 @@ graphFromEdges' x = (a,b) where
...
+--
+-- @(graph, nodeFromVertex, vertexFromKey) = graphFromEdges edgeList@
+--
+-- * @graph@ is the raw, array based adjacency list for the graph.
+-- * @nodeFromVertex@ returns the node associated with the given 0-based @Int@
+--   vertex, see /warning/ below.
```
treeowl: That comma should be a semicolon; the emphasis looks weird.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
```diff
@@ -298,8 +356,35 @@ graphFromEdges' x = (a,b) where
...
+--
+-- An empty graph.
+-- > (graph, nodeFromVertex, vertexFromKey) = G.graphFromEdges []
+-- > graph = array (0,-1) []
+--
+-- > (graph, nodeFromVertex, vertexFromKey) = G.graphFromEdges [("a", 'a', ['b']), ("b", 'b', ['c']), ("c", 'c', [])]
+-- > graph == array (0,2) [(0,[1]),(1,[2]),(2,[])]
+-- > nodeFromVertex 0 == ("a",'a',"b")
+-- > vertexFromKey 'a' == Just 0
+--
```
treeowl: I don't think you've really explained this very well yet, unfortunately. That's largely because the function is horrible, but we should do our best. It sounds like we're taking an adjacency list representing a graph with vertices of type `key`, whose vertices are labeled with values of type `node`, and producing a `Graph`-based representation of that list. The `Graph` represents the shape, and the functions explains how the original adjacency list fits in. The first function could really just be `Vertex -> (key, node)`, but for convenience it includes the edges from the key. I think. Terrible API.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
m-renaud: Yeah, the API here is a mess. I would have expected `Graph` to be a type (without its internals exposed) which kept track of the association functions, and then each of the functions in the API would just operate on `Graph`. I would enjoy taking a stab at re-writing this module all together, obviously it would break compatibility so that would have to be resolved, but I'm convinced we can make this much better.

I've updated the comment and provided another example for how to "safely" use the API. Let me know what you think.

m-renaud: There's also the question of if the `containers` package should even have a thorough graph library or if that should be left to another package. I'm not sure if there's a "canonical" graph package, although there's a bunch of existing packages which provide graphs with varying levels of functionality/documentation/testing.

treeowl: I think you might have fun modernizing `fgl`. It's a pretty design, but the API is very old-fashioned. I don't know much about graphs myself, so I'm probably not the best one to give detailed advice.

Data/Graph.hs
```diff
@@ -298,8 +358,51 @@ graphFromEdges' x = (a,b) where
...
 -- | Build a graph from a list of nodes uniquely identified by keys,
 -- with a list of keys of nodes this node should have edges to.
--- The out-list may contain keys that don't correspond to
--- nodes of the graph; they are ignored.
+--
+-- This function takes an adjacency list representing a graph with vertices of
+-- type @key@ labeled by values of type @node@ and produces a @Graph@-based
+-- representation of that list. The @Graph@ result represents the /shape/ of the
+-- graph, and the functions describe how to access the vertices and associated
+-- labels given a key.
```
treeowl: You only describe the purpose of one of the two functions here.

Data/Graph.hs
```diff
@@ -298,8 +358,51 @@ graphFromEdges' x = (a,b) where
...
+--
+-- * @graph :: Graph@ is the raw, array based adjacency list for the graph.
+-- * @nodeFromVertex :: Vertex -> (node, key, [key])@ returns the node
+--   associated with the given 0-based @Int@ vertex; see /warning/ below.
+-- * @vertexFromKey :: key -> Maybe Vertex@ returns the @Int@ vertex for the
+--   key if it exists in the graph, @Nothing@ otherwise.
+--
+-- To safely use this API you must first call @vertexFromKey k@ to check if a
+-- vertex corresponds to the key @k@ given, and then use @nodeFromVertex@ to
+-- access the labelled node. See below for examples.
```
treeowl: Not true. You can also get vertices by actually looking at the `Graph`. You might fold over its list of vertices, for instance.

Data/Graph.hs
m-renaud: I'll reach out to the author :)

Data/Graph.hs
```diff
@@ -298,8 +358,51 @@ graphFromEdges' x = (a,b) where
...
+--
+-- Note: The out-list may contain keys that don't correspond to nodes of the
+-- graph; they are ignored.
```
treeowl: Can you include an example of this below?

Data/Graph.hs
m-renaud: Fixed.

Data/Graph.hs
m-renaud: Good call, updated paragraph.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
treeowl: I would leave out "is partial" and just have the explanation of how.

Data/Graph.hs
```diff
@@ -246,44 +267,83 @@ stronglyConnCompR edges0
...
 -- | A table of the count of edges into each node.
-indegree :: Graph -> Table Int
+--
+-- ==== __Examples__
+--
+-- > indegree (buildG (0,-1) []) == array (0,-1) []
+--
+-- > indegree (buildG (0,2) [(0,1), (1,2)]) == array (0,2) [(0,0),(1,1),(2,1)]
+indegree :: Graph -> Array Vertex Int
 indegree  = outdegree . transposeG
```
treeowl: This implementation looks stupidly inefficient. Can you open a ticket to fix it, and maybe a PR? I imagine we probably want to just build the result directly.

m-renaud: Done #533

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
m-renaud: I wonder if there's any reasonable way to figure out who's even using this API. It looks like there's some pretty low hanging fruit when it comes to improving the API (like hiding the functions you get back from `graphFromEdges`), but that would be a breaking API change. Maybe we could create a new module like `Data.DirectedGraph` which had the nice new API and just leave the old `Data.Graph` alone (module performance and comment improvements). WDYT?

m-renaud:
Any other suggestions? :)

Data/Graph.hs
```diff
@@ -23,53 +23,74 @@
...
-        -- * External interface
-
-        -- At present the only one with a "nice" external interface
-        stronglyConnComp, stronglyConnCompR, SCC(..), flattenSCC, flattenSCCs,
+module Data.Graph (
 
-        -- * Graphs
+    -- * Graphs
+      Graph
+    , Table
```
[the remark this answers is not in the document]
m-renaud: Moved it to the bottom of this section, adding a new section caused more noise than it was worth.

Data/Graph.hs
```diff
@@ -23,53 +23,74 @@
...
+    , Bounds
+    , Edge
+    , Vertex
+    , SCC(..)
```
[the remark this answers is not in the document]
m-renaud: Oops, yup misunderstanding, moved to the proper section.

Data/Graph.hs
```diff
@@ -23,53 +23,74 @@
...
 
-        Graph, Table, Bounds, Edge, Vertex,
+    -- ** Graph Construction
+    , graphFromEdges
+    , graphFromEdges'
+    , buildG
+    , transposeG
```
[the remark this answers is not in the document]
m-renaud: Moved to new Graph Transformations section.

Data/Graph.hs
```diff
@@ -23,53 +23,76 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
--- A version of the graph algorithms described in:
+-- = Finite Graphs and Strongly Connected Components
```
treeowl: Strongly connected components don't go in the title. They're strongly connected components *of graphs*.

Data/Graph.hs
```diff
@@ -23,53 +23,76 @@
...
 --
---   /Structuring Depth-First Search Algorithms in Haskell/,
---   by David King and John Launchbury.
+-- The @'Graph'@ type is an adjacency list representation of a finite, directed
+-- graph with @Int@ vertices.
```
treeowl: Perhaps "vertices of type Int"?

Data/Graph.hs
```diff
@@ -23,53 +23,76 @@
...
+--
+-- The @'SCC'@ type represents a strongly-connected component of a graph.
```
treeowl: Web link to an explanation of what that means.

Data/Graph.hs
```diff
@@ -298,8 +360,60 @@ graphFromEdges' x = (a,b) where
...
+-- To safely use this API you must either extract the list of vertices directly
+-- from the graph or first call @vertexFromKey k@ to check if a vertex
+-- corresponds to the key @k@. Once it is known that a vertex exists you can use
+-- @nodeFromVertex@ to access the labelled node and adjacent vertices. See below
+-- for examples.
+--
+-- Note: The out-list may contain keys that don't correspond to nodes of the
+-- graph; they are ignored.
+--
+-- Warning: The @nodeFromVertex@ function is partial and will cause a runtime
```
treeowl: Remove "is partial and".

Data/Graph.hs
```diff
@@ -298,8 +360,60 @@ graphFromEdges' x = (a,b) where
...
+--
+--
+-- A graph with 3 vertices: ("a") -> ("b") -> ("c")
+--
+-- > (graph, nodeFromVertex, vertexFromKey) = graphFromEdges [("a", 'a', ['b']), ("b", 'b', ['c']), ("c", 'c', [])]
+-- > graph == array (0,2) [(0,[1]),(1,[2]),(2,[])]
+-- > nodeFromVertex 0 == ("a",'a',"b")
+-- > vertexFromKey 'a' == Just 0
+--
+-- How to use the API to get the label for a given key.
```
treeowl: Remove "How to use the API to". The entire point of the documentation is to explain how to use the API!

Data/Graph.hs
m-renaud: Removed.

Data/Graph.hs
m-renaud: Yeah, that's better.

Data/Graph.hs
m-renaud: Added link to wikipedia.

Data/Graph.hs
m-renaud: Done.

Data/Graph.hs
m-renaud: Haha, yes it is, removed :)

PR #534, opened 2018-02-10
Improve Data.Tree documentation.

m-renaud:
This partially addresses https://github.com/haskell/containers/issues/495.

### DIFF
--- a/Data/Tree.hs
+++ b/Data/Tree.hs
@@ -348,14 +348,28 @@ foldTree f = go where
 --
 -- ==== __Examples__
 --
--- Construct an infinite tree with a single chain of 1's.
+-- Construct the tree of @Integer@s where each node has two children:
+-- left = 2*x and right = 2*x + 1. Stop when the values get greater than 7.
 --
--- > take 3 $ levels $ unfoldTree (\x -> (x, [x])) 1 = [[1],[1],[1]]
+-- > let buildNode x = if 2*x + 1 > 7 then (x, []) else (x, [2*x, 2*x+1])
+-- > putStr $ drawTree $ fmap show $ unfoldTree buildNode 1
 --
--- Construct the infinite tree of @Integer@s where each node has two children:
--- left = 2*x and right = 2*x + 1.
+-- @
 --
--- > take 3 $ levels $ unfoldTree (\x -> (x, [2*x, 2*x+1])) 1 == [[1],[2,3],[4,5,6,7]]
+-- 1
+-- |
+-- +- 2
+-- |  |
+-- |  +- 4
+-- |  |
+-- |  `- 5
+-- |
+-- `- 3
+--    |
+--    +- 6
+--    |
+--    `- 7
+-- @
 --
 unfoldTree :: (b -> (a, [b])) -> b -> Tree a
 unfoldTree f b = let (a, bs) = f b in Node a (unfoldForest f bs)
--- a/Data/Tree.hs
+++ b/Data/Tree.hs
@@ -21,10 +21,10 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
--- = Multi-way (aka. Rose Trees) and Forests
+-- = Multi-way Trees and Forests
 --
 -- The @'Tree' a@ type represents a lazy, possibly infinite, multi-way tree
--- (/aka./ rose tree).
+-- (also known as a /rose tree/).
 --
 -- The @'Forest' a@ type represents a forest of @'Tree' a@s.
 --
--- a/Data/Tree.hs
+++ b/Data/Tree.hs
@@ -21,20 +21,38 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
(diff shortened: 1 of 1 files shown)


Data/Tree.hs
```diff
@@ -21,20 +21,38 @@
 -- Maintainer  :  libraries@haskell.org
 -- Portability :  portable
 --
--- Multi-way trees (/aka/ rose trees) and forests.
+-- = Multi-way (aka. Rose Trees) and Forests
```
treeowl: Please don't use the abbreviation "AKA" anywhere in the documentation of any package I (co-)maintain. Documentation is *formal writing*. That doesn't mean there can't be occasional, tasteful, and *actually funny* humor occasionally, but it does mean no language like that.

Data/Tree.hs
```diff
@@ -21,20 +21,38 @@
...
+--
+-- The @'Tree' a@ type represents a lazy, possibly infinite, multi-way tree
+-- (/aka./ rose tree).
+--
```
treeowl: Here too.

Data/Tree.hs
```diff
@@ -21,20 +21,38 @@
...
+-- The @'Forest' a@ type represents a forest of @'Tree' a@s.
 --
```
treeowl: I'm somewhat more inclined to stop using the `Forest` type synonym, and explain somewhere that a "forest" is a list of trees. What do you think?

Data/Tree.hs
m-renaud: Fair enough. It was in there before, I wasn't sure if you or someone else added it, I've removed it now.

Data/Tree.hs
m-renaud: Done.

Data/Tree.hs
m-renaud: I tend to agree, this definition doesn't even match what anyone who's familiar with the term forest would expect either since a forest from graph theory is an undirected graph where each connected component is a tree. That being said, I don't think this PR is the place to remove it :)

Data/Tree.hs
```diff
@@ -232,64 +280,137 @@ draw (Node x ts0) = lines x ++ drawSubTrees ts0
...
+-- @Node { rootLabel=b, subForest=[] }@ and repeatedly applying @f@ to each
+-- 'rootLabel' value in the tree's leaves to generate its 'subForest'.
+--
+-- For a monadic version see 'unfoldTreeM_BF'.
+--
+-- ==== __Examples__
+--
+-- Construct an infinite tree with a single chain of 1's.
+--
+-- > take 3 $ levels $ unfoldTree (\x -> (x, [x])) 1 = [[1],[1],[1]]
```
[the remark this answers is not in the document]
m-renaud: I wasn't sure of another way to show what part of an infinite tree looks like, I'm open to suggestions though.

Data/Tree.hs
```diff
@@ -232,64 +280,137 @@ draw (Node x ts0) = lines x ++ drawSubTrees ts0
...
+-- For a monadic version see 'unfoldForestM_BF'.
+--
 unfoldForest :: (b -> (a, [b])) -> [b] -> Forest a
 unfoldForest f = map (unfoldTree f)
 
--- | Monadic tree builder, in depth-first order
+-- | Monadic tree builder, in depth-first order.
+--
+-- Warning: This will hang forever if trying to construct an infinite tree,
+-- consider using 'unfoldTreeM_BF'.
```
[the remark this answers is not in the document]
m-renaud: Ah you're right, removed the warning.

Data/Tree.hs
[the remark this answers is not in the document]
treeowl: I'm not sure you need an infinite tree. One or two finite ones should probably be sufficient.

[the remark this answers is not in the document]
m-renaud: Used a finite tree example and the drawTree function to see the results.

Data/Tree.hs
```diff
@@ -232,34 +280,111 @@ draw (Node x ts0) = lines x ++ drawSubTrees ts0
...
--- | Build a tree from a seed value
+-- | Build a (possibly infinite) tree from a seed value in breadth-first order.
+--
+-- @unfoldTree f b@ constructs a tree by starting with the tree
+-- @Node { rootLabel=b, subForest=[] }@ and repeatedly applying @f@ to each
+-- 'rootLabel' value in the tree's leaves to generate its 'subForest'.
+--
+-- For a monadic version see 'unfoldTreeM_BF'.
+--
+-- ==== __Examples__
```
treeowl: There's only one example now.

m-renaud: Yeah, I wasn't sure if the other one was useful. I can add it back in if you think it would be.

treeowl: Don't remember what it was, but I suspect one good example should be sufficient.

m-renaud: It was a tree that consisted of a chain of 1's `Node 1 [Node 1 [...]]`, not super useful :P

m-renaud:
Anything else, or should I merge?

PR #536, opened 2018-02-14
Remove and disable deprecated things

treeowl:
Disable all deprecated functions by making them produce
type errors when used. This is an experimental removal approach.
It may or may not really be a good idea. The benefit is that
unlike deprecation it *forces* people to make a change, and
unlike immediate removal it tells them what change to make.


treeowl:
Ugh. Making this work with pre-7.6 compilers will be annoying. I guess I can do it, but I'm glad this crud well soon be gone. Also, those compilers will be gone in another two versions.

treeowl:
Since there isn't really much demand for pre-7.6 compilers, I just pulled support for those. Now we should be able to do this without too much trouble.

PR #540, opened 2018-02-28
Add Generic and Generic1 for Seq internals

treeowl:
Add `Generic` and `Generic1` instances for `Data.Sequence.Internal`
types: `FingerTree`, `Node`, `Elem`, and `Digit`.

### DIFF
--- a/Data/Sequence/Internal.hs
+++ b/Data/Sequence/Internal.hs
@@ -932,6 +932,14 @@ data FingerTree a
     deriving Show
 #endif
 
+#ifdef __GLASGOW_HASKELL__
+-- | @since 0.6.1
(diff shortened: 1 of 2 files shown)


treeowl:
There's a most unpleasant tension here. People mucking with `Data.Sequence` internals will want `Generic1`, and that's perfectly fair. They'll want `FingerTree` and `Node` to have `Functor` instances so they can do generic things with, e.g., `Comp1 FingerTree Node`. Perfectly fair. But someone wanting `FingerTree` for some other purpose (as an `(Int, +, <=)` finger tree) probably *doesn't* want that `Functor` instance, because it violates the annotation invariant. Ugh.

PR #541, opened 2018-03-05
Fix Documentation Typo in Data/Sequence.hs

staffehn:
In the module header, in a code example where “index” is used infix, the back-ticks need to be escaped, otherwise haddock interprets them as (an alternative form of) quotes for a hyperlinked identifier which in particular makes the back-ticks invisible in the documentation and displays invalid Haskell code in the code example.

### DIFF
--- a/Data/Sequence.hs
+++ b/Data/Sequence.hs
@@ -65,7 +65,7 @@
 --
 --     but
 --
---     @ (1 :<| undefined) `index` 0 = undefined @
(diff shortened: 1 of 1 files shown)


treeowl:
Ugh. Thanks a lot. Do you know if there are any other examples?

staffehn:
@treeowl I went and did a search through all of cointainers right now and I think I catched all of them now in https://github.com/haskell/containers/pull/545.

PR #545, opened 2018-03-09
Fix Documentation Typos

staffehn:
A few more typos like in: https://github.com/haskell/containers/pull/541

Additional remarks:
This problem to only affect code inside @s, not code after a > for some reason
I did a search through all of containers and also verified every place I fixed was a visible error on the Hackage packed documentation web-pages before.
There are (at least) 2 valid ways to get the infix \`s right. One is escape them with \s, so use \\\`s. This creates the infix notation in the documentation correctly and does not convert the identifier into a link. A second way is to use \`\`s (two \`s) on both sides (for example @ ... \`\`elem\`\` ... @). This works as opposed to ( ... \`'elem'\` ...) or anything else I tried, which gave me a few different kinds erroneous results, and it creates valid infix syntax in the doc _including_ a link to the definition.

I went with making everything links almost everywhere, in particular if they already were in a style where the author obviously _wanted_ a link, except for in a function’s documentation for the identifier of that function itself (that link would take you nowhere else), a style that seemed consistent with most of the rest of the documentation.


treeowl:
It looks like those failures were trouble downloading from Hackage. I doubt
there's anything wrong, but I restarted them to be sure.

treeowl:
By the way, code after `>` is totally literal and unformatted, which is why there are no problems with backticks there.

Data/Map/Internal.hs
```diff
@@ -1981,8 +1981,8 @@ intersection t1@(Bin _ k x l1 r1) t2
 -- found in a 'Set'.
 --
 -- @
--- m `'restrictKeys'` s = 'filterWithKey' (\k _ -> k `'Set.member'` s) m
--- m `'restrictKeys'` s = m `'intersect' 'fromSet' (const ()) s
+-- m \`restrictKeys\` s = 'filterWithKey' (\k _ -> k ``Set.member`` s) m
+-- m \`restrictKeys\` s = m ``intersect`` 'fromSet' (const ()) s
```
treeowl: What's going on here? Backslash on the left, double ticks on the right?

staffehn: This is in the documentation for restrictKeys. Just as I said: “I went with making everything links […], except for in a functions documentation for the identifier of that function itself (that link would take you nowhere else)”.

treeowl:
Wow, I really should've figured that out. But what do the doubled ticks do?
Link?

staffehn:
Yes it makes a link (you probably didn’t get that edit by mail). Just as I said above:

> There are (at least) 2 valid ways to get the infix \`s right. One is escape them with \s, so use \\\`s. This creates the infix notation in the documentation correctly and does not convert the identifier into a link. A second way is to use \`\`s (two \`s) on both sides (for example @ ... \`\`elem\`\` ... @). This works as opposed to ( ... \`'elem'\` ...) or anything else I tried, which gave me a few different kinds erroneous results, and it creates valid infix syntax in the doc _including_ a link to the definition.

staffehn:
Btw, the whole library documentation is _really_ inconsistent with where to place a link and where not, so I have absolutely no clue what would be the best option – that was just my guess, as I thought I saw it done the same style somewhere else in those documentations. Reviewing them now, I can’t really find where I got that impression from again, but only (as stated) that the whole story a pretty inconsistent and random thing.

treeowl:
Trying too hard to make things consistent will only make you crazy.

Data/Map/Strict/Internal.hs
```diff
@@ -1379,7 +1379,7 @@ mapWithKey f (Bin sx kx x l r) =
 #endif
 
 -- | /O(n)/.
--- @'traverseWithKey' f m == 'fromList' <$> 'traverse' (\(k, v) -> (\v' -> v' `seq` (k,v')) <$> f k v) ('toList' m)@
+-- @'traverseWithKey' f m == 'fromList' <$> 'traverse' (\(k, v) -> (\v' -> v' ``seq`` (k,v')) <$> f k v) ('toList' m)@
```
treeowl: I don't think we want links for Prelude basics like `seq`. The only exceptions (maybe) are for list functions that we're simulating.

staffehn: alright, changed that.

treeowl:
Meh. And yet someone linked to `fmap` and `snd`. Maybe we'll make things consistent some year (probably not). Thanks for your work here.

PR #549, opened 2018-03-30
Add functions to optimize graph creation time (when keys are consecutive), add benchmarks.

OlivierSohn:
The motivation for this PR is the following: I am developping a game where world creation is automated using graphs based on random matrices: in the process I need to create a lot of graphs before finding a valid configuration, hence I'm trying to optimize graph creation time.

Profiling showed that `graphFromEdges` was one of the bottlenecks. This function has a time complexity of O((E+V) * log V) (E is number of edges, V number of vertices). But if the keys are known to be ascending and consecutive, we can change the implementation and make the complexity drop to O(V+E).

This is what I did in this PR, and I added benchmarks to confirm this (note that bench.txt will be removed before merging of course)

This PR adds 2 functions to the public API of Data.Graph : `graphFromConsecutiveAscKeys` and `graphFromConsecutiveKeys`.

One further API change I'd like to introduce is to make the constructor of `Graph` public, to optimize graph creation even further, improving on the constant factor this time. But it is not part of this PR.

If needed I can discuss this on the mailing list but I open this PR as a first step, waiting for instructions on how to proceed next.

### DIFF
--- a/Data/Graph.hs
+++ b/Data/Graph.hs
@@ -38,6 +38,13 @@
 --   * /Structuring Depth-First Search Algorithms in Haskell/,
 --     by David King and John Launchbury.
 --
+-- == Complexity notations
+--
+-- The following conventions are used when documenting complexities:
+--
+-- * @E@ = Count of edges
+-- * @V@ = Count of vertices
+--
 -----------------------------------------------------------------------------
 
 module Data.Graph (
@@ -52,6 +59,8 @@ module Data.Graph (
     -- ** Graph Construction
     , graphFromEdges
     , graphFromEdges'
+    , graphFromConsecutiveKeys
+    , graphFromConsecutiveAscKeys
     , buildG
 
     -- ** Graph Properties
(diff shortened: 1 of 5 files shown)


treeowl:
`Graph` doesn't have a constructor; it's just a type synonym. What do you want to export?

OlivierSohn:
@treeowl Oups, my bad ! I was assuming it was a data or newtype...

OlivierSohn:
I'd like to mention that I'm not strongly attached to the names I used for the new functions : `graphFromConsecutive(Asc)Keys`. Initially, to be closer to the existing names I used `graphFromEdgesWithConsecutive(Asc)Keys`, but then thought it would be a long name so I tried to shorten it. Also, I was not convinced that `Edge` was the right concept to bring forward in the name, as IMO what we pass to those functions is closer to a list of Nodes (where a Node _contains_ a list of out edges).

But to be entirely consistent, if we go this way we might as well need to rename `graphFromEdges` to `graphFromKeys`, and this change would be maybe more subject to discussion as it would break the API for the users.

So I would understand if there is a need to go back to the old names mentionning Edges.

treeowl:
I think this should all be raised and discussed on the libraries list.

OlivierSohn:
ok, I'll initiate the discussion then, thanks!

OlivierSohn:
@treeowl I'll leave this opened if you think it's usefull, else I'll close it. I use another version of Graph for my projects anyway now, so merging it is not a priority. The discussion on the mailing list didn't raise any important issue. Thanks.

PR #551, opened 2018-04-10
Update documentation for strict insertWithKey.

AndreasPK:
The documentation describes the arguments as `insertWithKey f key value mp` .
This updates the later referenced argument names to match these.


treeowl:
Thanks!

PR #554, opened 2018-06-23
Express `spanAntitone` in terms of `partitionWithKey` instead of `partition`

sgraf812:
The current haddock doesn't type-check.


treeowl:
Good catch.

PR #559, opened 2018-08-03
Add examples for foldTree

ploeh:
The two existing examples could also be implemented using the `Foldable` instance of `Tree`, although I realize that the `Foldable` `maximum` implementation is partial, while the example using `foldTree` is total.

The two additional examples are operations that, as far as I can tell, can't be implemented with `Foldable`, but only with the full catamorphism. This, hopefully, might be compelling examples.

### DIFF
--- a/Data/Tree.hs
+++ b/Data/Tree.hs
@@ -313,6 +313,14 @@ levels t =
 --
 -- > foldTree (\x xs -> maximum (x:xs)) (Node 1 [Node 2 [], Node 3 []]) == 3
 --
+-- Count the number of leaves in the tree:
+--
+-- > foldTree (\_ xs -> if null xs then 1 else sum xs) (Node 1 [Node 2 [], Node 3 []]) == 2
+--
+-- Find depth of the tree; i.e. the number of branches from the root of the tree to the furthest leaf:
+--
+-- > foldTree (\_ xs -> if null xs then 0 else 1 + maximum xs) (Node 1 [Node 2[], Node 3 []]) == 1
+--
 --
 -- @since 0.5.8
 foldTree :: (a -> [b] -> b) -> Tree a -> b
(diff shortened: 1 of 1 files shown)


treeowl:
Yes. Could you also add an implementation of `traverse` using `foldTree`, to demonstrate its power?

ploeh:
I'll be happy to try 😄 If I succeed, do you want it as a new commit to this branch, or in a separate pull request?

treeowl:
You will certainly succeed. Use the same PR. Either add a commit or amend
this one.

ploeh:
I wasn't sure exactly how to format the `traverse` example, but I've now added a commit with that. Let me know if you want me to change the formatting, add more text, and so on.

Data/Tree.hs
```diff
@@ -313,6 +313,19 @@ levels t =
...
+-- > foldTree (\_ xs -> if null xs then 1 else sum xs) (Node 1 [Node 2 [], Node 3 []]) == 2
+--
+-- Find depth of the tree; i.e. the number of branches from the root of the tree to the furthest leaf:
+--
+-- > foldTree (\_ xs -> if null xs then 0 else 1 + maximum xs) (Node 1 [Node 2[], Node 3 []]) == 1
+--
+-- You can even implement traverse using foldTree:
+--
+-- > traverse' f = foldTree (\x xs -> liftA2 Node (f x) (sequenceA xs))
+-- > fmap sum (traverse' Just (Node 1 [Node 2 [], Node 3[]])) == Just 6
```
treeowl: We definitely don't need an example here of what `traverse` does.

ploeh: Removed that line.

ploeh:
Is there anything more I need to do to the pull request?

treeowl:
Nope. Thanks.

PR #560, opened 2018-08-09
Haddock fixes

harpocrates:
Tweaked documentation fixing Haddock markup. Mostly:

  * qualifying out-of-scope or ambiguous identifiers
  * properly escaping character literals in examples
  * several obvious typos

This dramatically reduces the number of Haddock warnings.

### DIFF
--- a/Data/Graph.hs
+++ b/Data/Graph.hs
@@ -400,7 +400,7 @@ graphFromEdges' x = (a,b) where
 -- > (graph, nodeFromVertex, vertexFromKey) = graphFromEdges []
 -- > graph = array (0,-1) []
 --
--- A graph where the out-list references unspecified nodes (@'c'@), these are
+-- A graph where the out-list references unspecified nodes (@\'c\'@), these are
 -- ignored.
 --
 -- > (graph, _, _) = graphFromEdges [("a", 'a', ['b']), ("b", 'b', ['c'])]
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -976,8 +976,8 @@ alter f k Nil     = case f Nothing of
 -- 'alterF' is the most general operation for working with an individual
 -- key that may or may not be in a given map.
 --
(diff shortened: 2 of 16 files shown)


harpocrates:
I noticed something suspicious that I wasn't sure what to do:

https://github.com/haskell/containers/blob/17494c9ecd5c5cf54d9cd11c7698f57b3945a0dd/Data/IntMap/Internal.hs#L2444-L2447

Should the `mapAccumR` there be `mapAccumRWithKey`? This is an issue in all three internal `Map` modules.

m-renaud:
Hi @harpocrates, thanks for the PR! Could you include the before/after Haddock warnings? Also, are the updated Haddocks available to browse somewhere?

harpocrates:
Output of `cabal new-haddock -w ghc-8.6.2` [before](https://github.com/haskell/containers/files/2636434/haddock-before.txt) and [after](https://github.com/haskell/containers/files/2636435/haddock-after.txt).

I do not have the updated Haddocks available anywhere, but `cabal new-haddock` should take 5 seconds to run. Since I haven't changed any actual content you should only see: more links, less warnings, and properly escaped quotes.

m-renaud:
Thanks! All the changes look good, I had forgotten that Haddock tries to autolink the `'a'` inside the `@ ... @` blocks, whereas using "bird tracks" (`>`) doesn't attempt to do any further markup (see the [Code Blocks](https://www.haskell.org/haddock/doc/html/ch03s08.html#idm140354810770608) section of the Haddock docs).

I'm wondering if we should just use preceeding `>` throughout for consistency since I don't think there's many (if any) places where we want autolinking inside the example blocks. Anyways, we don't need to do that in this PR.

harpocrates:
I'd prefer if you stuck with `@ ... @` code blocks. I'm not very fond of the bird tracks - they mean block quote pretty much everywhere outside of Haddock. If they hadn't been around for so long, I'd even consider deprecating them.

Obviously, it is entirely up to you.

PR #563, opened 2018-08-30
Add Lift instances to most exposed types.

quasicomputational:


quasicomputational:
Oh, ugh. template-haskell on GHC 7.8 and before has a containers dep, so this is a no go until that is out of the support window. Unless that's the way around that the dependency ought to be, in which case maybe template-haskell should re-grow the containers dep and add the instances there.

treeowl:
I think your instincts are right about the dependency order (i.e., that `containers` should depend on `template-haskell`), but it wouldn't hurt to ask the libraries list.

PR #567, opened 2018-09-28
Switch to using 'friendly' pygments style for docs.

m-renaud:
The 'sphinx' style changed to have a dull green background colour which is ugly
and more difficult to read. The 'friendly' style has a neutral background and is
nicer to look at (and looks like what the 'sphinx' style used to look like).


treeowl:
*Sigh*.

PR #573, opened 2018-10-16
Fix sample code of alterF in haddock

igrep:
Found by a question in StackOverflow https://stackoverflow.com/questions/52828746/haskell-data-map-strict-alterf-parse-error-on-input/52829097


treeowl:
Looks like I really goofed that documentation up! Thanks for the fix.

PR #576, opened 2018-11-08
Remove unused imports.

DavidEichmann:
Due to a bug in ghc, some unused imports do not yield warnings.
This commit will remove such unused imports in preparation for
the ghc bug fix (see https://ghc.haskell.org/trac/ghc/ticket/13064).


DavidEichmann:
Hi @treeowl, Thanks for looking at this. This is my first containers PR, so please let me know if there is anything I need to do to get this merged (I assume you have merge permission).

PR #585, opened 2018-12-19
Update Data.Set.unions documentation

morteako:
The previous documentation for Data.Set.unions said that you need a list of sets, 
but the type says that it can be any Foldable structure. `unions :: (Foldable f, Ord a) => f (Set a) -> Set a`.

Changed the wording to `The union of the sets in a Foldable structure`.
(Feel free to change the wording into something better, ofc)


m-renaud:
Thanks @morteako!

PR #592, opened 2019-01-17
Add forM

hasufell:
I'm not entirely sure if `forM s f = fmap fromList . (flip T.forM f) . toList $ s` would be faster.

### DIFF
--- a/containers/src/Data/Set.hs
+++ b/containers/src/Data/Set.hs
@@ -136,6 +136,12 @@ module Data.Set (
             -- * Map
             , S.map
             , mapMonotonic
+#if __GLASGOW_HASKELL__ >= 710
+            , S.forM
+            , S.for
+            , S.mapM
+            , S.traverse
+#endif
 
             -- * Folds
             , S.foldr
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -181,6 +181,12 @@ module Data.Set.Internal (
             -- * Map
             , map
             , mapMonotonic
+#if __GLASGOW_HASKELL__ >= 710
+            , mapM
+            , forM
+            , traverse
+            , for
+#endif
 
             -- * Folds
             , foldr
@@ -230,7 +236,8 @@ module Data.Set.Internal (
             , merge
             ) where
 
-import Prelude hiding (filter,foldl,foldr,null,map,take,drop,splitAt)
+import Prelude hiding (filter,foldl,foldr,null,map,take,drop,splitAt,mapM,traverse)
+import qualified Prelude
 import Control.Applicative (Const(..))
 import qualified Data.List as List
 import Data.Bits (shiftL, shiftR)
@@ -952,6 +959,33 @@ map f = fromList . List.map f . toList
 mapMonotonic :: (a->b) -> Set a -> Set b
 mapMonotonic _ Tip = Tip
 mapMonotonic f (Bin sz x l r) = Bin sz (f x) (mapMonotonic f l) (mapMonotonic f r)
+
+
+#if __GLASGOW_HASKELL__ >= 710
+-- | /O(n*log n)/.
+--
+-- Like 'traverse' from Data.Traversable. This is less generic, since 'Set'
+-- does not have a Traversable instance.
+traverse :: (Ord b, Applicative m) => (a -> m b) -> Set a -> m (Set b)
+-- This implementation is significantly faster. We don't know why.
+traverse f s0 = fmap (GHCExts.fromListN (size s0)) . Prelude.traverse f . toList $ s0
+
+-- | 'for' is 'traverse' with its arguments flipped.
+for :: (Ord b, Applicative m) => Set a -> (a -> m b) -> m (Set b)
+for = flip traverse
+
+-- | /O(n*log n)/.
+--
+-- Like 'mapM' from Data.Traversable. This is less generic, since 'Set'
+-- does not have a Traversable instance.
+mapM :: (Ord b, Monad m) => (a -> m b) -> Set a -> m (Set b)
+mapM = traverse
+
+-- | 'forM' is 'mapM' with its arguments flipped.
+forM :: (Ord b, Monad m) => Set a -> (a -> m b) -> m (Set b)
+forM = flip mapM
+#endif
+
 
 {--------------------------------------------------------------------
   Fold
--- a/containers-tests/benchmarks/Set.hs
+++ b/containers-tests/benchmarks/Set.hs
@@ -1,12 +1,18 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE BangPatterns #-}
 
 module Main where
 
 import Control.DeepSeq (rnf)
 import Control.Exception (evaluate)
-import Gauge (bench, defaultMain, whnf)
+import Gauge (bench, defaultMain, whnf, nf)
 import Data.List (foldl')
 import qualified Data.Set as S
+import Data.Functor.Identity
+#if __GLASGOW_HASKELL__ >= 710
+import Control.Monad.ST
+import Data.STRef
+#endif
 
 main = do
     let s = S.fromAscList elems :: S.Set Int
@@ -21,6 +27,22 @@ main = do
         , bench "filter" $ whnf (S.filter ((== 0) . (`mod` 2))) s
         , bench "partition" $ whnf (S.partition ((== 0) . (`mod` 2))) s
         , bench "fold" $ whnf (S.fold (:) []) s
+#if __GLASGOW_HASKELL__ >= 710
+        , bench "forM (Identity)" $ nf (\s -> runIdentity $ S.forM s (\e -> pure $ e + 1)) s
+        , bench "for (Identity)" $ nf (\s -> runIdentity $ S.for s (\e -> pure $ e + 1)) s
+        , bench "forM (ST)" $ nf (\s -> runST $ do
+              ref <- newSTRef 0
+              S.forM s (\e -> do
+                              modifySTRef ref (+ 1)
+                              x <- readSTRef ref
+                              pure $ x + e * 36 `mod` 11)) (S.map (`mod` 51) s)
+        , bench "for (ST)" $ nf (\s -> runST $ do
+              ref <- newSTRef 0
+              S.for s (\e -> do
+                             modifySTRef ref (+ 1)
+                             x <- readSTRef ref
+                             pure $ x + e * 36 `mod` 11)) (S.map (`mod` 51) s)
+#endif
         , bench "delete" $ whnf (del elems) s
         , bench "findMin" $ whnf S.findMin s
         , bench "findMax" $ whnf S.findMax s
--- a/containers-tests/tests/set-properties.hs
+++ b/containers-tests/tests/set-properties.hs
@@ -5,7 +5,7 @@ import qualified Data.List as List
 import Data.Monoid (mempty)
 import Data.Maybe
 import Data.Set
-import Prelude hiding (lookup, null, map, filter, foldr, foldl, all, take, drop, splitAt)
+import Prelude hiding (lookup, null, map, filter, foldr, foldl, forM, all, take, drop, splitAt)
 import Test.Framework
 import Test.Framework.Providers.HUnit
 import Test.Framework.Providers.QuickCheck2
@@ -88,6 +88,10 @@ main = defaultMain [ testCase "lookupLT" test_lookupLT
                    , testProperty "prop_map" prop_map
                    , testProperty "prop_map2" prop_map2
                    , testProperty "prop_mapMonotonic" prop_mapMonotonic
+#if __GLASGOW_HASKELL__ >= 710
+                   , testProperty "prop_forM" prop_forM
+                   , testProperty "prop_for" prop_for
+#endif
                    , testProperty "prop_maxView" prop_maxView
                    , testProperty "prop_minView" prop_minView
                    , testProperty "prop_split" prop_split
@@ -590,6 +594,14 @@ prop_map2 f g s = map (apply f) (map (apply g) s) === map (apply f . apply g) s
 
 prop_mapMonotonic :: Set Int -> Property
 prop_mapMonotonic s = mapMonotonic id s === s
+
+#if __GLASGOW_HASKELL__ >= 710
+prop_forM :: Set Int -> Bool
+prop_forM s = runIdentity (forM s (pure . id)) == map id s
+
+prop_for :: Set Int -> Bool
+prop_for s = runIdentity (for s (pure . id)) == map id s
+#endif
 
 prop_maxView :: Set Int -> Bool
 prop_maxView s = case maxView s of


treeowl:
It seems a bit odd to offer the flipped version of `mapM` and not `mapM`. And it seems quite odd to offer a version of `mapM` without offering a version of `traverse`. The `foldr'` is *extremely* surprising; I'd be astonished if that was what you really intended. The `toList` version should surely tolerate an `Applicative` constraint.

```haskell
traverseSet :: (Ord b, Applicative f) => (a -> f b) -> Set a -> f (Set b)
traverseSet f = fmap fromListN . traverse f . toList
```

I'm not sure whether we can do better than that. The monadic version should probably look like this:

```haskell
mapMSet :: (Ord b, Monad m) => (a -> m b) -> Set a -> m (Set b)
mapMSet f s0 = foldr go return s0 empty
  where
     go x r s = f x >>= \y -> r $! insert y s
```

or alternatively perhaps

```haskell
mapMSet :: (Ord b, Monad m) => (a -> m b) -> Set a -> m (Set b)
mapMSet f = foldM go empty
  where
    go s x = f x >>= \y -> pure $! insert y s
```

Note: you should definitely email the libraries list with a formal proposal; in particular, naming these things can be a tad tricky. `traverse` and `mapM` would probably be "right", but have the potential to break existing code. I don't feel strongly about the names.

hasufell:
I cannot really judge what is more efficient or more performant, since reasoning about those things are not easy in haskell.

treeowl:
Ah, I see now that you've written some very similar code yourself, in http://hackage.haskell.org/package/hpath-0.9.2/docs/src/System.Posix.Directory.Traversals.html#traverseDirectory

hasufell:
> Do you understand why the `foldr'`-based version you proposed does not do so?

Yes, becauce `b` in `(\a b -> f a >>= (\x -> fmap (insert x) b))` is the recursion, so we are fmapping into an action into an action...

However, if we never care about the Set, but only about the actions, then foldr would be more "efficient" no?

> do you understand why `Applicative` isn't sufficient to work incrementally?

I'm not sure what you mean by that.

hasufell:
ping

treeowl:
You still haven't come up with an implementation I consider reasonable. `foldr'` still has no business there.

treeowl:
Much better. But you'll still need to propose this on the libraries list to get community input.

hasufell:
> Much better. But you'll still need to propose this on the libraries list to get community input.

I did

treeowl:
Can you link to the discussion thread here? Unfortunately I don't remember it.

hasufell:
> Can you link to the discussion thread here? Unfortunately I don't remember it.

Actually, not I did, but someone else: https://mail.haskell.org/pipermail/libraries/2019-January/029335.html

hasufell:
I'm also interested in benchmarks and will test a bit I think.

treeowl:
Ah, yes, that "someone else" was actually me. But there wasn't any real discussion, aside from one comment by Henning. Could you try to reignite the conversation?

hasufell:
I compared

```hs
forM :: (Ord b, Monad m) => Set a -> (a -> m b) -> m (Set b)
forM s0 f = fmap (GHCExts.fromListN (size s0)) . traverse f . toList $ s0
```

with

```hs
forM :: (Ord b, Monad m) => Set a -> (a -> m b) -> m (Set b)
forM s0 f = foldr go return s0 empty
  where
     go x r s = f x >>= \y -> r $! insert y s
```

---

toList version:

```
benchmarked forM
time                 125.3 μs   (124.1 μs .. 126.2 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 124.8 μs   (123.9 μs .. 125.7 μs)
std dev              3.237 μs   (2.729 μs .. 4.030 μs)
```

foldr version:

```
benchmarked forM
time                 1.548 ms   (1.544 ms .. 1.553 ms)
mean                 1.539 ms   (1.535 ms .. 1.543 ms)
std dev              12.80 μs   (10.27 μs .. 16.49 μs)
```

hasufell:
> Ah, yes, that "someone else" was actually me. But there wasn't any real discussion, aside from one comment by Henning. Could you try to reignite the conversation?

Given the disinterest, does this justify two threads?

hasufell:
I find the benchmark results interesting and confusing.

treeowl:
What did your actual benchmark look like?

hasufell:
```
         bench "forM" $ whnf (\s -> runIdentity $ S.forM s (\e -> pure $ e + 1)) s
```

treeowl:
You're probably running into the weird special-case optimization in `fromList` for (strictly?) increasing lists.

hasufell:
```
bench "forM" $ whnf (\s -> runIdentity $ S.forM s (\e -> pure $ e * 12 `mod` 4)) s
```

makes the foldr version go faster:

```
benchmarked forM
time                 173.3 μs   (171.7 μs .. 175.2 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 173.2 μs   (172.5 μs .. 174.2 μs)
std dev              2.743 μs   (2.188 μs .. 3.780 μs)
```

andreasabel:
`whnf` only computes the weak head normal form, you need to use `nf`, the full normal form.  (I fell into this trap before, see https://github.com/agda/agda/commit/8dae71a19c9744a46c6237590c8c186477f27e85#commitcomment-36008099 ).

andreasabel:
My opinion on this PR:
- I think that we should have `mapM = traverse` rather than two different implementation.  Also, the effects should occur in order from the least set element to the greatest. Spec:
```
mapM f s = traverse f s = Set.fromList <$> do mapM f $ Set.toList s
```
- I think we don't need `for`; this is anyway a misnamed function in `Applicative` because the reasonable analogy of `forM = flip mapM` is `for = flip map` and this is violated here.

hasufell:
> `whnf` only computes the weak head normal form, you need to use `nf`, the full normal form. (I fell into this trap before, see [agda/agda@8dae71a#commitcomment-36008099](https://github.com/agda/agda/commit/8dae71a19c9744a46c6237590c8c186477f27e85#commitcomment-36008099) ).

Thanks, will adjust!

>     * I think that we should have `mapM = traverse` rather than two different implementation.

I did that, because they have different performance characteristics I think. It appears the applicative version is faster. See above.

>     * I think we don't need `for`; this is anyway a misnamed function in `Applicative` because the reasonable analogy of `forM = flip mapM` is `for = flip map` and this is violated here.

I prefer consistency here. `for` is already well known and I use it a lot.

treeowl:
@andreasabel, could you do me a favor and keep the discussion of what to include and by what name to the libraries list? It's useful to have it all in one place and to get more exposure to comments from other users.

containers/src/Data/Set/Internal.hs
```diff
@@ -953,6 +958,30 @@ mapMonotonic :: (a->b) -> Set a -> Set b
 mapMonotonic _ Tip = Tip
 mapMonotonic f (Bin sz x l r) = Bin sz (f x) (mapMonotonic f l) (mapMonotonic f r)
 
+
+-- | /O(n*log n)/.
+--
+-- Like 'traverse' from Data.Traversable. This is less generic, since 'Set'
+-- does not have a Traversable instance.
+traverse :: (Ord b, Applicative m) => (a -> m b) -> Set a -> m (Set b)
+traverse f s0 = fmap (GHCExts.fromListN (size s0)) . DT.traverse f . toList $ s0
```
hasufell: I know there are some subtle difference on whether to use `traverse f s0` vs `traverse f`, e.g. inlining. What's the verdict here? Do we want it half-pointfree or fully pointfree even?

hasufell: @massudaw any idea?

treeowl:
@hasufell, you should bring back the previous `mapM` at least long enough to benchmark it using a monad other than `Identity`! Try `ST s`, for example.

hasufell:
>     * I think that we should have `mapM = traverse` rather than two different implementation.  Also, the effects should occur in order from the least set element to the greatest. Spec:
> 
> 
> ```
> 
> ```

In fact, this breaks ancient GHC versions where Monad is a superclass of Applicative:
https://travis-ci.org/haskell/containers/jobs/635377304#L737

hasufell:
@treeowl the results are when keeping the two different implementations. This looks significant.

---
benchmarked forM (Identity)
time                 1.542 ms   (1.536 ms .. 1.549 ms)
mean                 1.539 ms   (1.536 ms .. 1.546 ms)
std dev              16.83 μs   (10.46 μs .. 28.81 μs)

benchmarked for (Identity)
time                 163.3 μs   (161.7 μs .. 165.0 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 163.8 μs   (162.7 μs .. 165.1 μs)
std dev              3.992 μs   (3.276 μs .. 5.028 μs)

benchmarked forM (ST)
time                 1.758 ms   (1.750 ms .. 1.764 ms)
mean                 1.765 ms   (1.760 ms .. 1.774 ms)
std dev              23.95 μs   (15.22 μs .. 38.66 μs)

benchmarked for (ST)
time                 277.3 μs   (274.9 μs .. 279.5 μs)
mean                 279.8 μs   (278.0 μs .. 281.2 μs)
std dev              5.724 μs   (4.727 μs .. 7.641 μs)

treeowl:
Please let me see your `ST` benchmark code, and please include it in the benchmarks you add to the package.

hasufell:
> Please let me see your `ST` benchmark code, and please include it in the benchmarks you add to the package.

ST is already included.

treeowl:
It looks like your `ST` benchmark produces elements in strictly increasing order, which bakes in an extra advantage for `fromListN`. Would you mind tweaking it to scatter elements around more to get something more realistic?

treeowl:
So how did the ordering change affect the benchmarks?

hasufell:
> So how did the ordering change affect the benchmarks?

Not at all

treeowl:
Okay, so for now we don't need to have `mapM`, right?

hasufell:
Actually I was lying. The new benchmarks ran with the unified implementation. Here are the benchmarks for the explicit foldr (`forM`) and the fromListN (`for`):

---

```
benchmarked forM (Identity)
time                 1.538 ms   (1.509 ms .. 1.561 ms)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 1.494 ms   (1.486 ms .. 1.505 ms)
std dev              31.80 μs   (26.01 μs .. 40.72 μs)

benchmarked for (Identity)
time                 171.3 μs   (167.4 μs .. 176.8 μs)
                     0.996 R²   (0.994 R² .. 0.999 R²)
mean                 171.8 μs   (170.5 μs .. 173.4 μs)
std dev              4.873 μs   (3.835 μs .. 6.653 μs)

benchmarked forM (ST)
time                 9.340 μs   (9.246 μs .. 9.427 μs)
mean                 9.413 μs   (9.377 μs .. 9.457 μs)
std dev              132.2 ns   (107.1 ns .. 161.1 ns)

benchmarked for (ST)
time                 5.228 μs   (5.134 μs .. 5.343 μs)
mean                 5.285 μs   (5.237 μs .. 5.355 μs)
std dev              202.2 ns   (151.9 ns .. 294.4 ns)
```

I can't make any sense out of this. Why the `forM` with Identity is so slow, but the ST one not... totally escapes me.

hasufell:
I noticed when I change the test implementation to:

```sh
         bench "forM (Identity)" $ nf (\s -> runIdentity $ S.forM s (\e -> pure $ e + 1)) s
        , bench "for (Identity)" $ nf (\s -> runIdentity $ S.for s (\e -> pure $ e + 1)) s
        , bench "forM (ST)" $ nf (\s -> runST $ do
              ref <- newSTRef 0
              S.forM s (\e -> do
                              modifySTRef ref (+ 1)
                              x <- readSTRef ref
                              pure $ x + e * 36 `mod` 11)) s
        , bench "for (ST)" $ nf (\s -> runST $ do
              ref <- newSTRef 0
              S.for s (\e -> do
                             modifySTRef ref (+ 1)
                             x <- readSTRef ref
                             pure $ x + e * 36 `mod` 11)) s
```

Then the results are:

```
benchmarked forM (Identity)
time                 1.514 ms   (1.495 ms .. 1.536 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 1.496 ms   (1.490 ms .. 1.504 ms)
std dev              22.49 μs   (17.10 μs .. 31.61 μs)

benchmarked for (Identity)
time                 169.9 μs   (168.5 μs .. 171.5 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 169.1 μs   (168.0 μs .. 170.6 μs)
std dev              4.032 μs   (3.006 μs .. 5.988 μs)

benchmarked forM (ST)
time                 1.744 ms   (1.735 ms .. 1.752 ms)
mean                 1.740 ms   (1.736 ms .. 1.746 ms)
std dev              16.39 μs   (10.13 μs .. 26.54 μs)

benchmarked for (ST)
time                 1.376 ms   (1.370 ms .. 1.380 ms)
mean                 1.383 ms   (1.378 ms .. 1.390 ms)
std dev              19.05 μs   (13.22 μs .. 26.90 μs)
```

It seems small changes to the tests have significant impact.

But the forM and for seem to converge at some point.

hasufell:
> Okay, so for now we don't need to have `mapM`, right?

I think it's better to have all 4 functions. Not my decision though.

treeowl:
If it doesn't add functionality and doesn't add performance, then there's a really high bar to adding it. Generally speaking, this is because it makes something *really important* significantly more convenient (like `(!?)` as an infix version of `lookup`) or because it's rather tricky to figure out how to do it most efficiently (like `Data.Sequence.intersperse`). I'm a bit skeptical about adding any of these functions, but I'm leaning toward letting you have `traverse` if others don't speak up, since all the work you've done has lessons to teach about how to make this perform well.

containers/src/Data/Set.hs
```diff
@@ -136,6 +136,12 @@ module Data.Set (
             -- * Map
             , S.map
             , mapMonotonic
+#if __GLASGOW_HASKELL__ >= 710
```
hvr: This is quite confusing and annoying; please avoid APIs exporting things conditional on GHC or `base` versions

treeowl: @hvr, don't worry; I won't let it through like that. Please join the discussion on the libraries list about whether to add some of these.

hasufell: @hvr there are two problems with that. 7.6 and 7.8 don't have Applicative constraint on Monad class and 7.6 does not have fromListN at all, which would result in two different implementations of the same function having different performance characteristics. How do you propose to solve it?

hasufell: ping

hasufell: ping

hasufell: @hvr ping

hasufell: As there is no response from @hvr can you @treeowl give a definite list of required changes that need to be made in order for this to be merged, including how you want multiple GHC support to happen?

hasufell:
No actionable input. I've lost interest in this PR.

sjakobi:
@treeowl Since you apparently don't want to let this die, I've assigned it to you. :)

hasufell:
This is dead. Please open your own PR if you're interested to continue. I'm cleaning up stale PRs every once in a while.

PR #593, opened 2019-01-21
Fix Foldable instance for IntMap (fixes #579)

### LINKED ISSUE #579: Foldable instance for IntMap is broken? (internally inconsistent, unlawful)
Was running some property tests and noticed that the order of traversals for `Foldable` methods for `IntMap` is internally inconsistent.  `foldr`, `foldl`, `toList` appear to traverse items in order of ascending keys (which I believe is the intended behavior, for consistency with `Map k`'s instance)

```haskell
m = IM.fromList $ zip [-5..5] ['a'..]

foldr (:) [] m
-- "abcdefghijk"

traverse_ (Const . (:[])) m
-- Const "abcdefghijk"

toList m
-- "abcdefghijk"
```

However, `foldMap` and `fold` (and `foldMapWithKey`) and `traverse` (and `traverseWithKey`) appear to traverse items at positive keys first, then items at negative keys second:

```haskell
foldMap (:[]) m
-- "fghijkabcde"

traverse (Const . (:[])) m
-- Const "fghijkabcde"
```

(Note that `traverse (Const . (:[]))` is different than `traverse_ (Const . (:[])`)

The reason appears to be that an in-order traversal for `IntMap` is a positive-then-negative traversal (due to how its internal structure is based on the bitstring represenation of Int, instead of the Ord instance like for Map).  `foldr` and `foldl` explicitly correct for this (and so also the default implementation of `traverse_`), but `fold` and `foldMap` and `traverse` do not (they are implemented the same way as they are for `Map`).

The consensus in #535 seems to be that the order of traversals for maps isn't particularly significant.  However, the traversal order should at least be internally consistent within different `Foldable` methods, because of the Foldable "laws":

```haskell
foldr f z t = appEndo (foldMap (Endo . f) t ) z
foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z
```

This implies that:

```haskell
foldMap (:[]) == foldr (:) []
```

which I believe is a reasonable expectation that is currently being broken.

In order to preserve the lawfulness/expectation of consistency of the `Foldable` instance, I think one of the three options are possible:

1. Change `foldMap` and `fold` and `traverse` to match the behavior of `foldr` and `foldl` and `toList`, and traverse items in order of ascending key by correcting for the internal structure of the map.  Presumably, `foldMapWithKey` would also need to be changed as well, to traverse in order of ascending keys.  This way makes the most sense to me (making the instances consistent with `Map k`), but is a potentially breaking change to anyone using `foldMap` and `fold` with noncommutative monoids (and `traverse` with noncommutative applicatives).
2. Change `Foldable.foldr` and `Foldable.foldl` and `Foldable.toList` to match the behavior of `foldMap` and `fold` (traverse positive, then negative), by making them *different* from `Data.IntMap.foldr` and `Data.IntMap.foldl` (traverse negative, then positive).  This keeps `Data.IntMap.foldr` and `Data.IntMap.foldl` consistent with `Data.Map.foldr` and `Data.Map.foldl`, and also fixes the inconsistencies in the `Foldable` instance.  (This method would also break `Foldable.toList == elems`, but I don't think this is a particularly important thing)
3. Change `Data.IntMap.foldr` and `Data.IntMap.foldl` to match the behavior of `foldMap` and `fold` (traverse positive, then negative), and so also `Foldable.foldr` and `Foldable.foldl` to match.  This fixes the inconsistencies in the `Foldable` instances, but breaks consistency with `Data.Map.foldr` and `Data.Map.foldl`.  This method also breaks `Foldable.toList == elems`.

To summarize:

| | Keep Current Situation | Change `foldMap`, `traverse` | Change `Foldable.foldr` | Change `Foldable.foldr` and `IntMap.foldr` |
|-|-|-|-|-|
| Internally consistent | | x | x | x |
| `Foldable.foldr` = `IntMap.foldr` | x | x | | x |
| Consistent with `Map` | | x | | |
| `Foldable.toList` = `elems` | x | x | | |
| Doesn't break code using non-commutative monoids or applicatives | x | | x | x |

Note that the `Traversable` instances are internally consistent at the moment (traversing positives, the negatives), but also are technically "unlawful" because they are not consistent with `Foldable` methods.  (2) and (3) would fix this, but (1) would require the `Traversable` instance to also change.

m-renaud:
# Overview

As reported in https://github.com/haskell/containers/issues/579 the `Foldable` instance for `IntMap` is unlawful and internally inconsistent. This was caused as a result of the internal representation used by `IntMap`.

More specifically, `fold`, `foldMap`, and `traverse` (via `traverseWithKey`) always placed positively keyed entries before negative keyed ones. To fix this we need to check to see if the mask is positive or negative.

This fixes #579.

# Testing

Tested by adding new property tests, verifying they failed with the implementation at HEAD, and then passed after the changes. The property tests check that:

- `foldr (:) [] == foldMap (:[])`
- `foldrWithKey (\_ -> (:)) [] == foldMapWithKey (\_ -> (:[]))`
- `getConst (traverse (Const . (:[]))) == getConst (Foldable.traverse_ (Const . (:[])))`

/cc @mstksg

### DIFF
--- a/tests/intmap-properties.hs
+++ b/tests/intmap-properties.hs
@@ -28,6 +28,7 @@ import Test.Framework.Providers.QuickCheck2
 import Test.HUnit hiding (Test, Testable)
 import Test.QuickCheck
 import Test.QuickCheck.Function (Fun(..), apply)
+import Test.QuickCheck.Poly (A, B)
 
 default (Int)
 
@@ -1167,16 +1168,16 @@ prop_foldl' n ys = length ys > 0 ==>
       foldlWithKey' (\xs k x -> (k,x):xs) [] m == reverse (List.sort xs)
 
 prop_foldrEqFoldMap :: [(Int, String)] -> Property
-prop_foldrEqFoldMap ys = property $
+prop_foldrEqFoldMap ys =
   let m  = fromList ys
-  in  foldr (:) [] m == Data.Foldable.foldMap (:[]) m
+  in  foldr (:) [] m === Data.Foldable.foldMap (:[]) m
 
 prop_foldrWithKeyEqFoldMapWithKey :: [(Int, String)] -> Property
-prop_foldrWithKeyEqFoldMapWithKey ys = property $
+prop_foldrWithKeyEqFoldMapWithKey ys =
   let m  = fromList ys
-  in  foldrWithKey (\k v -> ((k,v):)) [] m == foldMapWithKey (\k v -> ([(k,v)])) m
+  in  foldrWithKey (\k v -> ((k,v):)) [] m === foldMapWithKey (\k v -> ([(k,v)])) m
 
-prop_FoldableTraversableCompat :: Fun String [Int] -> IntMap String -> Property
+prop_FoldableTraversableCompat :: Fun A [B] -> IntMap A -> Property
 prop_FoldableTraversableCompat fun m = foldMap f m === foldMapDefault f m
   where f = apply fun
 
--- a/tests/intmap-properties.hs
+++ b/tests/intmap-properties.hs
@@ -8,6 +8,7 @@ import Data.IntMap.Lazy as Data.IntMap hiding (showTree)
 import Data.IntMap.Internal.Debug (showTree)
 import IntMapValidity (valid)
 
+import Control.Applicative (Applicative(..))
 import Data.Monoid
 import Data.Maybe hiding (mapMaybe)
 import qualified Data.Maybe as Maybe (mapMaybe)
--- a/tests/intmap-properties.hs
+++ b/tests/intmap-properties.hs
@@ -1174,22 +1174,19 @@ prop_foldl' n ys = length ys > 0 ==>
       foldlWithKey' (\b k _ -> k + b) n m == List.foldr (+) n (List.map fst xs) &&
       foldlWithKey' (\xs k x -> (k,x):xs) [] m == reverse (List.sort xs)
 
-prop_foldrEqFoldMap :: Int -> [(Int, String)] -> Property
-prop_foldrEqFoldMap n ys = length ys > 0 ==>
-  let xs = List.nubBy ((==) `on` fst) ys
-      m  = fromList xs
+prop_foldrEqFoldMap :: [(Int, String)] -> Property
+prop_foldrEqFoldMap ys = property $
+  let m  = fromList ys
   in  foldr (:) [] m == Data.Foldable.foldMap (:[]) m
 
-prop_foldrWithKeyEqFoldMapWithKey :: Int -> [(Int, String)] -> Property
-prop_foldrWithKeyEqFoldMapWithKey n ys = length ys > 0 ==>
-  let xs = List.nubBy ((==) `on` fst) ys
-      m  = fromList xs
+prop_foldrWithKeyEqFoldMapWithKey :: [(Int, String)] -> Property
+prop_foldrWithKeyEqFoldMapWithKey ys = property $
+  let m  = fromList ys
   in  foldrWithKey (\k v -> ((k,v):)) [] m == foldMapWithKey (\k v -> ([(k,v)])) m
 
-prop_traverseEqTraverse_ :: Int -> [(Int, String)] -> Property
-prop_traverseEqTraverse_ n ys = length ys > 0 ==>
-  let xs = List.nubBy ((==) `on` fst) ys
-      m  = fromList xs
+prop_traverseEqTraverse_ :: [(Int, String)] -> Property
+prop_traverseEqTraverse_ ys = property $
+  let m  = fromList ys
   in  getConsty (Data.Traversable.traverse (Consty . (:[])) m)
       == getConsty (Data.Foldable.traverse_ (Consty . (:[])) m)
 
--- a/tests/intmap-properties.hs
+++ b/tests/intmap-properties.hs
@@ -12,7 +12,7 @@ import Data.Monoid
 import Data.Maybe hiding (mapMaybe)
 import qualified Data.Maybe as Maybe (mapMaybe)
 import Data.Ord
-import qualified Data.Foldable (foldMap)
+import Data.Foldable (foldMap)
 import Data.Function
 import Data.Traversable (Traversable(traverse), foldMapDefault)
 import Prelude hiding (lookup, null, map, filter, foldr, foldl)
--- a/tests/intmap-properties.hs
+++ b/tests/intmap-properties.hs
@@ -15,7 +15,7 @@ import qualified Data.Maybe as Maybe (mapMaybe)
 import Data.Ord
 import qualified Data.Foldable (foldMap, traverse_)
 import Data.Function
-import Data.Traversable (Traversable(traverse))
+import Data.Traversable (Traversable(traverse), foldMapDefault)
 import Prelude hiding (lookup, null, map, filter, foldr, foldl)
 import qualified Prelude (map)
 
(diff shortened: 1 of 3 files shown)


m-renaud:
Test failures from missing imports on older GHC versions are fixed now. PTAL :)

m-renaud:
> Why not foldrWithKey (\k v -> ((k, v) :)) [] == foldMapWithKey (\k v -> [(k, v)]) instead?

Good call, done.

> Also, nubBy is really inefficient, even for a test. Why not lean on Data.Map a bit and use nubOrdOn?

`nubBy` is already used extensively throughout the test files, if we decided to switch to `nubOrdOn` I think it would be best to do it as a separate PR and update all uses. That being said, having a `containers` test depend on another part of `containers` seems like it could unnecessarily introduce confusing test failures if `Data.Map` was somehow broken by a change. In general, I feel like in tests its best to only depend on the simplest possible implementation to not introduce more moving parts than strictly necessary. If it was discovered that use of `nubBy` was causing unacceptable inefficiencies in the tests then switching to `nubOrdOn` could be an option, but I'd need to be convinced that's the case.

tests/intmap-properties.hs
```diff
@@ -1156,6 +1174,25 @@ prop_foldl' n ys = length ys > 0 ==>
       foldlWithKey' (\b k _ -> k + b) n m == List.foldr (+) n (List.map fst xs) &&
       foldlWithKey' (\xs k x -> (k,x):xs) [] m == reverse (List.sort xs)
 
+prop_foldrEqFoldMap :: Int -> [(Int, String)] -> Property
+prop_foldrEqFoldMap n ys = length ys > 0 ==>
```
treeowl: `length ys > 0` is better written `not (null ys)` in general. But in these tests, I see no need to require that at all! It's better to leave it out.

m-renaud: Doh, of course! Removed.

m-renaud:
Thanks for the quick feedback and turnaround time! Tests pass locally, just waiting for travis now :)

tests/intmap-properties.hs
```diff
@@ -28,6 +31,16 @@ import Test.QuickCheck.Function (Fun(..), apply)
...
+newtype Consty a b = Consty { getConsty :: a}
+  deriving (Eq)
+
+instance Functor (Consty a) where
+  fmap _ (Consty a) = Consty a
+
+instance Monoid m => Applicative (Consty m) where
+  pure = const (Consty mempty)
+  (Consty l) <*> (Consty r) = Consty (l `mappend` r)
+
```
treeowl: Why do we need this? Why not use `Control.Applicative.Const`? For that matter, `traverse_` is defined in terms of `foldr`. So the `prop_traverseEqTraverse_` test is really about making sure the `Foldable` and `Traversable` instances are compatible. A very direct way to do that is to use `foldMapDefault`, something like this:

```haskell
prop_FoldableTraversableCompat :: Fun A [B] -> IntMap A -> Property
prop_FoldableTraversableCompat fun m = foldMap f m === foldMapDefault f m
  where f = apply fun
```

m-renaud: > Why not use Control.Applicative.Const?

I was being consistent with map-properties.hs, which defines its own `Consty`, apparently by you in https://github.com/haskell/containers/commit/7d03d76b7647dc731da2e75dcdd45f6c2e667682 :) Not sure what the original rationale was, maybe because Const needed a separate package at some point in time? /shrug

 >  A very direct way to do that.... 

Great observation, I agree that's much better than relying on a specific applicative instance. I've used `String` and `Int` in place of `A` and `B`, let me know if you had something else in mind.

treeowl: `A` and `B` come from `Test.QuickCheck.Poly` or something. They're really `Integer` under the hood, but they make it clear we don't care.

treeowl: I dunno why I used `Consty` for `map-properties` either... Hrmmmmm.....

tests/intmap-properties.hs
```diff
@@ -1156,6 +1166,20 @@ prop_foldl' n ys = length ys > 0 ==>
       foldlWithKey' (\b k _ -> k + b) n m == List.foldr (+) n (List.map fst xs) &&
       foldlWithKey' (\xs k x -> (k,x):xs) [] m == reverse (List.sort xs)
 
+prop_foldrEqFoldMap :: [(Int, String)] -> Property
+prop_foldrEqFoldMap ys = property $
```
treeowl: Nicer to use `===` than `property` with `==`, I think.

tests/intmap-properties.hs
```diff
@@ -8,11 +8,14 @@ import Data.IntMap.Lazy as Data.IntMap hiding (showTree)
 import Data.IntMap.Internal.Debug (showTree)
 import IntMapValidity (valid)
 
+import Control.Applicative (Applicative(..))
```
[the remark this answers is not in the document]
m-renaud: Nope, need to get my editor tooling set up again to recognize unused imports.

tests/intmap-properties.hs
```diff
@@ -8,11 +8,14 @@ import Data.IntMap.Lazy as Data.IntMap hiding (showTree)
...
 import Data.Monoid
 import Data.Maybe hiding (mapMaybe)
 import qualified Data.Maybe as Maybe (mapMaybe)
 import Data.Ord
+import qualified Data.Foldable (foldMap, traverse_)
```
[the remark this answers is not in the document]
m-renaud: Nope, removed.

tests/intmap-properties.hs
m-renaud: Removed.

tests/intmap-properties.hs
```diff
@@ -12,7 +12,9 @@ import Data.Monoid
 import Data.Maybe hiding (mapMaybe)
 import qualified Data.Maybe as Maybe (mapMaybe)
 import Data.Ord
+import qualified Data.Foldable (foldMap)
```
treeowl:
```suggestion
import Data.Foldable (foldMap)
```

m-renaud:
Changelog updated. I was reading the PVP and it seemed like this only change only requires a patch release bump so I created a new section in the changelog for 0.6.0.2.

Since no "new bindings, types, classes, non-orphan instances or modules (but see below) were added to the interface" it seems that version number rule 2 does not necessarily apply, so I think its fine to leave the first three version components the same. Feel free to adjust to a minor release bump if desired.

treeowl:
Yes, I agree with your versioning determination. What's not entirely clear to me in a case like this is whether it is best to deprecate the previous sub-minor versions of this minor release. We have another question to consider: should we take the time to back-port this fix to earlier versions? The bug appears to have been introduced in https://github.com/haskell/containers/commit/29d3fbcc67ea4eebc7b381ea5c59673ef1e10dc9, about *six years* ago.

m-renaud:
> whether it is best to deprecate the previous sub-minor versions of this minor release

I think deprecating 0.6.0.1 would be reasonable, but of questionable impact. As long as folks aren't specifying `==0.6.0.1` then the next time they run `cabal update` they'll pick up the new version anyways. If they're specifically running into a bug in their code because of this they can always update. It would be useful to make sure that 0.6.0.2 is included in the next Stackage LTS thought to minimize the time horizon for which folks are likely to be exposed to this bug.

> We have another question to consider: should we take the time to back-port this fix to earlier versions?

Since its been present for 6 years and it was _just_ reported, I doubt many folks have ran into this, or more likely they didn't notice or it didn't matter. If someone is using an older version of containers (and is unable to update to a newer version) runs into this, then I think that would be the time to backport it, its of questionable utility to do all the work to backport it to every minor version now.

Do we have stats for which versions of containers are still being downloaded and used? Hackage just shows the number of downloads, not the breakdown by version. It would be interesting to see who is using old versions of the package, and since great care is taken so that new releases of containers work on old GHC versions it makes it less likely that folks are using old versions of containers purely because they're on an older GHC.

m-renaud:
In any case, once the new version is released we should make an announcement in case anyone is adversely affected by the previous incorrect behaviour.

3noch:
Deprecating certainly can't *hurt* so might as well IMO.

PR #596, opened 2019-01-22
Produce better code from maskW

Fuuzetsu:
```
complement (m-1) `xor` m == -m `xor` m
```

Latter produces slightly better assembly.

```asm
	movq	%rdi, %rax
	negq	%rax
	xorq	%rdi, %rax
```

Indeed if we give

```cpp
uint64_t f(uint64_t m_aQU2) {
  return ((m_aQU2 - 1) ^ 18446744073709551615UL) ^ m_aQU2;
}
```

to clang and compile with optimisation, that's the ASM it spits out.
Translating it back to Haskell gives the solution from this branch.
Without same optimisation, even in Haskell it saves an instruction.

```hs
maskW :: Word -> Word
maskW m = complement (m-1) `xor` m

{-
	movq %rax,%rbx
	xorq $-1,%rbx
	decq %rax
	xorq %rbx,%rax
-}
```

```hs
maskMine :: Word -> Word
maskMine m = (-m) `xor` m

{-
	movq %rax,%rbx
	negq %rbx
	xorq %rax,%rbx
-}
```

### DIFF
--- a/Data/IntMap/Internal.hs
+++ b/Data/IntMap/Internal.hs
@@ -1114,7 +1114,7 @@ withoutKeys t1@(Bin p1 m1 _ _) (IntSet.Tip p2 bm2) =
     let minbit = bitmapOf p1
         lt_minbit = minbit - 1
         maxbit = bitmapOf (p1 .|. (m1 .|. (m1 - 1)))
-        gt_maxbit = maxbit `xor` complement (maxbit - 1)
+        gt_maxbit = (-maxbit) `xor` maxbit
     -- TODO(wrengr): should we manually inline/unroll 'updatePrefix'
     -- and 'withoutBM' here, in order to avoid redundant case analyses?
     in updatePrefix p2 t1 $ withoutBM (bm2 .|. lt_minbit .|. gt_maxbit)
@@ -3333,7 +3333,7 @@ mask i m
 -- bit @m@.
 maskW :: Nat -> Nat -> Prefix
 maskW i m
-  = intFromNat (i .&. (complement (m-1) `xor` m))
+  = intFromNat (i .&. ((-m) `xor` m))
 {-# INLINE maskW #-}
 
 -- | Does the left switching bit specify a shorter prefix?


treeowl:
Good point! That *should* be better. But since you're micro-optimizing here, I think it would pay to get some benchmark results too, to make sure that you don't happen by chance to damage alignment and actually reduce performance. As long as benchmarks either improve or stay the same, I agree we should make this change.

Fuuzetsu:
I couldn't conclusively confirm which version is better just by benchmarking these separately. I think I'll have to run the containers benchmark and conclude from that.

treeowl:
Could you zoom in on those `lookup` benchmarks? Stick some other test in front to warm them up a bit, and run them harder. If that's a real regression (even if fairly minor), then I'm a bit concerned.

treeowl:
For future reference: I personally prefer to see benchmarks woven together: old-new, old-new, old-new, well labeled. Trying to compare by scrolling up and down is tiresome.

Fuuzetsu:
I think it's noise on my machine.

I added a "warmup" in front (it's just the same benchmark under different name) and I'm running with `-L 60`

without my change
```
benchmarking lookup_warmup
time                 186.0 μs   (185.5 μs .. 186.5 μs)
mean                 185.8 μs   (185.3 μs .. 186.4 μs)
std dev              2.706 μs   (2.027 μs .. 3.571 μs)

benchmarking lookup
time                 185.9 μs   (185.3 μs .. 186.4 μs)
mean                 185.4 μs   (185.1 μs .. 185.9 μs)
std dev              2.034 μs   (1.506 μs .. 3.355 μs)
```
with my change:
```
benchmarking lookup_warmup
time                 184.2 μs   (183.5 μs .. 184.9 μs)
mean                 183.4 μs   (183.0 μs .. 184.0 μs)
std dev              2.238 μs   (1.905 μs .. 2.833 μs)

benchmarking lookup
time                 185.2 μs   (184.7 μs .. 185.7 μs)
mean                 185.2 μs   (184.6 μs .. 185.9 μs)
std dev              2.974 μs   (2.459 μs .. 3.662 μs)
```

>For future reference: I personally prefer to see benchmarks woven together: old-new, old-new, old-new, well labeled. Trying to compare by scrolling up and down is tiresome.

I just opened the two side by side, no jumping around needed.

treeowl:
All righty then! Thanks.

PR #600, opened 2019-01-31
Speed up some Cartesian products

treeowl:
`Data.Set.cartesianProduct` was considerably slower when its first
argument was much larger than its second than it was in the opposite
situation. Hack around that to make it just as fast.

### DIFF
--- a/Data/Set/Internal.hs
+++ b/Data/Set/Internal.hs
@@ -1727,8 +1727,21 @@ powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
 --
 -- @since 0.5.11
 cartesianProduct :: Set a -> Set b -> Set (a, b)
-cartesianProduct as bs =
-  getMergeSet $ foldMap (\a -> MergeSet $ mapMonotonic ((,) a) bs) as
+cartesianProduct as bs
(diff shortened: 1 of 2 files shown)


Data/Set/Internal.hs
```diff
@@ -1727,8 +1727,21 @@ powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
...
+  -- When the sizes are very different (e.g., 10 vs. 2000), it's quite a bit
+  -- better to fold over the smaller set. I'm not exactly sure why that is so
+  -- in non-degenerate cases, but we might as well do it the fast way.
+  -- Note: Edward Kmett came up with this implementation (without the size
+  -- check) off the top of his head; he does not claim it is optimal. It might
+  -- be worth trying to find a better one, but it's definitely not obvious
+  -- how one might do so.
+  --
+  -- TODO: Attempt to calculate, or at least empirically estimate, the big-O
+  -- performance of this function.
```
treeowl: OK, so the mapping itself takes `O(m*n)` time. What about the merging? Well, we're doing `m - 1` merges, but I'm not sure how those merge times add up.

treeowl:
Um.... This is bogus. *headshake*. Thought I got the test passing.

PR #601, opened 2019-01-31
Speed up special cases of Cartesian products

treeowl:
If the second argument of `Data.Set.cartesianProduct` happens to
be empty or a singleton, then we don't need to merge a bunch of trees.
So let's not.

### DIFF
--- a/Data/Set/Internal.hs
+++ b/Data/Set/Internal.hs
@@ -1727,6 +1727,15 @@ powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
 --
 -- @since 0.5.11
 cartesianProduct :: Set a -> Set b -> Set (a, b)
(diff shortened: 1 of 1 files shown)


treeowl:
Ooh.... We don't know whether this algorithm is big-O optimal, but I think there's a very simple one that is obviously so. Suppose we want `cartesianProduct as bs`. We can calculate the size as `size as * size bs`. We can calculate the elements, in order, as `[(a,b) | a <- toList as, b <- toList bs]`. Now all that remains is to construct a perfectly balanced `Set` from that list with the known number of elements, which we can surely do in `O(size as * size bs)` time, which is also the order of the `mapMonotonic`s and  therefore of the total cost.

treeowl:
Empirically, though, Kmett's implementation is faster than I've been able to make the alternative, thus far. Hmph.

PR #605, opened 2019-02-07
Document the Semigroup for Map

chris-martin:
The Map type has many possible semigroup implementations, so the documentation needs to explain which one was chosen.


3noch:
Perhaps it would be worth mentioning https://hackage.haskell.org/package/monoidal-containers here to help people find a version they may prefer.

chris-martin:
Good idea. I think it might be too much of a distraction right there - Maybe it would be better to have links to related packages in the Cabal description?

3noch:
Seems sensible.

treeowl:
Thanks!

treeowl:
Er... Did you check the `IntMap` docs?

PR #610, opened 2019-03-05
Make strict IntMap merges strict

### LINKED ISSUE #609: Strict Merge not strict
Hi,

I believe that the `merge` function for `IntMap` is not properly strict.  Consider the following example:
```haskell
import qualified Data.IntMap.Merge.Strict as IntMap
import qualified Data.IntMap.Strict as IntMap

myMap = IntMap.fromList [(1,1),(2,2)]

seq (IntMap.merge IntMap.dropMissing IntMap.preserveMissing (IntMap.zipWithMatched (\t x y -> undefined)) myMap myMap) ()

seq (IntMap.merge IntMap.dropMissing IntMap.preserveMissing (IntMap.zipWithMatched undefined) myMap myMap) ()
```
When I evaluate either of the two `seq` lines, I get `()` instead of an error as I would expect.

treeowl:
* Make `Data.IntMap.Merge.Strict` tactics (except `preserveMissing`)
  strict.

* Add a strict `Data.Map.Merge.Strict` `preserveMissing'` tactic.
  We may want to just call this `preserveMissing`....

Fixes #609


treeowl:
@dwincort, please have a look.

dwincort:
I haven't looked super carefully (and it looks like you need to add a missing `Control.Applicative` import), but overall, this looks pretty good to me.

Thanks for being so prompt!

PR #612, opened 2019-03-18
benchmarks comparer: Use `/usr/bin/env` to find perl

Ericson2314:
This is needed on NixOS. FHS does say if perl is installed it shoul be
found at `/usr/bin/perl`, so can't say this is more portable in general,
but doesn't hurt FHS systems either


treeowl:
Can you explain what the new code does?

Ericson2314:
Sure. `/usr/bin/env` looks up its argument on the `$PATH`, and `exec`s that. This is helpful for NixOS, since Perl will be in some non-standard location.

PR #615, opened 2019-03-27
Improve isSubsetOf

### LINKED ISSUE #614: Improve isSubsetOf
Currently,  `isSubsetOf s1 s2` checks that `size s1 <= size s2`, and if so invokes `isSubsetOfX` which recursively splits `s2` along the structure of `s1` without further size checks. I doubt this is the right approach.

### Wrong side

Splitting a set takes logarithmic time and allocation, so I believe we should split the *smaller* set along the structure of the *larger* one whenever the sizes are different. This is most easily accomplished by splitting `s1` along the structure of `s2` instead. That changes the shape of the algorithm somewhat, but I think that should be okay.

### More size tests are probably better

A size check takes constant time and no allocation, and can save a split and potentially a lot of comparisons. So I think we should perform recursive size checks. Once we split `s1` at the root of `s2`, I think we should check that the left subtree of `s2` is no smaller than the left side of the split and that the right subtree of `s2` is no smaller than the right side of the split before proceeding recursively. For example, suppose we have `{1,2,3,4,5,6,7}` with a root of `4` and want to check if `{0,1,2,3,4,5,6}` is a subset. After the first split, we'll immediately see that it's not, because `{0,1,2,3}` has more elements than `{1,2,3}`.

treeowl:
* Add recursive size tests to `Data.Set.isSubsetOf`.

* Add a special case for singleton subsets to avoid extra splits
  at all the leaves.

Closes #614

### DIFF
--- a/Data/Map/Internal.hs
+++ b/Data/Map/Internal.hs
@@ -2742,8 +2742,8 @@ isSubmapOfBy f t1 t2
 -- size test. See Data.Set.Internal.isSubsetOfX for notes on
 -- implementation and analysis.
 submap' :: Ord a => (b -> c -> Bool) -> Map a b -> Map a c -> Bool
-submap' f Tip _ = True
-submap' f _ Tip = False
+submap' _ Tip _ = True
+submap' _ _ Tip = False
 submap' f (Bin 1 kx x _ _) t
   = case lookup kx t of
       Just y -> f x y
--- a/Data/Map/Internal.hs
+++ b/Data/Map/Internal.hs
@@ -2727,22 +2727,33 @@ isSubmapOf m1 m2 = isSubmapOfBy (==) m1 m2
  > isSubmapOfBy (<)  (fromList [('a',1)]) (fromList [('a',1),('b',2)])
  > isSubmapOfBy (==) (fromList [('a',1),('b',2)]) (fromList [('a',1)])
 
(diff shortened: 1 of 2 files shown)


treeowl:
Specialization isn't working right for `isSubmapOf` as yet. Still working on it.

treeowl:
`disjoint` also has terribly pessimistic bounds we can tighten.

Data/Map/Internal.hs
```diff
@@ -2727,22 +2727,33 @@ isSubmapOf m1 m2 = isSubmapOfBy (==) m1 m2
...
 {-# INLINABLE isSubmapOfBy #-}
 #endif
 
+-- Test whether a map is a submap of another without the *initial*
+-- size test. See Data.Set.Internal.isSubsetOfX for notes on
+-- implementation and analysis.
 submap' :: Ord a => (b -> c -> Bool) -> Map a b -> Map a c -> Bool
-submap' _ Tip _ = True
-submap' _ _ Tip = False
+submap' f Tip _ = True
```
treeowl:
```suggestion
submap' _ Tip _ = True
```

Data/Map/Internal.hs
```diff
@@ -2727,22 +2727,33 @@ isSubmapOf m1 m2 = isSubmapOfBy (==) m1 m2
...
+submap' f _ Tip = False
```
treeowl:
```suggestion
submap' _ _ Tip = False
```

PR #626, opened 2019-04-12
Convert to cabal.project

phadej:
- Move the lib into `containers`
- Move the tests and benchmarks into `containers-tests`
- Make some adjustments so -Whome-modules is happy
- `ln -s` stuff from `containers` to `containers-tests`
  so things work; use containers-tests to be the library under test and
  benchmarking


phadej:
Well, seems something changed in `master`, but the idea is this.

phadej:
This is one way. Let me think over weekend if I come up with better one.

phadej:
Travis: Green

hvriedel:
The approach looks reasonable to me

treeowl:
Is there a reason to move everything into an `src` directory?

phadej:
`ln -s src` is simpler and more future-proof (if one adds new top-level namespaces).

Also, often enough not having CWD as source directory prevents some unfortunate interactions.

phadej:
I rebased once again. Won't do anymore.

treeowl:
Thanks a lot for all the work. I'll merge, and maybe we'll find a way to cut the remaining redundancy some other year.

PR #627, opened 2019-04-13
Group/Order set-operations-* benchmarks better.

AndreasPK:
Benchmarks are now run ordered by method/data_variant/data_size/swapped.

Turns out the output of -l runs sort on the names, so that wasn't obvious until I ran all the benchmarks.

### DIFF
--- a/benchmarks/SetOperations/SetOperations.hs
+++ b/benchmarks/SetOperations/SetOperations.hs
@@ -4,7 +4,8 @@
 module SetOperations (benchmark) where
 
 import Gauge (bench, defaultMain, whnf)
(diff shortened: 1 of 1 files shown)


treeowl:
What base library versions were previously supported by the benchmarks? I honestly don't know. But `sortOn` is fairly new, and we don't really need it, so *if* that single-handedly limits us, I'd rather use `sortBy`. If other things limit us already, then fine.

AndreasPK:
sortOn was added in GHC 7.10 so quite a while ago. But there is no need to break it knowingly :)

treeowl:
Thanks!

PR #635, opened 2019-05-13
docs: fixed typo in Set's "map" function

unclechu:


treeowl:
Those weren't typos. "Noting" is a gerund form of the verb "note", and that's what is intended.

PR #646, opened 2019-06-16
Fixed example uses of map merge

olligobber:


treeowl:
There's still a `diffMapWithKey` that presumably needs to go. Thanks for catching the rest!

olligobber:
I fixed that and another error I had introduced

treeowl:
Thanks!

PR #651, opened 2019-06-27
fix documentation of isProperSubmapOfBy

dmwit:
The `isProperSubmapOfBy` function takes two maps with different contained value types. Its documentation specifies its behavior in terms of whether the two maps are equal, which is clearly not doable in Haskell. I think it should be specified in terms of whether the two maps have the same keys or not instead.

This tiny commit makes that documentation change.

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -2309,7 +2309,7 @@ isProperSubmapOf m1 m2
 
 {- | /O(n+m)/. Is this a proper submap? (ie. a submap but not equal).
  The expression (@'isProperSubmapOfBy' f m1 m2@) returns 'True' when
- @m1@ and @m2@ are not equal,
+ @keys m1@ and @keys m2@ are not equal,
  all keys in @m1@ are in @m2@, and when @f@ returns 'True' when
  applied to their respective values. For example, the following
  expressions are all 'True':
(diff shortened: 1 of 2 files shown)


treeowl:
Thanks. I am somewhat curious whether anyone actually uses that function (it seems like a strange thing to want to do), but it's always nice to have accurate documentation! Speaking of accurate docs.... The time bounds given for a number of merging operations in `Data.Map` and `Data.Set`, and probably also some operations in `Data.Sequence`, are incorrect in certain edge cases. Empty sets and maps, and probably `0` and `length s - 1` sequence indices, can make the given bounds bogus (`log 0`, `n/0`, operations taking 0 time). I've refrained from fixing these for the most part because I don't know how to do so without making messy formulas, but if you have any ideas, I'd love to use them.

dmwit:
Well, I don't have a use for `isProperSubmapOf[By]`, but I do have a use for `isSubmapOf[By]`: if you imagine a game board represented as a `Map Position Land` and a placement of pieces represented as `Map Position Piece`, and some pieces can only be placed on certain types of land, then `isSubmapOfBy` is just what you want for checking that all the pieces are in bounds of the board and above compatible land types.

Since you're in asymptotics land, the odd edge cases don't matter: you get to completely throw away the behavior of the function under any finite input bound you like.

Issue #288, opened 2016-06-18
IntSet.fromList slower than repeated IntSet.insert


amigalemming:
I ran the following benchmark:

```
module Main where

import Criterion.Main (defaultMain, bgroup, bench, nf, )

import qualified Data.IntSet as IntSet; import Data.IntSet (IntSet)
import qualified Data.List as List
import Data.Word (Word64)


{-# INLINE randomIndices #-}
randomIndices :: Int -> [Int]
randomIndices k =
   map (fromIntegral . flip mod (fromIntegral k)) $
   iterate (\x -> mod (x*40692) 2147483399) (1::Word64)

constructIntSet :: Int -> Int -> IntSet
constructIntSet n k =
   IntSet.fromList $ take n $ randomIndices k

insertsIntSet :: Int -> Int -> IntSet
insertsIntSet n k =
   List.foldl' (flip IntSet.insert) IntSet.empty $
   take n $ randomIndices k


main :: IO ()
main =
   defaultMain $
      (bgroup "construct" $
         bench "1000" (nf (constructIntSet 500) 1000) :
         bench "3000" (nf (constructIntSet 500) 3000) :
         bench "9000" (nf (constructIntSet 500) 9000) :
         []) :
      (bgroup "insert" $
         bench "1000" (nf (insertsIntSet 500) 1000) :
         bench "3000" (nf (insertsIntSet 500) 3000) :
         bench "9000" (nf (insertsIntSet 500) 9000) :
         []) :
      []
```

and got:

```
benchmarking construct/1000
time                 53.36 μs   (53.24 μs .. 53.55 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 53.46 μs   (53.32 μs .. 53.73 μs)
std dev              638.6 ns   (367.6 ns .. 1.215 μs)

benchmarking construct/3000
time                 66.19 μs   (65.96 μs .. 66.54 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 66.21 μs   (66.09 μs .. 66.35 μs)
std dev              422.9 ns   (350.3 ns .. 547.7 ns)

benchmarking construct/9000
time                 81.97 μs   (81.67 μs .. 82.39 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 83.00 μs   (82.68 μs .. 83.29 μs)
std dev              1.052 μs   (914.6 ns .. 1.202 μs)

benchmarking insert/1000
[...]

PR #653, opened 2019-07-03
Speed up fromList for IntMap

treeowl:
Make `fromList` and `fromListWithKey` for `IntMap` smarter. Rather
than rebuilding the path from the root for each element, insert
as many elements as possible into each subtree before backing out.

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -3078,10 +3078,60 @@ foldlFB = foldlWithKey
 -- > fromList [(5,"c"), (3,"b"), (5, "a")] == fromList [(5,"a"), (3,"b")]
 
 fromList :: [(Key,a)] -> IntMap a
-fromList xs
-  = Foldable.foldl' ins empty xs
-  where
-    ins t (k,x)  = insert k x t
+fromList = insertAll Nil
+
+-- [Note: fromList]
+--
+-- The obvious way to build a map from a list is just to fold over the list
+-- inserting each entry into the accumulator map. The problem is that this
+-- rebuilds the path from the root *every single time*. To avoid this, we
+-- insert as many elements as we can into the current subtree, backing out
+-- one level at a time when necessary.
+
+data Inserted a = Inserted !(IntMap a) ![(Key, a)]
+
+insertAll :: IntMap a -> [(Key, a)] -> IntMap a
+insertAll m [] = m
+insertAll m ((k,x) : kxs)
+  | Inserted m' r <- insertSome m k x kxs
+  = insertAll m' r
+
+-- | Insert at least one entry into an 'IntMap'. If others fit
+-- inside, insert them too. Return the new map and remaining
+-- values.
+insertSome :: IntMap a -> Key -> a -> [(Key, a)] -> Inserted a
+insertSome t@(Bin p m l r) !k x kxs
+  | nomatch k p m
+  = insertMany (link k (Tip k x) p t) kxs
+
+  | zero k m
+  , Inserted l' kxs' <- insertSome l k x kxs
+  = insertMany (Bin p m l' r) kxs'
+
+  | Inserted r' kxs' <- insertSome r k x kxs
+  = insertMany (Bin p m l r') kxs'
+
+insertSome t@(Tip ky _) k x kxs
+  | k == ky
+  = insertMany (Tip k x) kxs
+  | otherwise
+  = insertMany (link k (Tip k x) ky t) kxs
+
+insertSome Nil k x kxs = insertMany (Tip k x) kxs
+
+-- | Try to insert some entries into an 'IntMap', but only if
+-- they fit
+insertMany :: IntMap a -> [(Key, a)] -> Inserted a
+insertMany t [] = Inserted t []
+insertMany t@(Bin p m _ _) kxs@((k, x) : kxs')
+  | nomatch k p m
+  = Inserted t kxs
+  | otherwise
+  = insertSome t k x kxs'
+insertMany t@(Tip ky _) kxs@((k, x) : kxs')
+  | k==ky         = insertMany (Tip k x) kxs'
+  | otherwise     = Inserted t kxs
+insertMany Nil kxs = Inserted Nil kxs
 
 -- | /O(n*min(n,W))/. Create a map from a list of key\/value pairs with a combining function. See also 'fromAscListWith'.
 --
@@ -3099,10 +3149,60 @@ fromListWith f xs
 -- > fromListWithKey f [] == empty
 
 fromListWithKey :: (Key -> a -> a -> a) -> [(Key,a)] -> IntMap a
-fromListWithKey f xs
-  = Foldable.foldl' ins empty xs
-  where
-    ins t (k,x) = insertWithKey f k x t
+-- See [Note: fromList]
+fromListWithKey f = insertAllWithKey f Nil
+
+insertAllWithKey
+  :: (Key -> a -> a -> a)
+  -> IntMap a -> [(Key, a)] -> IntMap a
+insertAllWithKey _f m [] = m
+insertAllWithKey f m ((k,x) : kxs)
+  | Inserted m' r <- insertSomeWithKey f m k x kxs
+  = insertAllWithKey f m' r
+
+-- | Insert at least one entry into an 'IntMap'. If others fit
+-- inside, insert them too. Return the new map and remaining
+-- values.
+insertSomeWithKey
+  :: (Key -> a -> a -> a)
+  -> IntMap a -> Key -> a -> [(Key, a)] -> Inserted a
+insertSomeWithKey f t@(Bin p m l r) !k x kxs
+  | nomatch k p m
+  = insertManyWithKey f (link k (Tip k x) p t) kxs
+
+  | zero k m
+  , Inserted l' kxs' <- insertSomeWithKey f l k x kxs
+  = insertManyWithKey f (Bin p m l' r) kxs'
+
+  | Inserted r' kxs' <- insertSomeWithKey f r k x kxs
+  = insertManyWithKey f (Bin p m l r') kxs'
+
+insertSomeWithKey f t@(Tip ky y) k !x kxs
+  | k == ky
+  , y' <- f k x y
+  = insertManyWithKey f (Tip k y') kxs
+  | otherwise
+  = insertManyWithKey f (link k (Tip k x) ky t) kxs
+
+insertSomeWithKey f Nil k x kxs = insertManyWithKey f (Tip k x) kxs
+
+-- | Try to insert some entries into an 'IntMap', but only if
+-- they fit
+insertManyWithKey
+  :: (Key -> a -> a -> a)
+  -> IntMap a -> [(Key, a)] -> Inserted a
+insertManyWithKey _f t [] = Inserted t []
+insertManyWithKey f t@(Bin p m _ _) kxs@((k, x) : kxs')
+  | nomatch k p m
+  = Inserted t kxs
+  | otherwise
+  = insertSomeWithKey f t k x kxs'
+insertManyWithKey f t@(Tip ky y) kxs@((k, x) : kxs')
+  | k==ky
+  , y' <- f k x y
+  = insertManyWithKey f (Tip k y') kxs'
+  | otherwise     = Inserted t kxs
+insertManyWithKey _f Nil kxs = Inserted Nil kxs
 
 -- | /O(n)/. Build a map from a list of key\/value pairs where
 -- the keys are in ascending order.
--- a/containers/src/Data/IntMap/Strict/Internal.hs
+++ b/containers/src/Data/IntMap/Strict/Internal.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE PatternGuards #-}
 
 #include "containers.h"
 
@@ -1066,10 +1067,52 @@ fromSet f (IntSet.Tip kx bm) = buildTree f kx bm (IntSet.suffixBitMask + 1)
 -- > fromList [(5,"c"), (3,"b"), (5, "a")] == fromList [(5,"a"), (3,"b")]
 
 fromList :: [(Key,a)] -> IntMap a
-fromList xs
-  = Foldable.foldl' ins empty xs
-  where
-    ins t (k,x)  = insert k x t
+fromList = insertAll Nil
+
+data Inserted a = Inserted !(IntMap a) ![(Key, a)]
+
+insertAll :: IntMap a -> [(Key, a)] -> IntMap a
+insertAll m [] = m
+insertAll m ((k,x) : kxs)
+  | Inserted m' r <- insertSome m k x kxs
+  = insertAll m' r
+
+-- | Insert at least one entry into an 'IntMap'. If others fit
+-- inside, insert them too. Return the new map and remaining
+-- values.
+insertSome :: IntMap a -> Key -> a -> [(Key, a)] -> Inserted a
+insertSome t@(Bin p m l r) !k x kxs
+  | nomatch k p m
+  = insertMany (link k (Tip k x) p t) kxs
+
+  | zero k m
+  , Inserted l' kxs' <- insertSome l k x kxs
+  = insertMany (Bin p m l' r) kxs'
+
+  | Inserted r' kxs' <- insertSome r k x kxs
+  = insertMany (Bin p m l r') kxs'
+
+insertSome t@(Tip ky _) k !x kxs
+  | k == ky
+  = insertMany (Tip k x) kxs
+  | otherwise
+  = insertMany (link k (Tip k x) ky t) kxs
+
+insertSome Nil k x kxs = insertMany (Tip k x) kxs
+
+-- | Try to insert some entries into an 'IntMap', but only if
+-- they fit
+insertMany :: IntMap a -> [(Key, a)] -> Inserted a
+insertMany t [] = Inserted t []
+insertMany t@(Bin p m _ _) kxs@((k, x) : kxs')
+  | nomatch k p m
+  = Inserted t kxs
+  | otherwise
+  = insertSome t k x kxs'
+insertMany t@(Tip ky _) kxs@((k, x) : kxs')
+  | k==ky         = x `seq` insertMany (Tip k x) kxs'
+  | otherwise     = Inserted t kxs
+insertMany Nil kxs = Inserted Nil kxs
 
 -- | /O(n*min(n,W))/. Create a map from a list of key\/value pairs with a combining function. See also 'fromAscListWith'.
 --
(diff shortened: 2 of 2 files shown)


treeowl:
@jwaldmann, your review/benchmarking would be greatly appreciated.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3078,10 +3078,60 @@ foldlFB = foldlWithKey
...
+insertMany :: IntMap a -> [(Key, a)] -> Inserted a
+insertMany t [] = Inserted t []
+insertMany t@(Bin p m _ _) kxs@((k, x) : kxs')
+  | nomatch k p m
+  = Inserted t kxs
+  | otherwise
+  = insertSome t k x kxs'
+insertMany t@(Tip ky _) kxs@((k, x) : kxs')
+  | k==ky         = insertMany (Tip k x) kxs'
+  | otherwise     = Inserted t kxs
```
treeowl: We should use `insertSome` for the `Tip` case as well, for simplicity.

treeowl:
Interestingly, it seems that the new `fromList` implementation is faster than `fromAscList`. Surprisingly (to me), it's sometimes *much* faster. So if we replace `fromList`, we should probably define `fromAscList = fromList` unless and until someone finds a faster implementation of `fromAscList`.

treeowl:
Hmm..... I have the feeling that part of the trouble with `fromAscList` is that `fromDistinctAscList` is (probably) less efficient than it can be. It uses an *explicit* stack in a rather confusing fashion. If we could make it use the GHC reduction stack instead, that might cut allocation and speed things up while also making the code less hard to read. But it's so convoluted that I can't really work out the control flow....

jwaldmann:
Hi. I was reading your code and did some experiments.

Code: the description should make it more clear that `insertSome/Many` inserts elements as long as they fit, and stops when encountering the first element that does not (and returns everything after that). I had to scan the source to confirm. The wording of your comment would also allow the interpretation that all of the list is scanned for elements that fit. Of course we should not do that. Or should we? My first reaction is "that can become quadratic" but I am not sure.

Experiment: since the algorithm does not have look-ahead, it can be fooled. In the following, I am using
```
hard 3 = [0,8,1,9,2,10,3,11,4,12,5,13,6,14,7,15]
```
This is a  ghci session on Data.IntMap.Internal. Measurements are totally unscientific ...
```
easy e = [0 :: Int ..2^(e+1)-1]
m [] ys = ys ; m (x:xs) ys = x : m ys xs
hard e = m [0 :: Int ..2^e-1] [2^e .. 2^(e+1)-1]

:set +s

size $ fromList $ zip (easy 18) $ repeat ()  -- 7 seconds
size $ fromList $ zip (hard 18) $ repeat () -- 27 seconds

-- compare with naive fromList implementation:
fromList0 = Foldable.foldl' (\m (k,v) -> insert k v m) empty

size $ fromList0 $ zip (easy 18) $ repeat () -- 16 seconds
size $ fromList0 $ zip (hard 18) $ repeat () -- 17 seconds

-- compare with from*AscList
size $ fromAscList $ zip (easy 18) $ repeat () -- 6 sec
size $ fromDistinctAscList  $ zip (easy 18) $ repeat () -- 5 sec
```

Do you want to export `insertAll`? Of course `insertAll t kvs = union t (fromList kvs)` but it may be faster.
```
evens e = [0 :: Int , 2 .. 2^e-1]
odds e = [1 :: Int , 3 .. 2^e-1]

t = fromList $ zip (evens 19) $ repeat ()
size t -- just to evaluate it

size $ insertAll t $ zip (odds 19) $ repeat () -- 5 seconds
size $ union t $ fromList (zip (odds 19) $ repeat ()) -- 6 seconds
```

treeowl:
Thanks, @jwaldmann! If your results are for a `-O2`-compiled `Data.IntMap.Internal`, then the bad-case slowdown from 17 seconds to 27 seconds seems a bit unfortunate, though probably tolerable. Do you or @int-e have any ideas for cutting that down? I have to wonder if unproductive `insertMany` calls are to blame. In that case, we pattern match on the list, then throw away the result of doing so only to match on it again immediately. Perhaps there's some way to avoid that (maybe with unboxed sums)? It's also a bit sketchy to call these insertion functions on trees we *just* created.

treeowl:
Oh, and yes, I would prefer to export `insertAll`, but that will require some mailing list discussion to settle on a name.

jwaldmann:
> -O2 compiled 

they were for `ghci`, now I compiled them (https://github.com/jwaldmann/containers/tree/intmap-fromList) and get these numbers. The difference (between `contiguous` and `interleaved`) is smaller than with `ghci`  (where `interleaved` is a more descriptive name for `[0,2^e,1,2^e+1,..]`)
For (pseudo) random data, times are quite high - but don't differ from previous implementation.

```
stack bench --resolver=lts-13.27  containers-tests:intmap-benchmarks --ba "-m pattern fromList"
containers-tests> benchmarks
Running 1 benchmarks...
Benchmark intmap-benchmarks: RUNNING...
benchmarked pointwise/fromList/contiguous
time                 6.568 ms   (6.474 ms .. 6.694 ms)
mean                 6.388 ms   (6.280 ms .. 6.469 ms)
std dev              300.2 μs   (217.6 μs .. 439.2 μs)
variance introduced by outliers: 21% (moderately inflated)

benchmarked pointwise/fromList/interleaved
time                 21.05 ms   (20.77 ms .. 21.32 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 20.28 ms   (19.48 ms .. 20.56 ms)
std dev              986.3 μs   (459.2 μs .. 1.921 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarked pointwise/fromList/sparse
time                 7.084 ms   (6.953 ms .. 7.241 ms)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 6.603 ms   (6.466 ms .. 6.705 ms)
std dev              352.3 μs   (257.0 μs .. 522.1 μs)
variance introduced by outliers: 25% (moderately inflated)

benchmarked pointwise/fromList/random
time                 34.74 ms   (34.21 ms .. 35.22 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 34.33 ms   (33.80 ms .. 34.79 ms)
std dev              1.024 ms   (663.2 μs .. 1.433 ms)

benchmarked pointwise/fromList_via_foldl_insert/contiguous
time                 8.886 ms   (8.758 ms .. 9.022 ms)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 8.643 ms   (8.529 ms .. 8.733 ms)
std dev              282.4 μs   (207.7 μs .. 389.0 μs)

benchmarked pointwise/fromList_via_foldl_insert/interleaved
time                 18.84 ms   (18.36 ms .. 19.39 ms)
mean                 18.46 ms   (18.15 ms .. 18.72 ms)
std dev              670.2 μs   (488.1 μs .. 972.3 μs)

benchmarked pointwise/fromList_via_foldl_insert/sparse
time                 9.198 ms   (9.075 ms .. 9.353 ms)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 9.097 ms   (8.987 ms .. 9.184 ms)
std dev              277.5 μs   (200.0 μs .. 376.3 μs)
variance introduced by outliers: 10% (moderately inflated)

benchmarked pointwise/fromList_via_foldl_insert/random
time                 33.30 ms   (32.80 ms .. 33.76 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 33.55 ms   (33.27 ms .. 33.83 ms)
std dev              605.4 μs   (466.4 μs .. 791.3 μs)

Benchmark intmap-benchmarks: FINISH


```

treeowl:
Could you do me a favor and also run some tests with longer lists? I saw much bigger differences for contiguous data in my own testing, where I was using millions of elements.

jwaldmann:
one millon:
```
benchmarked 2^20/pointwise/contiguous/fromList                                                          
time                 34.67 ms   (33.36 ms .. 35.76 ms)                                                  
                     0.996 R²   (0.992 R² .. 0.999 R²)                                                  
mean                 33.09 ms   (31.78 ms .. 34.01 ms)                                                  
std dev              2.167 ms   (1.404 ms .. 2.926 ms)                                                  
 
benchmarking 2^20/pointwise/contiguous/fromList_via_foldl_insert ... took 9.431 s, total 56 iterations
benchmarked 2^20/pointwise/contiguous/fromList_via_foldl_insert
time                 148.2 ms   (140.6 ms .. 154.7 ms)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 162.6 ms   (157.9 ms .. 167.7 ms)
std dev              8.422 ms   (5.956 ms .. 13.63 ms)

benchmarking 2^20/pointwise/interleaved/fromList ... took 12.49 s, total 56 iterations
benchmarked 2^20/pointwise/interleaved/fromList
time                 211.5 ms   (209.6 ms .. 213.4 ms)
mean                 214.7 ms   (213.7 ms .. 215.7 ms)
std dev              1.685 ms   (1.233 ms .. 2.245 ms)

benchmarking 2^20/pointwise/interleaved/fromList_via_foldl_insert ... took 9.429 s, total 56 iterations
benchmarked 2^20/pointwise/interleaved/fromList_via_foldl_insert
time                 148.0 ms   (136.0 ms .. 155.3 ms)
                     0.994 R²   (0.985 R² .. 0.999 R²)
mean                 165.5 ms   (158.6 ms .. 173.5 ms)
std dev              12.84 ms   (9.967 ms .. 18.18 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarked 2^20/pointwise/sparse/fromList
time                 30.37 ms   (28.97 ms .. 31.55 ms)
                     0.995 R²   (0.991 R² .. 0.998 R²)
mean                 29.69 ms   (28.51 ms .. 30.51 ms)
std dev              2.073 ms   (1.329 ms .. 2.924 ms)
variance introduced by outliers: 25% (moderately inflated)

benchmarking 2^20/pointwise/sparse/fromList_via_foldl_insert ... took 10.16 s, total 56 iterations
benchmarked 2^20/pointwise/sparse/fromList_via_foldl_insert
time                 171.2 ms   (169.8 ms .. 172.9 ms)
mean                 174.0 ms   (173.1 ms .. 175.4 ms)
std dev              1.874 ms   (1.107 ms .. 2.941 ms)

benchmarking 2^20/pointwise/random/fromList ... took 52.66 s, total 56 iterations
benchmarked 2^20/pointwise/random/fromList
time                 972.0 ms   (961.2 ms .. 986.7 ms)
mean                 918.6 ms   (886.5 ms .. 941.2 ms)
std dev              45.42 ms   (24.70 ms .. 63.02 ms)

benchmarking 2^20/pointwise/random/fromList_via_foldl_insert ... took 51.92 s, total 56 iterations
benchmarked 2^20/pointwise/random/fromList_via_foldl_insert
time                 980.2 ms   (948.5 ms .. 1.005 s)
mean                 897.7 ms   (859.0 ms .. 925.6 ms)
std dev              53.87 ms   (37.18 ms .. 68.48 ms)



```

jwaldmann:
30 million:
```
benchmarking 2^25/pointwise/contiguous/fromList ... took 245.4 s, total 56 iterations
benchmarked 2^25/pointwise/contiguous/fromList
time                 4.108 s    (3.742 s .. 4.314 s)
                     0.992 R²   (0.978 R² .. 0.999 R²)
mean                 3.903 s    (3.583 s .. 4.124 s)
std dev              462.4 ms   (245.2 ms .. 738.7 ms)
variance introduced by outliers: 38% (moderately inflated)

benchmarking 2^25/pointwise/contiguous/fromList_via_foldl_insert ... took 477.9 s, total 56 iterations
benchmarked 2^25/pointwise/contiguous/fromList_via_foldl_insert
time                 8.201 s    (8.084 s .. 8.314 s)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 8.336 s    (8.229 s .. 8.582 s)
std dev              250.5 ms   (104.9 ms .. 427.0 ms)

benchmarking 2^25/pointwise/interleaved/fromList ... took 540.3 s, total 56 iterations
benchmarked 2^25/pointwise/interleaved/fromList
time                 9.415 s    (9.263 s .. 9.603 s)
mean                 9.467 s    (9.364 s .. 9.731 s)
std dev              256.9 ms   (96.10 ms .. 435.1 ms)

benchmarking 2^25/pointwise/interleaved/fromList_via_foldl_insert ... took 501.8 s, total 56 iterations
benchmarked 2^25/pointwise/interleaved/fromList_via_foldl_insert
time                 8.928 s    (8.745 s .. 9.141 s)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 8.585 s    (8.200 s .. 8.792 s)
std dev              436.2 ms   (234.8 ms .. 657.1 ms)
```


```

**Comment by treeowl** [comment]
It really looks like the effects we're seeing are real. The new `fromList` is much better when the keys are contiguous, strictly increasing, or presumably clustered in general. It's about the same for random keys, and it's significantly but not horribly worse in the interleaved case. I'm convinced it's good enough to use. But I'd be a bit happier if we could reduce the regression in the interleaved case. Any ideas for doing so? `fromDistinctAscList` passes a bit of extra information around to avoid extra pattern matching on the subtree; can we do something similar? Might it be worth passing something extra back in `Inserted`?

**Comment by 3noch** [comment]
Perhaps it makes sense to expose the old version as `fromListInterleaved` or similar.

**Comment by treeowl** [comment]
@3noch I'm not generally a big fan of cluttering the API with one or two line functions, especially when anyone likely to need them will almost certainly know how to write them. `fromListNaive` is nothing more than `foldl' (flip (uncurry insert)) empty`, and `fromListNaiveWithKey` isn't much more.

**Comment by 3noch** [comment]
Perhaps it makes sense then to simply describe where `fromList` is less ideal performance wise and show the better performing version in the docs for that use case.

**Comment by treeowl** [comment]
@3noch, yes, definitely. I'm just still hoping there's a way to cut down on the regression....

**Comment by int-e** [comment]
Somewhat surprisingly it looks like we can indeed speed things up slightly by avoiding the repeated destruction of the same list, based around the following return type:
```haskell
data Inserted' a
   = InsertedNil  !(IntMap a)
   | InsertedCons !(IntMap a) !Key a ![(Key,a)]
```
See https://gist.github.com/int-e/36578cb04d0a187252c366b0b45ddcb6#file-intmapfl-hs-L102-L163 for code and https://gist.github.com/int-e/36578cb04d0a187252c366b0b45ddcb6#file-intmapfl-out for some benchmarks.

But for the most part I believe the slowdown for that adverserial alternating list case is the price one has to pay for checking, on each level of the tree, whether the new key fits or not, where the `naive` version simply starts at the root again each time.

treeowl:
That is somewhat surprising. I wonder what the Core looks like. I'd have thought the best bet for that sort of thing would've been an unboxed sum as a result, but maybe GHC is doing something clever.... Did you experiment with passing along a key in the map and a mask the way you do for `fromAscList`?

treeowl:
> But for the most part I believe the slowdown for that adverserial alternating list case is the price one has to pay for checking, on each level of the tree, whether the new key fits or not, where the naive version simply starts at the root again each time.

Yes, there is certainly some inherent arithmetic cost to that. All we can hope to do is minimize additional costs from allocation and pattern matching.

int-e:
> That is somewhat surprising. I wonder what the Core looks like. I'd have thought the best bet for that sort of thing would've been an unboxed sum as a result, but maybe GHC is doing something clever.... Did you experiment with passing along a key in the map and a mask the way you do for `fromAscList`?

No I didn't try that... it's not directly applicable. The point with `fromAscList` is that we know that keys will never be inserted into subtrees we've already built, so we can summarize the current state into a current prefix (expressed by the combination of a key and a mask). Here, we're always potentially inserting into an existing tree, so I believe it's unavoidable to inspect the actual tree we built and there we will find the prefix (expressed as prefix and mask) anyway.

treeowl:
I'm not so sure. In the bad case, we call `insertMany` on a tree that *doesn't* fit the next key. So if we keep enough information around to determine that without inspecting the tree, then *maybe* we can win something.

int-e:
> That is somewhat surprising. I wonder what the Core looks like. I'd have thought the best bet for that sort of thing would've been an unboxed sum as a result, but maybe GHC is doing something clever.... 

IIRC GHC's "enter" operation on the STG level actually takes several continuations for multi-constructor datatypes. So the `Inserted'` values should never be allocated. But I'm not sure, and I was not  at all convinced that this idea would pay off when I started to implement it.

treeowl:
insertMany' Nil k x kxs' = InsertedNil Nil

doesn't look so great, because it makes insertMany' lazy in the key and
list. Since we don't reach that anyway, it might as well force those things.

int-e:
> insertMany' Nil k x kxs' = InsertedNil Nil doesn't look so great, because it makes insertMany' lazy in the key and list. Since we don't reach that anyway, it might as well force those things.

I changed that, but it had no effect on the running time.

treeowl:
Probably got inlined away.... Did you check the CSE between link and
branchMask calls? If I'm not mistaken, link makes essentially the same
branchMask call you do, but with the arguments flipped.

int-e:
> Somewhat surprisingly it looks like we can indeed speed things up slightly by avoiding the repeated destruction of the same list, based around the following return type:
> 
> ```haskell
> data Inserted' a
>    = InsertedNil  !(IntMap a)
>    | InsertedCons !(IntMap a) !Key a ![(Key,a)]
> ```

Btw, we should also consider that the resulting speedup is miniscule and the code becomes more difficult to read as a result.

int-e:
> Probably got inlined away.... Did you check the CSE between link and branchMask calls? If I'm not mistaken, link makes essentially the same branchMask call you do, but with the arguments flipped.

No I had not checked that... and no, it did not do the CSE, regardless of the order of arguments to `branchMask`. I've now added a variant of `link` that takes a precomputed mask (to avoid inlining `link` manually), and the result does seem to improve performance a bit for short lists without hurting long ones. But there's hardly any signal here... mostly noise.

```haskell
linkWithMask :: Mask -> Prefix -> IntMap a -> Prefix -> IntMap a -> IntMap a
linkWithMask m p1 t1 p2 t2
  | zero p1 m = Bin p m t1 t2
  | otherwise = Bin p m t2 t1
  where
    p = mask p1 m
{-# INLINE linkWithMask #-}
```

```
[1,1]/fromAscList1a                      mean 19.91 ns  ( +- 14.99 ps  )
[1,1]/fromAscList1a'                     mean 19.38 ns  ( +- 199.5 ps  )
[10,1]/fromAscList1a                     mean 145.6 ns  ( +- 1.948 ns  )
[10,1]/fromAscList1a'                    mean 140.1 ns  ( +- 3.528 ns  )
[-10,51791]/fromAscList1a                mean 160.2 ns  ( +- 2.175 ns  )
[-10,51791]/fromAscList1a'               mean 149.8 ns  ( +- 3.113 ns  )
[100,1]/fromAscList1a                    mean 1.549 μs  ( +- 28.80 ns  )
[100,1]/fromAscList1a'                   mean 1.444 μs  ( +- 23.56 ns  )
[-100,51791]/fromAscList1a               mean 1.614 μs  ( +- 48.31 ns  )
[-100,51791]/fromAscList1a'              mean 1.454 μs  ( +- 46.67 ns  )
[1000,1]/fromAscList1a                   mean 16.24 μs  ( +- 183.1 ns  )
[1000,1]/fromAscList1a'                  mean 15.57 μs  ( +- 167.7 ns  )
[-1000,51791]/fromAscList1a              mean 17.01 μs  ( +- 492.8 ns  )
[-1000,51791]/fromAscList1a'             mean 16.07 μs  ( +- 466.1 ns  )
[10000,1]/fromAscList1a                  mean 232.7 μs  ( +- 2.194 μs  )
[10000,1]/fromAscList1a'                 mean 232.3 μs  ( +- 2.882 μs  )
[-10000,51791]/fromAscList1a             mean 250.8 μs  ( +- 4.353 μs  )
[-10000,51791]/fromAscList1a'            mean 246.1 μs  ( +- 5.896 μs  )
[100000,1]/fromAscList1a                 mean 9.101 ms  ( +- 227.0 μs  )
[100000,1]/fromAscList1a'                mean 9.097 ms  ( +- 240.5 μs  )
[-100000,51791]/fromAscList1a            mean 9.188 ms  ( +- 339.0 μs  )
[-100000,51791]/fromAscList1a'           mean 9.189 ms  ( +- 412.6 μs  )
[1000000,1]/fromAscList1a                mean 94.98 ms  ( +- 12.07 ms  )
[1000000,1]/fromAscList1a'               mean 96.65 ms  ( +- 12.22 ms  )
[-1000000,51791]/fromAscList1a           mean 97.83 ms  ( +- 10.69 ms  )
[-1000000,51791]/fromAscList1a'          mean 97.62 ms  ( +- 10.81 ms  )
```

int-e:
Oops, I believe we got #653 and #658 mixed up in the last 3 comments!

sjakobi:
I'm currently doing a bit of triage on the open PRs.

Could someone please summarize the current state of this PR, and possibly also sketch out what needs to be done until it can be merged?

int-e:
A lot going on here. I'll attempt a summary.

* the goal is to speed up `fromList` (borrowing ideas from `fromAscList`, which builds the whole map in a bottom-up fashion without deconstructing partial maps) and in particular try to make it no worse than repeated insertions (#288)
* we lack adequate benchmarks that test `fromList` under a variety of inputs with different characteristics
* I have some improvements to the PR code in a gist that could be incorporated, and made some effort to benchmark against various types of lists
* I believe this is stalled mainly because the code is still worse than before in some cases, and also because this development took place in parallel to improving `fromAscList` (#658) which was far more successful

sjakobi:
Thanks, @int-e! :)

Could the benchmarks be merged separately, so they are not lost or bit-rotted, and so they could potentially be used for other attempts at speeding up `fromList`?

Also, how about incorporating your code improvements into this PR? I guess the easiest way to do that would be if you'd make a PR against @treeowl's branch.

Of course it would also be nice to make progress on the original problem, but since that is currently stalled, let's try to ensure that the value created so far is saved.

int-e:
I'm not going to work on this. The benchmarks are in a very ad-hoc state. Ideally, the benchmarks would use *realistic* inputs for `fromList` and I don't really know what those are.

As for my improvements, after looking at the benchmark data again, there's either no speedup at all or it's so small that it looks like noise. While my changes were relatively minor tweaks and refactorings, they make the code more complicated too, so they're almost certainly not worthwhile. The real question is... is there a better approach? But that requires some serious meditation (and time).

sjakobi:
Thanks, @int-e! :)

Let's shelve this until we have better benchmarks then. (See #657)

meooow25:
This looks like a pretty good idea. I'm going to see if we can do this in a fusion-friendly way and how it holds up in benchmarks.

meooow25:
I think we can also apply this to Set/Map. The trouble there is that inserting makes the tree unbalanced. But we can back out just enough to restore balance, then continue on to the next element.

meooow25:
Closing in favor of #1137.

Issue #654, opened 2019-07-04
Can we improve fromDistinctAscList for IntSet and IntMap


treeowl:
The control flow of `fromDistinctAscList` for `IntSet` and `IntMap` is rather ... opaque. It operates using an explicit stack that *seems* to represent a sort of zipper for the tree being constructed. That stack is, of course, allocated on the heap, which doesn't seem ideal. The fact that the code is just a bit more complicated than I can really understand is unfortunate. My *guess* is that it's probably possible to shift that explicit stack onto GHC's native reduction stack, which seems likely to make the code less unreadable and more efficient.

treeowl:
@int-e I know you were involved in writing this code many years ago; I wonder if you might be interested in revisiting it.

treeowl:
The reason this came up: #653 seems to make `fromList` for these types almost as fast as `fromDistinctAscList`, which hints that the latter should be able to do better.

int-e:
I wouldn't expect a huge speedup... most of the speedup of `fromDistrinctAscList` comes from avoiding the repeated traversal and recreation of paths all the way from the root, and with the improved `fromList` you get that same effect. That said, I'm seeing a small speedup over the existing `fromDistinctAscList` with the following code:

```haskell
fromDistinctAscList1 :: [(Key,a)] -> IntMap a
fromDistinctAscList1 []              = Nil
fromDistinctAscList1 ((kx,vx) : zs0) = addAll kx (Tip kx vx) zs0
  where
    -- invariant for `addAll` and `addMany`: `kx` is a key in the map `tx`.
    addAll !kx !tx []
        = tx
    addAll !kx !tx ((ky,vy) : zs)
        | m <- branchMask kx ky
        , Inserted ty zs' <- addMany m ky (Tip ky vy) zs
        = addAll (mask kx m) (link ky ty kx tx) zs'

    -- addMany adds all elements that have the same prefix as `kx` w.r.t.
    -- the branch mask `m` to `tx`.
    addMany !m !kx tx []
        = Inserted tx []
    addMany !m !kx tx zs0@((ky,vy) : zs)
        | mask kx m /= mask ky m
        = Inserted tx zs0
        | Inserted ty zs' <- addMany (branchMask kx ky) ky (Tip ky vy) zs
        = addMany m kx (link ky ty kx tx) zs'

data Inserted a = Inserted !(IntMap a) ![(Key,a)]
```

[...]

PR #658, opened 2019-07-06
Improve fromAscList and friends.

int-e:
The relevant ticket is #654.

I have some supporting material, including more benchmarks, here:
https://gist.github.com/int-e/36578cb04d0a187252c366b0b45ddcb6

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -266,6 +266,7 @@ module Data.IntMap.Internal (
     , natFromInt
     , intFromNat
     , link
+    , linkWithMask
     , bin
     , binCheckLeft
     , binCheckRight
@@ -3111,8 +3112,8 @@ fromListWithKey f xs
 -- > fromAscList [(3,"b"), (5,"a"), (5,"b")] == fromList [(3, "b"), (5, "b")]
 
 fromAscList :: [(Key,a)] -> IntMap a
-fromAscList xs
-  = fromAscListWithKey (\_ x _ -> x) xs
+fromAscList = fromMonoListWithKey Nondistinct (\_ x _ -> x)
+{-# NOINLINE fromAscList #-}
 
 -- | /O(n)/. Build a map from a list of key\/value pairs where
 -- the keys are in ascending order, with a combining function on equal keys.
@@ -3121,8 +3122,8 @@ fromAscList xs
 -- > fromAscListWith (++) [(3,"b"), (5,"a"), (5,"b")] == fromList [(3, "b"), (5, "ba")]
 
 fromAscListWith :: (a -> a -> a) -> [(Key,a)] -> IntMap a
-fromAscListWith f xs
-  = fromAscListWithKey (\_ x y -> f x y) xs
+fromAscListWith f = fromMonoListWithKey Nondistinct (\_ x y -> f x y)
+{-# NOINLINE fromAscListWith #-}
 
 -- | /O(n)/. Build a map from a list of key\/value pairs where
 -- the keys are in ascending order, with a combining function on equal keys.
@@ -3132,14 +3133,8 @@ fromAscListWith f xs
 -- > fromAscListWithKey f [(3,"b"), (5,"a"), (5,"b")] == fromList [(3, "b"), (5, "5:b|a")]
 
 fromAscListWithKey :: (Key -> a -> a -> a) -> [(Key,a)] -> IntMap a
-fromAscListWithKey _ []         = Nil
-fromAscListWithKey f (x0 : xs0) = fromDistinctAscList (combineEq x0 xs0)
-  where
-    -- [combineEq f xs] combines equal elements with function [f] in an ordered list [xs]
-    combineEq z [] = [z]
-    combineEq z@(kz,zz) (x@(kx,xx):xs)
-      | kx==kz    = let yy = f kx xx zz in combineEq (kx,yy) xs
-      | otherwise = z:combineEq x xs
+fromAscListWithKey f = fromMonoListWithKey Nondistinct f
+{-# NOINLINE fromAscListWithKey #-}
 
 -- | /O(n)/. Build a map from a list of key\/value pairs where
 -- the keys are in ascending order and all distinct.
@@ -3147,35 +3142,71 @@ fromAscListWithKey f (x0 : xs0) = fromDistinctAscList (combineEq x0 xs0)
 --
 -- > fromDistinctAscList [(3,"b"), (5,"a")] == fromList [(3, "b"), (5, "a")]
 
-#if __GLASGOW_HASKELL__
-fromDistinctAscList :: forall a. [(Key,a)] -> IntMap a
-#else
-fromDistinctAscList ::            [(Key,a)] -> IntMap a
-#endif
-fromDistinctAscList []         = Nil
-fromDistinctAscList (z0 : zs0) = work z0 zs0 Nada
+fromDistinctAscList :: [(Key,a)] -> IntMap a
+fromDistinctAscList = fromMonoListWithKey Distinct (\_ x _ -> x)
+{-# NOINLINE fromDistinctAscList #-}
+
+-- | /O(n)/. Build a map from a list of key\/value pairs with monotonic keys
+-- and a combining function.
+--
+-- The precise conditions under which this function works are subtle:
+-- For any branch mask, keys with the same prefix w.r.t. the branch
+-- mask must occur consecutively in the list.
+
+fromMonoListWithKey :: Distinct -> (Key -> a -> a -> a) -> [(Key,a)] -> IntMap a
+fromMonoListWithKey distinct f = go
   where
-    work (kx,vx) []            stk = finish kx (Tip kx vx) stk
-    work (kx,vx) (z@(kz,_):zs) stk = reduce z zs (branchMask kx kz) kx (Tip kx vx) stk
+    go []              = Nil
+    go ((kx,vx) : zs1) = addAll' kx vx zs1
 
-#if __GLASGOW_HASKELL__
-    reduce :: (Key,a) -> [(Key,a)] -> Mask -> Prefix -> IntMap a -> Stack a -> IntMap a
-#endif
-    reduce z zs _ px tx Nada = work z zs (Push px tx Nada)
-    reduce z zs m px tx stk@(Push py ty stk') =
-        let mxy = branchMask px py
-            pxy = mask px mxy
-        in  if shorter m mxy
-            then reduce z zs m pxy (Bin pxy mxy ty tx) stk'
-            else work z zs (Push px tx stk)
+    -- `addAll'` collects all keys equal to `kx` into a single value,
+    -- and then proceeds with `addAll`.
+    addAll' !kx vx []
+        = Tip kx vx
+    addAll' !kx vx ((ky,vy) : zs)
+        | Nondistinct <- distinct, kx == ky
+        = let v = f kx vy vx in addAll' ky v zs
+        -- inlined: | otherwise = addAll kx (Tip kx vx) (ky : zs)
+        | m <- branchMask kx ky
+        , Inserted ty zs' <- addMany' m ky vy zs
+        = addAll kx (linkWithMask m ky ty {-kx-} (Tip kx vx)) zs'
 
-    finish _  t  Nada = t
-    finish px tx (Push py ty stk) = finish p (link py ty px tx) stk
-        where m = branchMask px py
-              p = mask px m
+    -- for `addAll` and `addMany`, kx is /a/ key inside the tree `tx`
+    -- `addAll` consumes the rest of the list, adding to the tree `tx`
+    addAll !kx !tx []
+        = tx
+    addAll !kx !tx ((ky,vy) : zs)
+        | m <- branchMask kx ky
+        , Inserted ty zs' <- addMany' m ky vy zs
+        = addAll kx (linkWithMask m ky ty {-kx-} tx) zs'
 
-data Stack a = Push {-# UNPACK #-} !Prefix !(IntMap a) !(Stack a) | Nada
+    -- `addMany'` is similar to `addAll'`, but proceeds with `addMany'`.
+    addMany' !m !kx vx []
+        = Inserted (Tip kx vx) []
+    addMany' !m !kx vx zs0@((ky,vy) : zs)
+        | Nondistinct <- distinct, kx == ky
+        = let v = f kx vy vx in addMany' m ky v zs
+        -- inlined: | otherwise = addMany m kx (Tip kx vx) (ky : zs)
+        | mask kx m /= mask ky m
+        = Inserted (Tip kx vx) zs0
+        | mxy <- branchMask kx ky
+        , Inserted ty zs' <- addMany' mxy ky vy zs
+        = addMany m kx (linkWithMask mxy ky ty {-kx-} (Tip kx vx)) zs'
 
+    -- `addAll` adds to `tx` all keys whose prefix w.r.t. `m` agrees with `kx`.
+    addMany !m !kx tx []
+        = Inserted tx []
+    addMany !m !kx tx zs0@((ky,vy) : zs)
+        | mask kx m /= mask ky m
+        = Inserted tx zs0
+        | mxy <- branchMask kx ky
+        , Inserted ty zs' <- addMany' mxy ky vy zs
+        = addMany m kx (linkWithMask mxy ky ty {-kx-} tx) zs'
+{-# INLINE fromMonoListWithKey #-}
+
+data Inserted a = Inserted !(IntMap a) ![(Key,a)]
+
+data Distinct = Distinct | Nondistinct
 
 {--------------------------------------------------------------------
   Eq
@@ -3297,13 +3328,17 @@ INSTANCE_TYPEABLE1(IntMap)
   Link
 --------------------------------------------------------------------}
 link :: Prefix -> IntMap a -> Prefix -> IntMap a -> IntMap a
-link p1 t1 p2 t2
+link p1 t1 p2 t2 = linkWithMask (branchMask p1 p2) p1 t1 {-p2-} t2
+{-# INLINE link #-}
+
+-- `linkWithMask` is useful when the `branchMask` has already been computed
+linkWithMask :: Mask -> Prefix -> IntMap a -> IntMap a -> IntMap a
+linkWithMask m p1 t1 {-p2-} t2
   | zero p1 m = Bin p m t1 t2
   | otherwise = Bin p m t2 t1
   where
-    m = branchMask p1 p2
     p = mask p1 m
-{-# INLINE link #-}
+{-# INLINE linkWithMask #-}
 
 {--------------------------------------------------------------------
   @bin@ assures that we never have empty trees within a tree.
--- a/containers/src/Data/IntMap/Strict/Internal.hs
+++ b/containers/src/Data/IntMap/Strict/Internal.hs
@@ -270,6 +270,7 @@ import Data.IntMap.Internal
   , binCheckLeft
   , binCheckRight
   , link
+  , linkWithMask
 
   , (\\)
   , (!)
@@ -1098,8 +1099,8 @@ fromListWithKey f xs
 -- > fromAscList [(3,"b"), (5,"a"), (5,"b")] == fromList [(3, "b"), (5, "b")]
 
 fromAscList :: [(Key,a)] -> IntMap a
-fromAscList xs
-  = fromAscListWithKey (\_ x _ -> x) xs
+fromAscList = fromMonoListWithKey Nondistinct (\_ x _ -> x)
+{-# NOINLINE fromAscList #-}
 
 -- | /O(n)/. Build a map from a list of key\/value pairs where
 -- the keys are in ascending order, with a combining function on equal keys.
@@ -1108,8 +1109,8 @@ fromAscList xs
 -- > fromAscListWith (++) [(3,"b"), (5,"a"), (5,"b")] == fromList [(3, "b"), (5, "ba")]
 
 fromAscListWith :: (a -> a -> a) -> [(Key,a)] -> IntMap a
-fromAscListWith f xs
-  = fromAscListWithKey (\_ x y -> f x y) xs
+fromAscListWith f = fromMonoListWithKey Nondistinct (\_ x y -> f x y)
+{-# NOINLINE fromAscListWith #-}
 
 -- | /O(n)/. Build a map from a list of key\/value pairs where
 -- the keys are in ascending order, with a combining function on equal keys.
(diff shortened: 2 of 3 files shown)


treeowl:
`fromMonoListWithKey` won't inline, so `fromAsclist` for lazy maps will leave stupid thunks in the structure. You can restructure `fromMonoListWithKey` to inline it with a known `f` and then mark the exported functions `NOINLINE`.

int-e:
> `fromMonoListWithKey` won't inline, so `fromAsclist` for lazy maps will leave stupid thunks in the structure. You can restructure `fromMonoListWithKey` to inline it with a known `f` and then mark the exported functions `NOINLINE`.

You're probably right; I'll verify tomorrow. I forgot about the pitfall that functions that are used only once are still inlined, so my benchmark attempt for this scenario was flawed.

int-e:
As expected, you were right about the extra thunks. Interestingly that was not a regression... the previous `fromAscList` also produced extra thunks.

I've sketched a test for such thunks in bade9f73d973f0c58bd55e562f7c9e682c174240... it's not yet properly split by module (Data.{,Int}Map{,.Strict}). I should probably split that out into a separate pull request, but I want to see if the current code passes the test for all relevant compilers first.

treeowl:
The approach I was suggesting was more like

```haskell
fromMonoListWithKey f = \ xs0 -> go xs0
  where
    go [] = []
    ....
```

I'm not sure how that will compare with your way in the non-inlined case (`fromAscListWithKey`), but it should be the same for the inlined case `fromAscList`.

int-e:
> The approach I was suggesting was more like [...]

That's what I tried at first, but it didn't cure the creation of thunks by `fromAscList`. So I added the `{-# INLINE #-}` and realized that the auxiliary `go` was no longer required.

int-e:
New benchmark result: for this particular benchmark inexplicably the code is 10% slower than before inlining `fromMonoListWithKey` now (inexplicably because the code path where `f` would matter is never taken.)
```
fromList                                 mean 242.6 μs  ( +- 994.5 ns  )
fromAscList                              mean 91.39 μs  ( +- 734.3 ns  )
fromDistinctAscList                      mean 91.29 μs  ( +- 473.5 ns  )
```
In contrast, running the code from https://gist.github.com/int-e/36578cb04d0a187252c366b0b45ddcb6#file-intmapfal2-hs the previous code was a tiny bit slower.

Note that none of these benchmarks actually uses duplicate keys. I cooked up a benchmark with duplicate keys using the following generator for keys:

```haskell
    keys = map ((sk*) . (`div` 3) . (*2)) (if sz < 0 then [2*sz `div` 3.. -sz `div` 3] else [0..sz])
```

In that case, avoiding the thunks actually results in a speedup, especially for larger map sizes:

```
[1,1]/fromAscList                        mean 13.65 ns  ( +- 62.14 ps  )
[1,1]/fromAscList1a                      mean 12.36 ns  ( +- 53.82 ps  )
[10,1]/fromAscList                       mean 122.2 ns  ( +- 1.424 ns  )
[10,1]/fromAscList1a                     mean 102.1 ns  ( +- 693.7 ps  )
[-10,51791]/fromAscList                  mean 139.7 ns  ( +- 1.247 ns  )
[-10,51791]/fromAscList1a                mean 117.8 ns  ( +- 763.6 ps  )
[100,1]/fromAscList                      mean 1.304 μs  ( +- 29.38 ns  )
[100,1]/fromAscList1a                    mean 1.232 μs  ( +- 67.38 ns  )
[-100,51791]/fromAscList                 mean 1.402 μs  ( +- 49.40 ns  )
[-100,51791]/fromAscList1a               mean 1.316 μs  ( +- 57.71 ns  )
[1000,1]/fromAscList                     mean 13.90 μs  ( +- 173.5 ns  )
[1000,1]/fromAscList1a                   mean 12.29 μs  ( +- 187.5 ns  )
[-1000,51791]/fromAscList                mean 15.12 μs  ( +- 545.2 ns  )
[-1000,51791]/fromAscList1a              mean 13.69 μs  ( +- 650.7 ns  )
[10000,1]/fromAscList                    mean 208.6 μs  ( +- 1.979 μs  )
[10000,1]/fromAscList1a                  mean 155.3 μs  ( +- 5.096 μs  )
[-10000,51791]/fromAscList               mean 226.5 μs  ( +- 4.236 μs  )
[-10000,51791]/fromAscList1a             mean 167.3 μs  ( +- 3.117 μs  )
[100000,1]/fromAscList                   mean 8.410 ms  ( +- 204.8 μs  )
[100000,1]/fromAscList1a                 mean 6.218 ms  ( +- 203.4 μs  )
[-100000,51791]/fromAscList              mean 8.877 ms  ( +- 322.3 μs  )
[-100000,51791]/fromAscList1a            mean 6.312 ms  ( +- 228.8 μs  )
```

where `fromAscList` is the version without inlined `fromMonoListWithKey` and I've checked that `fromAscList1a` matches the speed of the inlined one.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3132,50 +3132,72 @@ fromAscListWith f xs
...
+        | Inserted ty zs' <- addMany' (branchMask kx ky) ky vy zs
+        = addMany m kx (link ky ty kx (Tip kx vx)) zs'
+
+    -- `addAll` adds to `tx` all keys whose prefix w.r.t. `m` agrees with `kx`.
+    addMany !m !kx tx []
+        = Inserted tx []
+    addMany !m !kx tx zs0@((ky,vy) : zs)
+        | mask kx m /= mask ky m
+        = Inserted tx zs0
+        | Inserted ty zs' <- addMany' (branchMask kx ky) ky vy zs
```
treeowl: Same CSE point here.

treeowl:
I keep wondering if tracking *some* key in the map is really the best approach. In particular, I *gather* that `highestBitMask`, and therefore `branchMask`, are *relatively* expensive as bitwise operations go. What if we didn't need to use so many? Suppose we track the `Mask` of a subtree as well as a key within. If we take our new key, `xor` it with the one in the subtree, and mask off the result using the subtree's mask, we should be able to use

treeowl:
Sorry about that; accidentally hit "close and comment" when I wasn't even finished writing my non-closing comment....

treeowl:
But also I realize my last substantive comment was all wrong..... Hrm.... Never mind.

int-e:
I suppose the trick from https://github.com/haskell/containers/pull/653#issuecomment-509039744 is applicable here as well.

containers/src/Data/IntMap/Internal.hs
int-e: (see above)

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3132,50 +3132,72 @@ fromAscListWith f xs
...
+fromMonoListWithKey f ((kx,vx) : zs1) = addAll' kx vx zs1
+  where
+    -- `addAll'` collects all keys equal to `kx` into a single value,
+    -- and then proceeds with `addAll`.
+    addAll' !kx vx [] = Tip kx vx
+    addAll' !kx vx ((ky,vy) : zs)
+        | kx == ky
+        = let v = f kx vy vx in addAll' ky v zs
+        -- inlined: | otherwise = addAll kx (Tip kx vx) (ky : zs)
+        | m <- branchMask kx ky
```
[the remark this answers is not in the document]
int-e: Interestingly, changing this (and the other occurrences of `branchMask kx ky`) has no discernable effect:
```
[1,1]/fromAscList1a                      mean 19.62 ns  ( +- 30.11 ps  )
[1,1]/fromAscList1a'                     mean 20.45 ns  ( +- 65.89 ps  )
[10,1]/fromAscList1a                     mean 149.6 ns  ( +- 1.718 ns  )
[10,1]/fromAscList1a'                    mean 151.8 ns  ( +- 1.080 ns  )
[-10,51791]/fromAscList1a                mean 162.2 ns  ( +- 1.408 ns  )
[-10,51791]/fromAscList1a'               mean 166.8 ns  ( +- 1.161 ns  )
[100,1]/fromAscList1a                    mean 1.591 μs  ( +- 18.07 ns  )
[100,1]/fromAscList1a'                   mean 1.660 μs  ( +- 10.97 ns  )
[-100,51791]/fromAscList1a               mean 1.612 μs  ( +- 36.84 ns  )
[-100,51791]/fromAscList1a'              mean 1.735 μs  ( +- 27.26 ns  )
[1000,1]/fromAscList1a                   mean 16.77 μs  ( +- 189.0 ns  )
[1000,1]/fromAscList1a'                  mean 16.92 μs  ( +- 218.1 ns  )
[-1000,51791]/fromAscList1a              mean 17.97 μs  ( +- 388.3 ns  )
[-1000,51791]/fromAscList1a'             mean 18.31 μs  ( +- 701.3 ns  )
[10000,1]/fromAscList1a                  mean 240.1 μs  ( +- 1.948 μs  )
[10000,1]/fromAscList1a'                 mean 239.6 μs  ( +- 1.872 μs  )
[-10000,51791]/fromAscList1a             mean 263.3 μs  ( +- 5.600 μs  )
[-10000,51791]/fromAscList1a'            mean 254.0 μs  ( +- 5.248 μs  )
[100000,1]/fromAscList1a                 mean 9.113 ms  ( +- 263.2 μs  )
[100000,1]/fromAscList1a'                mean 9.182 ms  ( +- 244.9 μs  )
[-100000,51791]/fromAscList1a            mean 9.310 ms  ( +- 389.0 μs  )
[-100000,51791]/fromAscList1a'           mean 9.372 ms  ( +- 428.5 μs  )
```
I wonder whether ghc knows that `xor` is commutative. It would not surprise me much.

int-e:
> I suppose the trick from [#653 (comment)](https://github.com/haskell/containers/pull/653#issuecomment-509039744) is applicable here as well.

While applicable, it doesn't seem to pay off here. Selected benchmarks... other sizes show a small slowdown as well: (the new versions are `fromAscList2` and `fromAscList2a`)

```
[10000,1]/fromAscList1a                  mean 236.0 μs  ( +- 1.631 μs  )
[10000,1]/fromAscList2                   mean 268.3 μs  ( +- 22.73 μs  )
[10000,1]/fromAscList2a                  mean 266.5 μs  ( +- 21.94 μs  )

[-10000,51791]/fromAscList1a             mean 255.4 μs  ( +- 5.849 μs  )
[-10000,51791]/fromAscList2              mean 309.0 μs  ( +- 22.05 μs  )
[-10000,51791]/fromAscList2a             mean 293.4 μs  ( +- 21.61 μs  )
```

The code is not pretty either:

```haskell
fromAscList2a :: [(Key,a)] -> IntMap a
fromAscList2a []              = Nil
fromAscList2a ((kx,vx) : zs1) = addAll' kx vx zs1
  where
    -- `addAll'` collects all keys equal to `kx` into a single value,
    -- and then proceeds with `addAll`.
    addAll' !kx vx [] = Tip kx vx
    addAll' !kx vx ((ky,vy) : zs)
        | kx == ky
        = addAll' ky vy zs
        | m <- branchMask ky kx
        = case addMany' m ky vy zs of
            InsertedNil  ty           -> link ky ty kx (Tip kx vx)
            InsertedCons ty kz vz zs' -> addAll kx (link ky ty kx (Tip kx vx)) kz vz zs'

    -- for `addAll` and `addMany`, kx is /a/ key inside the tree `tx`
    -- `addAll` consumes the rest of the list, adding to the tree `tx`
    addAll !kx !tx !ky vy zs
        | m <- branchMask ky kx
        = case addMany' m ky vy zs of
            InsertedNil  ty           -> link ky ty kx tx
            InsertedCons ty kz vz zs' -> addAll kx (link ky ty kx tx) kz vz zs'

    -- `addMany'` is similar to `addAll'`, but proceeds with `addMany'`.
    addMany' !m !kx vx [] = InsertedNil (Tip kx vx)
    addMany' !m !kx vx ((ky,vy) : zs)
        | kx == ky
        = addMany' m ky vy zs
        | mask kx m /= mask ky m
        = InsertedCons (Tip kx vx) ky vy zs
        | otherwise
        = case addMany' (branchMask ky kx) ky vy zs of
            InsertedNil  ty           -> InsertedNil (link ky ty kx (Tip kx vx))
            InsertedCons ty kz vz zs' -> addMany m kx (link ky ty kx (Tip kx vx)) kz vz zs'

    -- `addAll` adds to `tx` all keys whose prefix w.r.t. `m` agrees with `kx`.
    addMany !m !kx tx !ky vy zs
        | mask kx m /= mask ky m
        = InsertedCons tx ky vy zs
        | otherwise
        = case addMany' (branchMask ky kx) ky vy zs of
            InsertedNil  ty           -> InsertedNil (link ky ty kx tx)
            InsertedCons ty kz vz zs' -> addMany m kx (link ky ty kx tx) kz vz zs'

data Inserted' a
   = InsertedNil  !(IntMap a)
   | InsertedCons !(IntMap a) !Key a ![(Key,a)]
```

treeowl:
Even if there's no reliably detectable advantage to the CSEed version, it should at least produce slightly smaller code, which is better. I'm a bit mystified about why the fancy `Inserted` representation would help there and not here.

containers/src/Data/IntMap/Internal.hs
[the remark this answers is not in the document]
treeowl: Which one is  `a` and which `a'`? Based on these numbers, `a` looks a drop better, no?

int-e:
Last action for tonight... I've pushed the CSE changes. (I'm making separate patches for each data structure because in the end I want to squash everything into the first three commits that I pushed.)

I'm also mystified by what exactly the benefits and drawbacks of the `Inserted'` trick are.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3297,13 +3322,17 @@ INSTANCE_TYPEABLE1(IntMap)
...
-link p1 t1 p2 t2
+link p1 t1 p2 t2 = linkWithMask (branchMask p1 p2) p1 t1 p2 t2
+{-# INLINE link #-}
+
+-- `linkWithMask` is useful when the `branchMask` has already been computed
+linkWithMask :: Mask -> Prefix -> IntMap a -> Prefix -> IntMap a -> IntMap a
+linkWithMask m p1 t1 p2 t2
   | zero p1 m = Bin p m t1 t2
   | otherwise = Bin p m t2 t1
   where
```
treeowl: It won't matter, thanks to inlining, but `linkWithMask` doesn't use `p2`, so it shouldn't take it.

containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -1134,24 +1129,64 @@ fromAscListWithKey f (x0 : xs0) = fromDistinctAscList (combineEq x0 xs0)
...
+    addAll !kx !tx []
+        = tx
+    addAll !kx !tx ((ky,!vy) : zs)
+        | m <- branchMask kx ky
+        , Inserted ty zs' <- addMany' m ky vy zs
+        = addAll kx (linkWithMask m ky ty kx tx) zs'
+
+    -- `addMany'` is similar to `addAll'`, but proceeds with `addMany'`.
+    addMany' !m !kx vx [] = Inserted (Tip kx vx) []
+    addMany' !m !kx vx zs0@((ky,!vy) : zs)
```
treeowl: The bang on `vy` doesn't look right. If I remember correctly, we're not supposed to force values for duplicate keys; we force the result of combining them with the old values instead.

containers/src/Data/IntMap/Internal.hs
int-e: FWIW, my motivation was to make it easy to switch between `lift` and `liftWithMask`. But of course we may later end up in a situation where `p2` is not readily available... Choices, choices...

containers/src/Data/IntMap/Strict/Internal.hs
int-e: Good point. I was working under the assumption that all values are supposed to be forced, but indeed we shouldn't make the function more strict than it currently is without a very good reason.

For your amusement (the results may look a bit surprising, but given that `fromList{,With}` works by inserting values into the map as it goes, the bottoms are unavoidable:
```
(using containers-0.6.2.1)
Prelude Data.IntMap.Strict> fromList [(1,undefined),(1,2)] `seq` ()
*** Exception: Prelude.undefined
Prelude Data.IntMap.Strict> fromAscList [(1,undefined),(1,2)] `seq` ()
()
Prelude Data.IntMap.Strict> fromListWith (\_ _ -> 2) [(1,undefined),(1,1)] `seq` ()
*** Exception: Prelude.undefined
Prelude Data.IntMap.Strict> fromListWith (\_ _ -> 2) [(1,1),(1,undefined)] `seq` ()
()
Prelude Data.IntMap.Strict> fromAscListWith (\_ _ -> 2) [(1,undefined),(1,1)] `seq` ()
()
Prelude Data.IntMap.Strict> fromAscListWith (\_ _ -> 2) [(1,1),(1,undefined)] `seq` ()
()
```

treeowl:
Yeah, the semantics are weird all right. But I think that needs to be a
separate discussion.

containers/src/Data/IntMap/Internal.hs
int-e: done

containers/src/Data/IntMap/Strict/Internal.hs
int-e: done

containers/src/Data/IntSet/Internal.hs
```diff
@@ -1060,41 +1060,69 @@ fromList xs
...
 -- /The precondition (input list is ascending) is not checked./
 fromAscList :: [Key] -> IntSet
-fromAscList [] = Nil
-fromAscList (x0 : xs0) = fromDistinctAscList (combineEq x0 xs0)
-  where
-    combineEq x' [] = [x']
-    combineEq x' (x:xs)
-      | x==x'     = combineEq x' xs
-      | otherwise = x' : combineEq x xs
+fromAscList = fromMonoList
```
int-e: Should this be marked as `NOINLINE`?

This question also applies to `fromDistinctAscList` below, and to the `from*AscList*` functions in `Data.IntMap` and `Data.IntMap.Strict`.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3132,50 +3133,74 @@ fromAscListWith f xs
...
 
+fromMonoListWithKey :: (Key -> a -> a -> a) -> [(Key,a)] -> IntMap a
+fromMonoListWithKey f []              = Nil
+fromMonoListWithKey f ((kx,vx) : zs1) = addAll' kx vx zs1
+  where
+    -- `addAll'` collects all keys equal to `kx` into a single value,
+    -- and then proceeds with `addAll`.
+    addAll' !kx vx [] = Tip kx vx
+    addAll' !kx vx ((ky,vy) : zs)
+        | kx == ky
```
int-e: We could pass an extra `distinct :: Bool` to `fromMonoListWithKey` (set to true for `fromDisctinctAscList`) and replace this guard by `not distinct && kx == ky`. That way it should be optimized away for `fromDistinctAscList` without any source code duplication.

This also applies to `fromMonoListWithKey` in `Data.IntMap.Strict`.

containers/src/Data/IntSet/Internal.hs
treeowl: Probably.

containers/src/Data/IntMap/Internal.hs
treeowl: Do the benchmarks say that helps?

int-e: Well... it hardly makes a difference on my computer. There may be the tiniest of speedups here. At least it doesn't hurt...

```
[1,1]/fromAscList                        mean 18.61 ns  ( +- 21.01 ps  )
[1,1]/fromDistinctAscList                mean 17.64 ns  ( +- 33.79 ps  )
[10,1]/fromAscList                       mean 135.2 ns  ( +- 2.796 ns  )
[10,1]/fromDistinctAscList               mean 136.9 ns  ( +- 1.888 ns  )
[-10,51791]/fromAscList                  mean 145.9 ns  ( +- 1.740 ns  )
[-10,51791]/fromDistinctAscList          mean 145.6 ns  ( +- 910.6 ps  )
[100,1]/fromAscList                      mean 1.507 μs  ( +- 29.49 ns  )
[100,1]/fromDistinctAscList              mean 1.499 μs  ( +- 20.96 ns  )
[-100,51791]/fromAscList                 mean 1.437 μs  ( +- 38.05 ns  )
[-100,51791]/fromDistinctAscList         mean 1.685 μs  ( +- 36.02 ns  )
[1000,1]/fromAscList                     mean 15.86 μs  ( +- 176.7 ns  )
[1000,1]/fromDistinctAscList             mean 15.95 μs  ( +- 140.2 ns  )
[-1000,51791]/fromAscList                mean 17.69 μs  ( +- 415.3 ns  )
[-1000,51791]/fromDistinctAscList        mean 17.51 μs  ( +- 355.6 ns  )
[10000,1]/fromAscList                    mean 233.1 μs  ( +- 1.309 μs  )
[10000,1]/fromDistinctAscList            mean 230.1 μs  ( +- 2.108 μs  )
[-10000,51791]/fromAscList               mean 265.6 μs  ( +- 3.846 μs  )
[-10000,51791]/fromDistinctAscList       mean 259.5 μs  ( +- 3.787 μs  )
[100000,1]/fromAscList                   mean 9.184 ms  ( +- 269.9 μs  )
[100000,1]/fromDistinctAscList           mean 9.270 ms  ( +- 311.3 μs  )
[-100000,51791]/fromAscList              mean 9.560 ms  ( +- 417.8 μs  )
[-100000,51791]/fromDistinctAscList      mean 9.596 ms  ( +- 431.5 μs  )
[1000000,1]/fromAscList                  mean 96.69 ms  ( +- 12.43 ms  )
[1000000,1]/fromDistinctAscList          mean 98.32 ms  ( +- 12.72 ms  )
[-1000000,51791]/fromAscList             mean 100.6 ms  ( +- 11.41 ms  )
[-1000000,51791]/fromDistinctAscList     mean 99.89 ms  ( +- 10.85 ms  )
```

int-e: See 69d843d26aae874e6ab43e612f3f49bd1f6771b1, 72e0242f3f82bc5b5dd0b1a74afb2176098dad48. I'm inclined to do this because it should never hurt, but I'm ready to be overruled.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3154,12 +3154,15 @@ fromDistinctAscList = fromAscList
 -- mask must occur consecutively in the list.
 
 fromMonoListWithKey :: (Key -> a -> a -> a) -> [(Key,a)] -> IntMap a
-fromMonoListWithKey f []              = Nil
-fromMonoListWithKey f ((kx,vx) : zs1) = addAll' kx vx zs1
+fromMonoListWithKey f = go
```
int-e: Note: I went back on an earlier decision and introduced a helper here because I'm slightly worried that otherwise, all the other helpers, which constitute the bulk of the `fromMonoListWithKey` implementation, may be duplicated.

treeowl: Do you consider this ready to merge?

treeowl:
Do you consider this ready to merge? If so, could you squash whatever commits you want?

containers/src/Data/IntMap/Internal.hs
int-e: I've included this in my latest push.

int-e:
> Do you consider this ready to merge? If so, could you squash whatever commits you want?

I think it's ready to go.

treeowl:
One more request: could you use a custom 2-value type instead of `Bool` to indicate the distinctness requirement?

int-e:
> One more request: could you use a custom 2-value type instead of `Bool` to indicate the distinctness requirement?

Done. I struggled with the naming for a bit; in the end I went with `data Distinct = Distinct | Nondistinct`, but I don't particularily like it.

treeowl:
It's not the best, but we can fix it later. At least it says something!

treeowl:
Many thanks!

PR #662, opened 2019-07-08
add target for *-strictness tests

int-e:
(this enables `make intset-strictness` and so forth)


treeowl:
Sorry that took so long to merge.

PR #663, opened 2019-07-08
add fromList tests

int-e:
This is adding some test cases that were previously part of #658.

Note that one test is currently failing because `Data.IntMap.fromAscList` is a bit lazier than expected.

### DIFF
--- a/containers-tests/tests/Utils/IsUnit.hs
+++ b/containers-tests/tests/Utils/IsUnit.hs
@@ -1,0 +1,42 @@
+{-# LANGUAGE CPP #-}
+#ifdef __GLASGOW_HASKELL__
+{-# LANGUAGE MagicHash #-}
+#endif
+
+module Utils.IsUnit (isUnit, isUnitSupported) where
+
+#ifdef __GLASGOW_HASKELL__
+import GHC.Exts
+#endif
+
+-- | Check whether the argument is a fully evaluated unit `()`.
+--
+-- Always returns `False` is `isUnitSupported` returns `False`.
+--
+-- Uses `reallyUnsafePtrEquality#`.
+isUnit :: () -> Bool
+
+-- | Checks whether `isUnit` is supported by the Haskell implementation.
+--
+-- Currently returns `True` for ghc and `False` for all other implementations.
+isUnitSupported :: Bool
+
+#ifdef __GLASGOW_HASKELL__
+
+-- simplified from  Utils.Containers.Internal.PtrEquality
+ptrEq :: a -> a -> Bool
+ptrEq x y = case reallyUnsafePtrEquality# x y of
+    0# -> False
+    _  -> True
+
+isUnit = ptrEq ()
+
+isUnitSupported = True
+
+#else /* !__GLASGOW_HASKELL__ */
+
+isUnit = False
+
+isUnitSupported = False
+
+#endif
--- a/containers-tests/tests/intmap-strictness.hs
+++ b/containers-tests/tests/intmap-strictness.hs
@@ -3,15 +3,19 @@
 module Main (main) where
 
 import Test.ChasingBottoms.IsBottom
-import Test.Framework (Test, defaultMain, testGroup)
+import Test.Framework (Test, TestName, defaultMain, testGroup)
 import Test.Framework.Providers.QuickCheck2 (testProperty)
 import Test.QuickCheck (Arbitrary(arbitrary))
 import Test.QuickCheck.Function (Fun(..), apply)
+import Test.Framework.Providers.HUnit
+import Test.HUnit hiding (Test)
 
 import Data.IntMap.Strict (IntMap)
 import qualified Data.IntMap.Strict as M
 import qualified Data.IntMap as L
 import Data.Containers.ListUtils
+
+import Utils.IsUnit
 
 instance Arbitrary v => Arbitrary (IntMap v) where
     arbitrary = M.fromList `fmap` arbitrary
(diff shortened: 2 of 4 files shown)


containers-tests/tests/Utils/IsUnit.hs
```diff
@@ -0,0 +1,16 @@
...
+same :: a -> a -> Bool
+same x y = case reallyUnsafePtrEquality# x y of
+    0# -> False
+    _  -> True
+
+-- | Check whether the argument is a fully evaluated unit `()`.
+--
+-- Uses `reallyUnsafePtrEquality#`.
+isUnit :: () -> Bool
+isUnit = same ()
```
int-e: I'm wondering whether I'm reinventing a wheel here. This seems closely related to what ChasingBottoms is doing, but I didn't find anything like this there.

treeowl:
Do you consider the test or the code correct? Or neither?

int-e:
I believe the test is correct (it will pass once #658 is merged).

treeowl:
Can you rebase please?

int-e:
> Can you rebase please?

Done. (I don't know whether the force-push itself causes a notification.)

containers-tests/tests/Utils/IsUnit.hs
treeowl: This smells funny/fragile to me. I would generally feel more comfortable with an approach that uses `unsafePerformIO` to find out when/how many times things are evaluated.

treeowl: Oh, but of course we can't always do that .... Hrmmm. Could you at least rename `same` to match our internal pointer equality function?

int-e: Oh I wasn't aware of `ptrEq`, or I would have used that; will change.

I should probably make the tests conditional on __GLASGOW_HASKELL__ being defined... and  maybe make isUnit tristate? (Yes | No | Maybe; the last indicating that the result is unknown.)

treeowl:
The test suite is already pretty GHC-specific, so I wouldn't worry about
that. Tristate ... not sure. Depends what you're trying to check. Add more
comments?

containers-tests/tests/Utils/IsUnit.hs
int-e: > Oh I wasn't aware of `ptrEq`, or I would have used that; will change.

Ah it's not so easy because the module is hidden, and we probably want to keep the library parts of `containers-tests` and `containers` the same.

> I should probably make the tests conditional on `__GLASGOW_HASKELL__` being defined... and maybe make isUnit tristate? (Yes | No | Maybe; the last indicating that the result is unknown.)

I will go with `Maybe Bool`.

int-e: Given your later comment, I stuck with `Bool`. I did add some logic to deal with non-ghc implementations though.

sjakobi:
@int-e, @treeowl, what's the status here?

int-e:
Nothing new from my side... the branch rebased cleanly on current master, and the tests currently pass (for me).

containers-tests/tests/intmap-strictness.hs
```diff
@@ -76,6 +82,70 @@ pInsertLookupWithKeyValueStrict f k v m
                      not (isBottom $ M.insertLookupWithKey (const3 1) k bottom m)
     | otherwise    = isBottom $ M.insertLookupWithKey (apply3 f) k bottom m
 
+pFromAscListLazy :: [Int] -> Bool
+pFromAscListLazy ks = not . isBottom $ M.fromAscList elems
+  where
+    ks' = map head . group . sort $ ks
```
sjakobi, line 88: Why not just use `Data.Containers.ListUtils.nubInt`? And why do the elements have to be unique at all? We're not testing `fromDistinctAscList` here.

containers-tests/tests/intmap-strictness.hs
```diff
@@ -76,6 +82,70 @@ pInsertLookupWithKeyValueStrict f k v m
...
+    | otherwise = isBottom $ M.fromList elems
+  where
+    ks' = map head . group . sort $ ks
+    elems = [(k, v) | k <- ks', v <- [undefined, undefined, ()]]
+
+------------------------------------------------------------------------
+-- check for extra thunks
+
+tExtraThunksM :: Test
+tExtraThunksM = testGroup "IntMap.Strict - extra thunks" $
```
sjakobi, line 103: What are extra thunks? Maybe add a brief explanation and/or a reference to a definition or discussion.

containers-tests/tests/intmap-strictness.hs
```diff
@@ -76,6 +82,70 @@ pInsertLookupWithKeyValueStrict f k v m
...
+    check :: TestName -> Bool -> IntMap () -> Test
+    check n e m = testCase n $ case M.lookup 42 m of
+        Just v -> assertBool msg (e == isUnit v)
+        _      -> assertString "key not found"
+      where
+        msg | e         = "too lazy -- expected fully evaluated ()"
+            | otherwise = "too strict -- expected a thunk"
+
+tExtraThunksL :: Test
+tExtraThunksL = testGroup "IntMap - extra thunks" $
```
sjakobi, line 126:
```suggestion
tExtraThunksL = testGroup "IntMap.Lazy - extra thunks" $
```

containers-tests/tests/intmap-strictness.hs
```diff
@@ -76,6 +82,70 @@ pInsertLookupWithKeyValueStrict f k v m
...
+    ]
+  where
+    m0 = M.singleton 42 ()
+    check :: TestName -> Bool -> IntMap () -> Test
+    check n e m = testCase n $ case M.lookup 42 m of
+        Just v -> assertBool msg (e == isUnit v)
+        _      -> assertString "key not found"
+      where
+        msg | e         = "too lazy -- expected fully evaluated ()"
+            | otherwise = "too strict -- expected a thunk"
```
sjakobi, line 123: Looks like we never use `check ... False` here, so this could possibly be simplified?!

containers-tests/tests/intmap-strictness.hs
int-e: Duplicate keys would actually invalidate this check; `fromAscList [(0,undefined),(0,()),(0,undefined),(0,())]` is bottom. (Not saying that this is how it should be... but that's how it is and has been for a long time.) 

I was not actively aware of `nubInt`, somehow.

Anyway I've addressed this and the other comments as well; I found a few minor discrepancies on top (like `Map` tests having `IntMap` in their name...)

containers-tests/tests/intmap-strictness.hs
```diff
@@ -76,6 +82,76 @@ pInsertLookupWithKeyValueStrict f k v m
                      not (isBottom $ M.insertLookupWithKey (const3 1) k bottom m)
     | otherwise    = isBottom $ M.insertLookupWithKey (apply3 f) k bottom m
 
+------------------------------------------------------------------------
+-- test a corner case of fromAscList: if the list contains duplicate
+-- keys, then (only) the first of the given values is not evaluated.
```
sjakobi, line 87: That seems like a surprising property to me. I think https://github.com/haskell/containers/issues/473 tracks what we could about it. Could you add a reference to that issue?

int-e: done.

sjakobi:
> @treeowl, please review!

Ping @treeowl! :)

PR #664, opened 2019-07-14
Add Data.IntSet.mapMonotonic.

Javran:
This patch fills in the missing piece for Data.IntSet, since Data.{Map,IntMap} and Data.Set all have this function. (for Map variant it's `mapKeysMonotonic`)

### DIFF
--- a/containers-tests/tests/intset-properties.hs
+++ b/containers-tests/tests/intset-properties.hs
@@ -337,6 +337,11 @@ prop_foldL' s = foldl' (flip (:)) [] s == List.foldl' (flip (:)) [] (toList s)
 prop_map :: IntSet -> Bool
 prop_map s = map id s == s
 
+prop_mapMonotonic :: (Key -> Key) -> IntSet -> Property
+prop_mapMonotonic f s =
+  (\x y -> x < y ==> f x < f y) .&&.
+  (mapMonotonic f s == map f s)
+
 prop_maxView :: IntSet -> Bool
 prop_maxView s = case maxView s of
     Nothing -> null s
--- a/containers-tests/tests/intset-properties.hs
+++ b/containers-tests/tests/intset-properties.hs
@@ -340,7 +340,7 @@ prop_map s = map id s == s
 prop_mapMonotonic :: (Key -> Key) -> IntSet -> Property
 prop_mapMonotonic f s =
   (\x y -> x < y ==> f x < f y) .&&.
-  (mapMonotonic f s == map f s)
+  mapMonotonic f s == map f s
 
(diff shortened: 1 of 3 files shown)


containers-tests/tests/intset-properties.hs
```diff
@@ -337,6 +337,10 @@ prop_foldL' s = foldl' (flip (:)) [] s == List.foldl' (flip (:)) [] (toList s)
 prop_map :: IntSet -> Bool
 prop_map s = map id s == s
 
+prop_mapMonotonic :: IntSet -> Bool
+prop_mapMonotonic s = mapMonotonic f s == map f s
+  where f = (* 7)
+
```
treeowl: This test is insufficient. Please generate arbitrary monotonic functions.

Javran: Thanks for the quick response. I've updated the patch as requested. =
However I noticed that Data.Set doesn't have sufficient test either: https://github.com/haskell/containers/blob/14c46116a8b757907f0b56dbb3aee3712ea1d63f/containers-tests/tests/set-properties.hs#L539-L540. Would you mind if I add some similar tests to that in this PR as well?

Javran:
Sorry about that forced push - for a second I forgot I can git squash after the review is done.

Javran:
Friendly ping. Any suggestions?

containers-tests/tests/intset-properties.hs
```diff
@@ -337,6 +337,11 @@ prop_foldL' s = foldl' (flip (:)) [] s == List.foldl' (flip (:)) [] (toList s)
 prop_map :: IntSet -> Bool
 prop_map s = map id s == s
 
+prop_mapMonotonic :: (Key -> Key) -> IntSet -> Property
+prop_mapMonotonic f s =
+  (\x y -> x < y ==> f x < f y) .&&.
+  mapMonotonic f s === map f s
```
treeowl: This is going to be too inefficient; too many test cases will be rejected. Also, properties should not take actual functions as arguments. QuickCheck offers a `Fun` type with growing and shrinking. I will have to think a bit about how to write this property optimally. For now, it might be okay to use the test you had before as long as you add a comment in the `mapMonotonic` implementation indicating that the test is insufficient to support any fancier implementation.

Javran: I guess I can still do better than I originally did - I break down properties into 2: one to make sure it works correctly with `id`, and another just construct `f x = a*x + b` with positive `a`. So this at least get arbitrary linear functions covered.

Javran:
Any updates on this?

treeowl:
Done! Thanks.

PR #667, opened 2019-07-18
Improve foldl and foldr for IntSet

### LINKED ISSUE #666: IntSet foldl and foldr smell funny
The implementations of `foldl'Bits` and `foldr'Bits` look efficient. The implementations of `foldlBits` and `foldrBits`, on the other hand, don't look efficient for their purposes. I'd expect `foldlBits` to look more like `foldr'Bits`, and I'd expect `foldrBits` to look more like `foldl'Bits`.

treeowl:
`foldrBits` was previously defined just like `foldr'Bits`, and
`foldlBits` was defined just like `foldl'Bits`. That is, each of them
built up an accumulator, lazily, across a whole leaf before producing
anything. Swap that around so the folds can process values promptly.

Fixes #666

### DIFF
--- a/containers-tests/benchmarks/IntSet.hs
+++ b/containers-tests/benchmarks/IntSet.hs
@@ -15,6 +15,8 @@ main = do
     evaluate $ rnf [s, s_even, s_odd]
     defaultMain
         [ bench "member" $ whnf (member elems) s
+        , bench "foldr" $ whnf (sumUntilL (2^14 - 5)) fish
+        , bench "foldl" $ whnf (sumUntilR 5) fish
         , bench "insert" $ whnf (ins elems) S.empty
         , bench "map" $ whnf (S.map (+ 1)) s
         , bench "filter" $ whnf (S.filter ((== 0) . (`mod` 2))) s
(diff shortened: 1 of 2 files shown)


treeowl:
I've done a bit of benchmarking. My results aren't very stable, but it seems, generally, that `foldr` is pretty similar (perhaps because of the cost of reversing the bits) while `foldl` is improved. Perhaps someone better at benchmarking and/or micro-optimization should take a look at it.

treeowl:
Er.... no, that's not it. The reversing happens with `foldl`.... I have no explanation.

treeowl:
@oisdk, this looks like the sort of thing you might know how to think about and test properly. Any ideas?

meooow25:
@treeowl do you want to continue with this PR? The changes to `foldrBits` and `foldlBits` seem ok to me.  
Once rebased we could add some benchmarks similar to what we recently added for `Data.Tree`'s `foldr` and measure if there's an improvement.

treeowl:
I'm currently recovering from COVID, so can't really do much this week.

meooow25:
Hope you get well soon! I didn't mean to imply any sort of urgency, this can surely wait.

treeowl:
I feel bad for making you wait so long on multiple things.

meooow25:
I have had some free time, don't worry about it 😅

meooow25:
Fixed by #1079.

PR #672, opened 2019-07-23
Add 'compose' for maps

alexfmpe:
Implement #647 based on community input in mailing list.

I see most functions don't have an entry in the benchmarks. Should I add anything with this PR?

cc @Ericson2314


containers-tests/tests/intmap-properties.hs
```diff
@@ -884,6 +886,9 @@ prop_intersectionWithKeyModel xs ys
 prop_disjoint :: UMap -> UMap -> Property
 prop_disjoint m1 m2 = disjoint m1 m2 === null (intersection m1 m2)
 
+prop_compose :: IMap -> IMap -> Int -> Property
+prop_compose bc ab k = (compose bc ab !? k) === (flip lookup ab >=> flip lookup bc) k
```
treeowl: Why not `((bc !?) <=< (ab !?)) k` to match the composition order and style? No big deal either way, of course.

containers-tests/tests/map-properties.hs
```diff
@@ -1074,6 +1076,9 @@ prop_intersectionWithKeyModel xs ys
 prop_disjoint :: UMap -> UMap -> Property
 prop_disjoint m1 m2 = disjoint m1 m2 === null (intersection m1 m2)
 
+prop_compose :: IMap -> IMap -> Int -> Property
+prop_compose bc ab k = (compose bc ab !? k) === (flip lookup ab >=> flip lookup bc) k
```
treeowl: Same question here.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -761,6 +764,25 @@ disjoint t1@(Bin p1 m1 l1 r1) t2@(Bin p2 m2 l2 r2)
...
+-- > compose (fromList [('a', "A"), ('b', "B")]) (fromList [(1,'a'),(2,'b'),(3,'z')]) = fromList [(1,"A"),(2,"B")]
+--
+-- @
+-- ('compose' bc ab '!?') = (bc '!?') <=< (ab '!?')
+-- @
+--
+-- @since UNRELEASED
+compose :: IntMap c -> IntMap Int -> IntMap c
+compose bc ab = if null bc
+                then empty
```
treeowl: This is surprisingly lazy in the second argument. Is that special case a good special case? I'm not so convinced yet. Please make an argument.

containers/src/Data/Map/Internal.hs
```diff
@@ -2085,6 +2088,25 @@ disjoint (Bin _ k _ l r) t
   where
     (lt,found,gt) = splitMember k t
 
+{--------------------------------------------------------------------
+  Compose
+--------------------------------------------------------------------}
+-- | /O(|ab|*log(|bc|))/. Relate the keys of one map to the values of
+-- the other, by using the values of the former as keys for lookups
+-- on the later.
```
treeowl:
```suggestion
-- in the latter.
```

containers/src/Data/IntMap/Internal.hs
```diff
@@ -761,6 +764,25 @@ disjoint t1@(Bin p1 m1 l1 r1) t2@(Bin p2 m2 l2 r2)
               | zero p1 m2       = disjoint t1 l2
               | otherwise        = disjoint t1 r2
 
+{--------------------------------------------------------------------
+  Compose
+--------------------------------------------------------------------}
+-- | /O(|ab|*log(|bc|))/. Relate the keys of one map to the values of
```
treeowl: That bound looks wrong. Poke around the rest of the module to see what the bounds look like.

containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -717,6 +720,25 @@ intersectionWithKey :: (Key -> a -> b -> c) -> IntMap a -> IntMap b -> IntMap c
...
+-- @
+-- ('compose' bc ab '!?') = (bc '!?') <=< (ab '!?')
+-- @
+--
+-- @since UNRELEASED
+compose :: IntMap c -> IntMap Int -> IntMap c
+compose bc ab = if null bc
+                then empty
+                else mapMaybe (bc !?) ab
+
```
treeowl: Same correction and comments apply here.

containers/src/Data/Map/Internal.hs
```diff
@@ -2085,6 +2088,25 @@ disjoint (Bin _ k _ l r) t
   where
     (lt,found,gt) = splitMember k t
 
+{--------------------------------------------------------------------
+  Compose
+--------------------------------------------------------------------}
+-- | /O(|ab|*log(|bc|))/. Relate the keys of one map to the values of
```
treeowl: This bound is a bit funny when |bc| ≤ 1. Is it worth fixing? Not sure.

containers/src/Data/IntMap/Internal.hs
alexfmpe: Do you mean that it should use `n` & `m` like the other functions? I strayed from that because other bounds seem to be commutative in their arguments, so they don't need to say which argument map is `n` and which is `m`, but `compose` is asymmetric

containers/src/Data/Map/Internal.hs
alexfmpe: Wouldn't that also apply to the existing `O(n log n)` bounds?
Big O notation is asymptotic, so in a sense there's nothing to fix.
Is there something specific about this bound you intend to make clearer? e.g. the `if null bc` case

containers/src/Data/IntMap/Internal.hs
alexfmpe: I don't think I yet sufficiently grasp the laziness expectations in these modules to answer properly. Here's what I managed to gather after reading your question:

The strict variants say
> If all values stored in all maps in the arguments are in WHNF, then all values stored in all maps in the results will be in WHNF once those maps are evaluated.

Does the same apply to keys for lazy maps given we have
>  A Map is strict in its keys but lazy in its values.

, and in that case `ab` should have its keys (and values, if strict) always evaluated if the result map is evaluated? Thus making optimizations like this one here somewhat break the interface?

containers/src/Data/IntMap/Internal.hs
treeowl: No, I mean that `IntMap` performance doesn't tend to involve logarithms of the size; instead, it usually relies on the word size. This is because the depth of the tree is not bounded by the logarithm of the total size. Check the bounds for `(!?)`, which you use here.

containers/src/Data/IntMap/Internal.hs
alexfmpe: Oh I see, the second argument is potentially lazier than the first

containers/src/Data/IntMap/Internal.hs
alexfmpe: Ah yes, makes sense

alexfmpe:
After doing a few rounds of tweaks for all 4 implementations, given that they all kind of have the same implementation, I wonder if it would simplify maintenance to have a neutral/generic compose:
```haskell
compose empty' lookup' bc !ab
  | null bc = empty'
  | otherwise = mapMaybe (flip lookup' bc) ab
```
so that the actual implementations would be something like
```haskell
compose = Utils.compose empty lookup
```

alexfmpe:
@treeowl I think I addressed the remarks - is there anything else I should do for this PR?

treeowl:
I'll try to have another look tonight. I expect this to go in before the next release.

treeowl:
Would you please rebase? I will then merge. Sorry it's been a bit.

treeowl:
Thanks a lot!

PR #710, opened 2020-03-24
Fix deleteFindMin Haddock example

zliu41:
deleteFindMin was missing the map argument (`empty`).


m-renaud:
Thanks for the fix! Will push a new version of the docs tonight.

PR #714, opened 2020-04-14
Make Map an instance of Bifoldable

josephcsible:


josephcsible:
Post to libraries@haskell.org: https://mail.haskell.org/pipermail/libraries/2020-April/030363.html

If this is accepted, I'd like my name to be listed as "Joseph C. Sible" in the changelog.

containers-tests/tests/map-properties.hs
```diff
@@ -17,6 +17,10 @@ import Data.Maybe hiding (mapMaybe)
 import qualified Data.Maybe as Maybe (mapMaybe)
 import Data.Ord
 import Data.Function
+import qualified Data.Foldable as Foldable
+#if MIN_VERSION_base(4,10,0)
+import Data.Bifoldable
```
treeowl, line 22: Please import this qualified.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,12 +1371,46 @@ prop_splitModel n ys = length ys > 0 ==>
   in  toAscList l == sort [(k, v) | (k,v) <- xs, k < n] &&
       toAscList r == sort [(k, v) | (k,v) <- xs, k > n]
 
+prop_fold :: [(Int, Int)] -> Property
```
treeowl, line 1374: Let's have one property for each fold.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,12 +1371,46 @@ prop_splitModel n ys = length ys > 0 ==>
...
+prop_fold ys = length ys > 0 ==>
+  let xs = List.nubBy ((==) `on` fst) ys
+      m  = fromList xs
+  in  getSum (Foldable.fold (map Sum m)) == sum (List.map snd xs) &&
+#if MIN_VERSION_base(4,10,0)
+      -- This seems to fit the pattern of the other test cases, but isn't valid here:
+      --getSum (bifold (mapKeys (const mempty) (map Sum m))) == sum (List.map fst xs) &&
+      -- since it would map all of the keys to the same thing, thus removing all but one element.
+      getSum (bifold (mapKeys Sum (map (const mempty) m))) == sum (List.map fst xs) &&
+      bifold (mapKeys (:[]) (map (:[]) m)) == concatMap (\(k,v) -> [k,v]) (List.sort xs) &&
```
treeowl, line 1384: I don't understand what any of this crud is doing. I think we should either use the list monoid exclusively or, if that is too expensive, the sequence monoid. I don't care for the nubbing or the use of `fromList`.  Take a `Map` as an argument (which gives us lots of ways to balance arbitrarily) and use `toList`.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,12 +1371,46 @@ prop_splitModel n ys = length ys > 0 ==>
...
+#if MIN_VERSION_base(4,10,0)
+      getSum (bifoldMap (const mempty) Sum m) == sum (List.map snd xs) &&
+      getSum (bifoldMap Sum (const mempty) m) == sum (List.map fst xs) &&
+      bifoldMap (:[]) (:[]) m == concatMap (\(k,v) -> [k,v]) (List.sort xs) &&
+#endif
+      Foldable.foldMap (:[]) m == List.map snd (List.sort xs) &&
+      getSum (foldMapWithKey (\_ v -> Sum v) m) == sum (List.map snd xs) &&
+      getSum (foldMapWithKey (\k _ -> Sum k) m) == sum (List.map fst xs) &&
+      foldMapWithKey (\k v -> [k,v]) m == concatMap (\(k,v) -> [k,v]) (List.sort xs)
+
```
treeowl, line 1402: All the same crud applies here as above.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,12 +1371,46 @@ prop_splitModel n ys = length ys > 0 ==>
...
 prop_foldr :: Int -> [(Int, Int)] -> Property
 prop_foldr n ys = length ys > 0 ==>
   let xs = List.nubBy ((==) `on` fst) ys
       m  = fromList xs
   in  foldr (+) n m == List.foldr (+) n (List.map snd xs) &&
       foldr (:) [] m == List.map snd (List.sort xs) &&
+#if MIN_VERSION_base(4,10,0)
+      bifoldr (const id) (+) n m == List.foldr (+) n (List.map snd xs) &&
+      bifoldr (+) (const id) n m == List.foldr (+) n (List.map fst xs) &&
+      bifoldr (:) (:) [] m == concatMap (\(k,v) -> [k,v]) (List.sort xs) &&
```
treeowl, line 1412: And here too. Fix the other one while you're at it.

treeowl, line 1412: And here too. Fix the other one while you're at it.

containers-tests/tests/map-properties.hs
josephcsible: But `toList` is implemented in terms of `foldrWithKey`, so wouldn't doing that make the tests basically worthless, since `foldrWithKey` is one of the things we're supposed to be testing?

josephcsible:
> Hmmm... Do we have a clever `Arbitrary` instance here, or is that only for sets? If the latter, now seems like a good time to fix that.

We have **an** `Arbitrary` instance: https://github.com/haskell/containers/blob/cd4733d9b857bed3aaa83e5d2cda3068b74ae719/containers-tests/tests/map-properties.hs#L234-L254

I'm not sure what you mean by whether it's "clever", though.

treeowl:
How would you feel about making the `Arbitrary` instance here more like the one for sets? That one tries to generate trees with the full range of shapes. Yes, you're right; `foldrWithKey` can't be tested properly. Let's own that and not try.

josephcsible:
I think `concatMap (\(k,v) -> [k,v])` is unavoidable, since there's no way to get a `[(k,v)]` from the `Bifoldable` instance. The rest of your suggestions should be done, though.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,46 +1450,116 @@ prop_splitModel n ys = length ys > 0 ==>
   in  toAscList l == sort [(k, v) | (k,v) <- xs, k < n] &&
       toAscList r == sort [(k, v) | (k,v) <- xs, k > n]
 
-prop_foldr :: Int -> [(Int, Int)] -> Property
-prop_foldr n ys = length ys > 0 ==>
-  let xs = List.nubBy ((==) `on` fst) ys
-      m  = fromList xs
+prop_fold :: Map Int Int -> Property
+prop_fold m = size m > 0 ==>
```
treeowl, line 1454: I don't see the need for the `size m > 0` precondition.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,46 +1450,116 @@ prop_splitModel n ys = length ys > 0 ==>
...
+  let xs = toList m
+  in  Foldable.fold (map (:[]) m) == List.map snd xs
+
+prop_foldMap :: Map Int Int -> Property
+prop_foldMap m = size m > 0 ==>
```
treeowl, line 1459: Same here.

containers-tests/tests/map-properties.hs
treeowl: This should use `Map Int A` instead of `Map Int Int`.

containers-tests/tests/map-properties.hs
```diff
@@ -1365,46 +1450,116 @@ prop_splitModel n ys = length ys > 0 ==>
...
+  let xs = toList m
+  in  Foldable.foldMap (:[]) m == List.map snd xs
+
+prop_foldMapWithKey :: Map Int Int -> Property
+prop_foldMapWithKey m = size m > 0 ==>
+  let xs = toList m
+  in  foldMapWithKey (\k v -> [k,v]) m == concatMap (\(k,v) -> [k,v]) xs
```
treeowl, line 1466: What is this trying to test? I don't understand. We could do something very simple, like this:

```haskell
prop_foldMapWithKey :: Map Int A -> Property
prop_foldMapWithKey = \m -> foldMapWithKey f m === foldMap (uncurry f) (toList m)
  where
    f k v = [(k, v)]
```

containers-tests/tests/map-properties.hs
```diff
@@ -1365,46 +1450,116 @@ prop_splitModel n ys = length ys > 0 ==>
...
+
+prop_foldr :: Int -> Map Int Int -> Property
+prop_foldr n m = size m > 0 ==>
+  let xs = toList m
   in  foldr (+) n m == List.foldr (+) n (List.map snd xs) &&
-      foldr (:) [] m == List.map snd (List.sort xs) &&
-      foldrWithKey (\_ a b -> a + b) n m == List.foldr (+) n (List.map snd xs) &&
+      foldr (:) [] m == List.map snd xs
```
treeowl, line 1472: Why not just

```haskell
prop_foldr :: Fun (A, B) B -> B -> Map Int A -> Property
prop_foldr c n m = foldr c' n m = foldr c' n (F.toList m)
  where
     c' = curry (apply f)
```

josephcsible: `Data.Foldable.toList` and `elems` are implemented in terms of `foldr`, so I didn't use them for it, but I did this for the rest of the folds.

treeowl:
Thanks!

PR #715, opened 2020-04-14
Document internals of liftA2

Lysxia:
The extra type information compared to `<*` (cf. similar additions at #713) helps quite a bit. The grouping of arguments feels like it makes sense here, I somehow have less of an itch to change it than with `<*`.

The most error-prone part is to not mix up `pr` and `sf` since they have the same type and play extremely similar roles. That's really the main motivation for invariant 1. Technically, it's necessary to prove a complete specification of `(<*>)`, but if it were not for `pr`/`sf`, I could convince myself the function is correct with only the other invariants about sizes and by trusting parametricity to fill in the gaps.

### DIFF
--- a/containers/src/Data/Sequence/Internal.hs
+++ b/containers/src/Data/Sequence/Internal.hs
@@ -531,7 +531,7 @@ apSeq fs xs@(Seq xsFT) = case viewl fs of
          RigidFull r@(Rigid s pr _m sf) -> Seq $
                Deep (s * length fs)
                     (fmap (fmap firstf) (nodeToDigit pr))
-                    (aptyMiddle (fmap firstf) (fmap lastf) fmap fs''FT r)
+                    (liftA2Middle (fmap firstf) (fmap lastf) fmap fs''FT r)
                     (fmap (fmap lastf) (nodeToDigit sf))
 {-# NOINLINE [1] apSeq #-}
 
@@ -591,7 +591,7 @@ liftA2Seq f xs ys@(Seq ysFT) = case viewl xs of
       RigidFull r@(Rigid s pr _m sf) -> Seq $
         Deep (s * length xs)
              (fmap (fmap (f firstx)) (nodeToDigit pr))
-             (aptyMiddle (fmap (f firstx)) (fmap (f lastx)) (lift_elem f) xs''FT r)
+             (liftA2Middle (fmap (f firstx)) (fmap (f lastx)) (lift_elem f) xs''FT r)
              (fmap (fmap (f lastx)) (nodeToDigit sf))
   where
     lift_elem :: (a -> b -> c) -> a -> Elem b -> Elem c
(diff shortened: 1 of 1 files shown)


containers/src/Data/Sequence/Internal.hs
```diff
@@ -650,13 +650,42 @@ type Digit23 a = Node a
...
+-- > 3. size (firstf y) = size (lastf y) = size (map23 x y) = size y
+-- >      for any (x :: a) (y :: b)
+--
+-- Projecting invariant 1 on sizes, using 2 and 3 to simplify, we have the
+-- following corollary.
+-- It is weaker than invariant 1, but it may be easier to keep track of.
+--
+-- > 1a. size result = s * (size fs + 1) + size m
+--
+-- In invariant 1, the types of the auxiliary functions are as follows:
```
treeowl, line 676: Could you either move these auxiliary functions to the top or mention at the top that they are described below? When I first read this I immediately wondered if these were real functions in the module somewhere, or if I was supposed to guess their meanings from their names, or what.

containers/src/Data/Sequence/Internal.hs
```diff
@@ -650,13 +650,42 @@ type Digit23 a = Node a
...
-  :: (b -> c)
-     -> (b -> c)
-     -> (a -> b -> c)
-     -> FingerTree (Elem a)
-     -> Rigid b
-     -> FingerTree (Node c)
+  :: (b -> c)              -- ^ @firstf@
+  -> (b -> c)              -- ^ @lastf@
+  -> (a -> b -> c)         -- ^ @map23@
+  -> FingerTree (Elem a)   -- ^ @fs@
```
treeowl, line 686: I don't really like the name `fs`, because while that's good for `<*>` it's a bit strange for `liftA2`; I kind of want to call that `xs`. But what are `firstf` and `lastf`? Not sure how to give those sensible names.

Lysxia: The binary function `a -> b -> c` could be just `f`, then the two `(b -> c)` functions are `f` applied to the first and last element of the toplevel `xs` (maybe called `ffirst`, `flast`?), and `fs` is the middle part of `xs`, so renamed to `midxs`.

```
xs = first <: midxs :> last
ffirst = f first
flast = f last
```

treeowl: Yes, that sounds better!

treeowl: Is it really the same `f`? If so, can we just pass the first and last `x`? I don't remember how this goes.

Lysxia: Good question. The difference is that the current version has better sharing. Otherwise instead of `fmap firstf` you would write `fmap (f first)`. In the case of an `f` that can share data when partially applied, that would cause that data to be recomputed at every recursive call. I can add a mention of this in the comments too.

treeowl: Do we also get that sharing in the middle?

treeowl: We really need a section of the documentation devoted to details of the performance of the `Applicative` instance.

Lysxia: > Do we also get that sharing in the middle?

I think so, `map23 x` will expand to `fmap @Node (... (fmap @Node (f x)) ...)` and the `fmap`'s will duplicate the closure with the one `x`. (It might be good to actually check that `fmap . map23`, in the recursive call of `aptyMiddle`, doesn't duplicate `map23 x`.) `map23` will be applied to each `x` in the middle once.

The problem with `firstf` and `lastf` is that they must be used multiple times to build the digits on the sides.

Lysxia: Linear types might be relevant to track this. I'm wildly speculating.

treeowl: Oof. Not going for linear types here! Ouch! I have been thinking it would be awfully nice to see if we can use Liquid Haskell in this module, mostly to prove totality of `zipWith` and `chunksOf` (where partial matches are *essential* for performance).

containers/src/Data/Sequence/Internal.hs
```diff
@@ -637,65 +637,99 @@ data Digit12 a = One12 a | Two12 a a
...
+--
+-- In invariant 1, the types of the auxiliary functions are as follows:
+--
+-- > toListFTE   :: FingerTree (Elem a) -> [a]
+-- > toListFTN   :: FingerTree (Node c) -> [c]
+-- > toListThinN :: Thin (Node b) -> [b]
+-- > toListD     :: Digit12 b -> [b]
+liftA2Middle
+  :: (b -> c)              -- ^ @ffirst@
+  -> (b -> c)              -- ^ @flast@
```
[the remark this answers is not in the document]
Lysxia: Yeah I can do that while it's fresh in my mind.

Lysxia:
Ready for review.

Do you prefer seeing the individual commits as this progresses or having them squashed so they're ready to merge at all times?

treeowl:
I tend to like separate commits when there's something interesting in an older one, or when a change may be controversial, or when different commits have different authors. Otherwise, I tend to like things fairly squashed.

treeowl:
Oh, and separate commits are good if they're doing unrelated things, of course.

treeowl:
Thanks a lot!

PR #755, opened 2020-10-30
Enable two warnings

tomjaguarpaw:
because they will be part of -Wall in the future, see
https://gitlab.haskell.org/ghc/ghc/-/issues/15656

Because containers is a boot library it's important that these warnings are not triggered because otherwise GHC's test cases will fail and it will be impossible to move these warnings into -Wall!

### DIFF
--- a/containers/containers.cabal
+++ b/containers/containers.cabal
@@ -35,7 +35,7 @@ Library
     default-language: Haskell2010
     build-depends: base >= 4.6 && < 5, array >= 0.4.0.0, deepseq >= 1.2 && < 1.5
     hs-source-dirs: src
-    ghc-options: -O2 -Wall
+    ghc-options: -O2 -Wall -fwarn-incomplete-uni-patterns -fwarn-incomplete-record-updates
 
     other-extensions: CPP, BangPatterns
 
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -13,6 +13,7 @@
 #endif
 
 {-# OPTIONS_HADDOCK not-home #-}
+{-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
 
 #include "containers.h"
 
--- a/containers/src/Data/IntMap/Strict/Internal.hs
+++ b/containers/src/Data/IntMap/Strict/Internal.hs
@@ -2,6 +2,8 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE PatternGuards #-}
 
+{-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
+
 #include "containers.h"
 
 -----------------------------------------------------------------------------
--- a/containers/src/Data/Sequence/Internal.hs
+++ b/containers/src/Data/Sequence/Internal.hs
@@ -17,6 +17,7 @@
 {-# LANGUAGE PatternGuards #-}
 
 {-# OPTIONS_HADDOCK not-home #-}
+{-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
 
 -----------------------------------------------------------------------------
 -- |


containers/containers.cabal
```diff
@@ -35,7 +35,7 @@ Library
     default-language: Haskell2010
     build-depends: base >= 4.6 && < 5, array >= 0.4.0.0, deepseq >= 1.2 && < 1.5
     hs-source-dirs: src
-    ghc-options: -O2 -Wall
+    ghc-options: -O2 -Wall -fwarn-incomplete-uni-patterns -fwarn-incomplete-record-updates
```
Lysxia, line 38: Why `-fwarn...` instead of `-W...`?

tomjaguarpaw: Because `-W` is post 8.0 only. I didn't actually check whether containers is supposed to support 7.x still, I just took the conservative option.

tomjaguarpaw:
Please let me know if there are any further questions about this.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -13,6 +13,7 @@
 #endif
 
 {-# OPTIONS_HADDOCK not-home #-}
+{-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
```
sjakobi, line 16: If these warnings matter, why are they simply disabled?

tomjaguarpaw: I'm not claiming they matter.

containers is a boot library. If it doesn't build then GHC will not build. These warnings [will be enabled by `-Wall` in futures GHCs](https://gitlab.haskell.org/ghc/ghc/-/issues/15656) therefore if containers triggers these warnings it will prevent GHC from building. I added these warnings to `ghc-options` not because they matter but because code triggering them shouldn't be added accidentally to containers.

tomjaguarpaw: Put another way, it is a means of achieving forward compatibility.

sjakobi: > I added these warnings to `ghc-options` not because they matter but because code triggering them shouldn't be added accidentally to containers.

How can we tell that potentially problematic code that would trigger the warnings is added, when the warning is disabled in several modules?

sjakobi: Basically, disabled warnings are code smell to me. I think it would be nice if the proponents of the new warning would actually address the problems in the code.

tomjaguarpaw: It's a very valid concern whether containers should contain code that is not clean for these warnings. However, that is not my concern in this particular PR.

My concern is that code in boot libraries (of which containers is one) should not trigger `incomplete-uni-patterns` and `incomplete-record-updates` *if thery were to be enabled*. That is because the plan is to enable these warnings in `-Wall`. As containers is built with `-Wall`, and GHC enables `-Werror` on its boot libraries, containers not being clean for these warnings will prevent them being added to `-Wall`.

To ensure that containers is clean for these warnings we have a few options, from most difficult to most simple

1. Explicitly enable the warnings in containers's cabal file, find all code that triggers them and rewrite it
2. Explicitly enable the warnings in container's cabal file, find all code that triggers them and disable the warnings on a per-module basis
3. Explicitly *dis*able the warnings in container's cabal file and do nothing to existing code

This PR implements option 2 because it has a nice combination of easy to implement and providing actual value to containers. If you prefer 3 then please let me know. I can do that. 1 is probably more controversial since it would require actually changing some of the code of containers.

But perhaps I'm not actually explaning sufficiently well the background to this change and why it is necessary. Would you like me to go into more detail?

tomjaguarpaw: > it would be nice if the proponents of the new warning would actually address the problems in the code.

That would be nice. However

1. This PR strictly increases the warning coverage of containers. The files in which this PR disables warnings do not currently have those warnings enabled either.

2. The proponents of the new warning [are the GHC committee](https://github.com/ghc-proposals/ghc-proposals/pull/71#issuecomment-364715424). I don't think it's unreasonable to ask boot libraries to make small changes to accommodate accepted GHC proposals.

sjakobi: > 1. Explicitly enable the warnings in containers's cabal file, find all code that triggers them and rewrite it
> 2. Explicitly enable the warnings in container's cabal file, find all code that triggers them and disable the warnings on a per-module basis
> 3. Explicitly _dis_able the warnings in container's cabal file and do nothing to existing code
> 
> This PR implements option 2 because it has a nice combination of easy to implement and providing actual value to containers. If you prefer 3 then please let me know. I can do that. 1 is probably more controversial since it would require actually changing some of the code of containers.

My preference would of course be option 1, because that would also demonstrate how code that currently triggers the warnings can be migrated and improved. That would also have brought https://github.com/haskell/containers/pull/591 to a proper conclusion.

But I understand that you're not trying to sell the warnings, since the decision to add them to `-Wall` has already been made.

I realize that `containers` won't be worse off after merging this PR, but the disabled warnings kind of feel like another "TODO". ;)

sjakobi (approved):

tomjaguarpaw:
Any chance we could get this merged to master? I'm trying to get these warnings into -Wall in GHC.

treeowl:
Yes, but not till this weekend, unless @sjakobi does it.

tomjaguarpaw:
Thanks very much

sjakobi:
Thanks for the PR, @tomjaguarpaw. I look forward to having these warnings included in `-Wall`.

PR #756, opened 2020-12-06
Implement 'intersections', and add an 'Intersection' newtype along with a 'Semigroup' instance

TOTBWF:
## Patch Description

This PR adds the following function for computing the intersection of a series of sets:
```
intersections :: Ord a => NonEmpty (Set a) -> Set a
intersections (s :| ss) = Foldable.foldl' intersection s ss
```

It also adds a newtype around `Set` dubbed `Intersection`, which provides a `Semigroup` instance that uses `intersections`.

Mailing list discussion can be found here: https://mail.haskell.org/pipermail/libraries/2020-December/030951.html

### DIFF
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -906,7 +906,7 @@ intersection t1@(Bin _ x l1 r1) t2
 #endif
 
 #if (MIN_VERSION_base(4,9,0))
--- | The intersection of a series of sets.
+-- | The intersection of a series of sets. Intersections are performed left-to-right.
 intersections :: Ord a => NonEmpty (Set a) -> Set a
 intersections (s :| ss) = Foldable.foldl' intersection s ss
 
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -908,7 +908,11 @@ intersection t1@(Bin _ x l1 r1) t2
 #if (MIN_VERSION_base(4,9,0))
 -- | The intersection of a series of sets. Intersections are performed left-to-right.
 intersections :: Ord a => NonEmpty (Set a) -> Set a
-intersections (s :| ss) = Foldable.foldl' intersection s ss
+intersections (s0 :| ss) = List.foldr go id ss s0
+    where
+      go s r acc
+          | null acc = empty
+          | otherwise = r (intersection acc s)
 
 -- | Sets form a 'Semigroup' under 'intersection'.
 newtype Intersection a = Intersection { getIntersection :: Set a }
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -156,9 +156,15 @@ module Data.Set.Internal (
             , unions
             , difference
             , intersection
+#if (MIN_VERSION_base(4,9,0))
             , intersections
+#endif
             , cartesianProduct
             , disjointUnion
+#if (MIN_VERSION_base(4,9,0))
+            , Intersection(..)
+#endif
+
 
             -- * Filter
             , filter
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -156,6 +156,7 @@ module Data.Set.Internal (
             , unions
             , difference
             , intersection
+            , intersections
             , cartesianProduct
             , disjointUnion
 
@@ -239,12 +240,13 @@ import Data.Monoid (Monoid(..))
 #endif
 #if MIN_VERSION_base(4,9,0)
 import Data.Semigroup (Semigroup(stimes))
+import Data.List.NonEmpty (NonEmpty(..))
 #endif
 #if !(MIN_VERSION_base(4,11,0)) && MIN_VERSION_base(4,9,0)
 import Data.Semigroup (Semigroup((<>)))
 #endif
 #if MIN_VERSION_base(4,9,0)
-import Data.Semigroup (stimesIdempotentMonoid)
+import Data.Semigroup (stimesIdempotentMonoid, stimesIdempotent)
 import Data.Functor.Classes
 #endif
 #if MIN_VERSION_base(4,8,0)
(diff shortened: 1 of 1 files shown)


ChrisPenner:
I went looking for this function today as well and was a little surprised to see it was missing.

treeowl:
The implementation should short-circuit. If the accumulator every becomes empty, then there's no need to continue.

treeowl:
Something like this:

```haskell
intersections ss = foldr go id ss empty
  where
    go s r acc
      | null acc = empty
      | otherwise = r (intersection acc s)
```

treeowl:
No, that last comment was wrong.

treeowl:
Corrected version: something like this:

```haskell
Something like this:

intersections (s0 :| ss) = foldr go id ss s0
  where
    go s r acc
      | null acc = empty
      | otherwise = r (intersection acc s)
```

TOTBWF:
Excellent point! I'll go update my PR over on the containers repo with this.

TOTBWF:
Actually, upon second thought, the implementation suggested already short
circuits, due to the laziness properties of intersection:

  intersection Tip _ = Tip

To confirm this, we can sprinkle in some trace statements:

  intersections $ (trace "forcing 0" $ fromList [1,2,3]) :| [trace "forcing
1" $ fromList [4,5,6], trace "forcing 2" $ fromList [7,8,9]]
  >>> fromList forcing 0
  >>> forcing 1
  >>> []

treeowl:
@TOTBWF, `foldl'` won't short-circuit regardless. In principle, you good use a naive `foldr` (with no accumulator), but that will be much worst than what I did in the non-short-circuiting case.

treeowl:
Please be sure to document that the intersections are taken from left to right. Even aside from short-cutting, the smaller the accumulator the better.

treeowl:
Still need to quit using `foldl'`. If you don't believe me, try an infinite list of `empty`.

TOTBWF:
Sorry about that, had a laziness brainfart! Will make the requested changes.

containers/src/Data/Set/Internal.hs
```diff
@@ -897,6 +905,24 @@ intersection t1@(Bin _ x l1 r1) t2
...
+          | null acc = empty
+          | otherwise = r (intersection acc s)
+
+-- | Sets form a 'Semigroup' under 'intersection'.
+newtype Intersection a = Intersection { getIntersection :: Set a }
+    deriving (Show, Eq, Ord)
+
+instance (Ord a) => Semigroup (Intersection a) where
+    (Intersection a) <> (Intersection b) = Intersection $ intersection a b
+    stimes = stimesIdempotent
```
sjakobi, line 923:
```suggestion
    sconcat = coerce intersections
    stimes = stimesIdempotent
```

containers/src/Data/Set/Internal.hs
```diff
@@ -897,6 +905,24 @@ intersection t1@(Bin _ x l1 r1) t2
 {-# INLINABLE intersection #-}
 #endif
 
+#if (MIN_VERSION_base(4,9,0))
+-- | The intersection of a series of sets. Intersections are performed left-to-right.
```
sjakobi, line 909: Doctests would be nice.

containers/src/Data/Set/Internal.hs
```diff
@@ -897,6 +905,24 @@ intersection t1@(Bin _ x l1 r1) t2
...
+intersections :: Ord a => NonEmpty (Set a) -> Set a
+intersections (s0 :| ss) = List.foldr go id ss s0
+    where
+      go s r acc
+          | null acc = empty
+          | otherwise = r (intersection acc s)
+
+-- | Sets form a 'Semigroup' under 'intersection'.
+newtype Intersection a = Intersection { getIntersection :: Set a }
```
sjakobi, line 918: It would be good to reference this newtype from the documentation for the `union`-based `Semigroup Set` instance.

containers/src/Data/Set/Internal.hs
sjakobi, line 909: Please mention the limitation to `base >= 4.9` in the docs.

containers/src/Data/Set/Internal.hs
```diff
@@ -897,6 +905,24 @@ intersection t1@(Bin _ x l1 r1) t2
...
+#if (MIN_VERSION_base(4,9,0))
+-- | The intersection of a series of sets. Intersections are performed left-to-right.
+intersections :: Ord a => NonEmpty (Set a) -> Set a
+intersections (s0 :| ss) = List.foldr go id ss s0
+    where
+      go s r acc
+          | null acc = empty
+          | otherwise = r (intersection acc s)
+
+-- | Sets form a 'Semigroup' under 'intersection'.
```
sjakobi, line 917: The docs should mention that this type is only included for `base >= 4.9`.

sjakobi (commented):
@TOTBWF could you summarise the feedback you got on the mailing list?

sjakobi:
If we add an `Intersection` newtype for `Set`, I think we should also add one for `IntSet`.

sjakobi:
> If we add an `Intersection` newtype for `Set`, I think we should also add one for `IntSet`.

And `intersections` too of course.

TOTBWF:
I don't think there was consensus. There were some +1's, but a lot of arguing over whether or not `intersections [] = empty` or not. I'm of the opinion that it shouldn't be the case, but this isn't universally held.

If we decide that this is worth adding, then I'll make the requested changes.

sjakobi:
> I don't think there was consensus. There were some +1's, but a lot of arguing over whether or not `intersections [] = empty` or not. I'm of the opinion that it shouldn't be the case, but this isn't universally held.

Yes, `intersections [] = empty` would suggest that `Intersection` has a `Monoid` instance which however wouldn't satisfy the monoid laws. It's good that `intersections` avoids this misunderstanding by taking a `NonEmpty`.

@dfeuer How do you feel about this PR?

Boarders:
Would be nice to get this merged!

treeowl:
Fell off my radar. Oops.

containers/src/Data/Set/Internal.hs
treeowl: Better, I think: include the type and let it be useless without the `Semigroup` instance.

containers/src/Data/Set/Internal.hs
treeowl: Do we need `intersections`? Isn't it just `sconcat`? I'm not 100% allergic to conditional exports (we have pattern synonyms for sequences), but I'd generally prefer to avoid them.

isovector:
Can we get this merged? It's a great addition.

treeowl:
There seem to be a few unresolved discussions, especially regarding documentation

treeowl:
Merged in 89ded1c3b8f5b5d96f7a6364d40f260280e0b741, 3fecabad3972e67e201e775823261519991262cd, eb6cf523080aa2d800fb33ba95815f6caf19a3bc, and d0c91736c856fa5adc159ddd9d52ed75a45f8f59.

meooow25:
`intersections` and the `Intersection` newtype was never exported from `Data.Set`. Seems like an oversight?

PR #757, opened 2020-12-06
Add `pop` and `popWithDefault` to `Map` and `IntMap`

martijnbastiaan:
Proposal:

* Add `pop` and `popWithDefault` to `Data.Map` and `Data.IntMap`.

Why:

* They're useful functions I expected to be in `Data.Map` and `Data.IntMap`. (This might be influenced by the fact that they're defined on Python's `dict`.)
* Their implementations (~ `updateLookupWithKey (\_ _ -> Nothing)`) is harder to parse than a simple `pop`, which should help Haskell codebases become a bit cleaner :). 
* Their implementations are a bit non-obvious. My first instinct was to write `(Map.lookup ..., Map.delete ...)`, which would have done two traversals. Having "properly" implemented functions in the lib would prevent people from writing their own suboptimal ones.

Mailing list discussion:

* https://mail.haskell.org/pipermail/libraries/2020-December/030954.html
* Reservations on name `pop`. Proposed alternatives: `extract`, `lookupRemove`. I'll leave it `pop` for now, until we've settled on a name.

### DIFF
--- a/containers/src/Data/Sequence/Internal.hs
+++ b/containers/src/Data/Sequence/Internal.hs
@@ -146,6 +146,8 @@ module Data.Sequence.Internal (
     adjust,         -- :: (a -> a) -> Int -> Seq a -> Seq a
     adjust',        -- :: (a -> a) -> Int -> Seq a -> Seq a
     update,         -- :: Int -> a -> Seq a -> Seq a
+    pop,            -- :: Int -> Seq a -> (Maybe a, Seq a)
+    popWithDefault, -- :: a -> Int -> Seq a -> (a, Seq a)
     take,           -- :: Int -> Seq a -> Seq a
     drop,           -- :: Int -> Seq a -> Seq a
     insertAt,       -- :: Int -> a -> Seq a -> Seq a
@@ -207,6 +209,8 @@ import Control.Applicative (Applicative(..), (<$>), (<**>),  Alternative,
 import qualified Control.Applicative as Applicative
 import Control.DeepSeq (NFData(rnf))
 import Control.Monad (MonadPlus(..))
+import qualified Control.Arrow as Arrow
+import Data.Maybe (fromMaybe)
 import Data.Monoid (Monoid(..))
 import Data.Functor (Functor(..))
 import Utils.Containers.Internal.State (State(..), execState)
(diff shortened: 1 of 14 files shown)


containers/src/Data/Sequence/Internal.hs
```diff
@@ -2461,6 +2465,26 @@ update i x (Seq xs)
...
+
+-- | \( O(\log(\min(i,n-i))) \). Lookup and delete.
+-- If the index is found, the value corresponding to it is returned and removed
+-- from the sequence. If the index is not found, a default value is returned
+-- together with the original map.
+--
+-- @since 0.6.5
+popWithDefault :: a -> Int -> Seq a -> (a, Seq a)
+popWithDefault a i s = Arrow.first (fromMaybe a) (pop i s)
+
```
treeowl, line 2487: I'm pretty sure `Arrow.first` is too lazy for this.

containers/src/Data/Sequence/Internal.hs
```diff
@@ -2461,6 +2465,26 @@ update i x (Seq xs)
   | fromIntegral i < (fromIntegral (size xs) :: Word) = Seq (updateTree (Elem x) i xs)
   | otherwise   = Seq xs
 
+-- | \( O(\log(\min(i,n-i))) \). Lookup and delete.
+-- If the index is found, the value corresponding to it is returned and removed
+-- from the sequence. If the index is not found, 'Nothing' is returned together
+-- with the original sequence.
+--
+-- @since 0.6.5
+pop :: Int -> Seq a -> (Maybe a, Seq a)
```
treeowl, line 2474: I don't like this at all, tradition be damned. Put the `Maybe` on the outside.

martijnbastiaan: Likewise for the `IntMap` / `Map` versions?

treeowl:
Yes.

martijnbastiaan:
@treeowl @emilypi @sjakobi How do we proceed with this PR? I think there are three points the mailing list made:

1. Name: `pop`, `lookupThenRemove`, `lookupRemove`, `lookupAndRemove`, `extract`.
2. Order of result tuple `(Maybe a, IntMap a)` vs `(IntMap a, Maybe a)` wrt monadic use.
3. Fairbairn threshold.

For what its worth, my personal opinions on the points made are:

1. `pop`, as a former Python programmer, still feels most natural to me. I think it also makes sense to copy names from one of the most popular programming languages instead of inventing our own. `lookup{then,And,}Remove` is pretty readable on its own. `extract` seems to be the worst of both worlds; it's neither copied from a big programming language, nor does it make sense without looking up documentation.
2. Map as state, maybe as computation result makes the most sense to me.
3. I don't particularly like the fairbarin threshold as an argument as it can be used to shut down _any_ function proposal bar the very basic ones. Determining how popular the proposed functions would seem almost impossible as the most straightforward implementation consists of two separate function calls that are okay-ish inline.

I'm happy to close the PR if there's no interest in merging it too, but I'd love some definitive answer :)

sjakobi:
Here's my view after re-reading the mailing list discussion:

2\. The result type should be `(Maybe v, Map k v)` – a `Nothing` result doesn't indicate that the map is empty, so a stateful computation would probably continue. This result type would also be consistent with the `insertLookupWithKey` and `updateLookupWithKey` functions.

1\. `lookupDelete` is the best name, because the new operation combines the `lookup` and `delete` operations. `pop` is worse, because it's unclear for people unfamiliar with the Python dict method. A `Seq` variant should possibly be named `lookupDeleteAt`, because it combines the `lookup` and `deleteAt` operations.

3\. I'm _weakly_ against adding this function to `containers`. There didn't seem to be a lot of demand for adding it, and it's easy to implement via `updateLookupWithKey`. It seems possible that a revised proposal would see more demand though.

---

Regarding a final decision, my understanding is that it's the maintainer's, i.e. @treeowl's call. [If he perceives the issue as controversial, he may ask the CLC for their view.](https://wiki.haskell.org/Core_Libraries_Committee#Controversial_decisions) I assume that this would require that `container` be adopted as a core library by the CLC though (which would be a good thing in my view). @emilypi or @chessai should be more familiar with the appropriate process.

treeowl:
There's certainly no need for CLC involvement in this sort of thing.

martijnbastiaan:
Very well. @treeowl: I'm happy to implement the changes mentioned by @sjakobi. If you feel this function shouldn't be added at all, this PR can be closed.

treeowl:
I'll take one more look this weekend, but I'm not optimistic.

martijnbastiaan:
Thanks for considering this PR. Given previous comments, this PR still needs changes even if the general idea is accepted. I don't want to spend any more time on it though, so I'm going to close it. If anyone wants to pick it up, feel free to clone my code.

epoberezkin:
`pop` (or `lookupDelete`) can be defined with available API as:

```
pop = alterF (,Nothing)
```

But it's very unintuitive. This API is needed frequently enough to justify it

PR #759, opened 2020-12-21
Add missing `ys` argument to shift2Right

Munksgaard:
Perhaps I'm misreading the code here, but I think there's supposed to be a `ys` argument here, right?


sjakobi (approved):
Nice catch!

treeowl:
Yes, definitely. Thanks.

PR #761, opened 2021-01-04
Add `Ord (Tree a)` instance.

Ericson2314:
I assume this is an oversight because we do have `Ord1 Tree`.

Some of you may be interested in knowing that I caught this due to trying out the the changes discusssed in the thread that began in https://mail.haskell.org/pipermail/libraries/2020-March/030306.html.


sjakobi (approved):

treeowl:
Thanks. Yes, this is quite clearly an oversight.

PR #762, opened 2021-01-19
Prepare for specialization of Data.List

phadej:
i.e. if it didn't export `Foldable` stuff.

### DIFF
--- a/containers/src/Data/Sequence/Internal.hs
+++ b/containers/src/Data/Sequence/Internal.hs
@@ -202,7 +202,6 @@ import Prelude hiding (
     null, length, lookup, take, drop, splitAt, foldl, foldl1, foldr, foldr1,
     scanl, scanl1, scanr, scanr1, replicate, zip, zipWith, zip3, zipWith3,
     unzip, takeWhile, dropWhile, iterate, reverse, filter, mapM, sum, all)
-import qualified Data.List
 import Control.Applicative (Applicative(..), (<$>), (<**>),  Alternative,
                             liftA2, liftA3)
 import qualified Control.Applicative as Applicative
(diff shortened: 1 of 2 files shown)


containers/src/Data/Sequence/Internal.hs
```diff
@@ -212,6 +211,11 @@ import Data.Monoid (Monoid(..))
 import Data.Functor (Functor(..))
 import Utils.Containers.Internal.State (State(..), execState)
 import Data.Foldable (Foldable(foldl, foldl1, foldr, foldr1, foldMap, foldl', foldr'), toList)
+import qualified Data.Foldable as F
+
+#if !(__GLASGOW_HASKELL__ >= 708)
```
sjakobi, line 216: Intuitively I'd prefer if CPP for `base` features was done with `MIN_VERSION_base`. I find it hard to remember the `base` version numbers though, so I wouldn't mind if we'd keep using `__GLASGOW_HASKELL__`.

sjakobi (approved):
Is there a GHC ticket for this `Data.List` specialization?

containers/src/Data/Sequence/Internal.hs
treeowl: Definitely should use `MIN_VERSION_base` for this. Is there any reasonable way to avoid CPP for this altogether?

phadej: Because he use of `Data.List` in the file is on `__GLASGOW_HASKELL__ >= 708`. https://github.com/haskell/containers/blob/master/containers/src/Data/Sequence/Internal.hs#L4388-L4392

I'm not keen to change / audit the CPP usage. `__GLASGOW_HASKELL_ >= 708` is as good as `MIN_VERSION_base` to guard `coerce` usage.

phadej: To be clear: say how you want this to be changed. I think the only way to make this nice is to drop pre-GHC-7.8 support (i.e. 7.6). But that rather be done in a separate patch going though whole of `containers`.

treeowl:
Sorry, didn't realize this was about coerce. That's fine.

phadej:
ping.

See https://downloads.haskell.org/ghc/8.10.3/docs/html/users_guide/using-warnings.html?highlight=wcompat#ghc-flag--Wcompat-unqualified-imports for motivation.

This patch is required to proceed with a plan.

treeowl:
Thanks.

PR #774, opened 2021-04-14
Fix map-strict-properties

infinity0:
It was actually testing the lazy version


sjakobi (approved):
Nice catch! I'm glad the tests still pass! 😅

treeowl:
Hmm.... Why would the tests pass if they were testing the wrong version?
Let's understand this better.

infinity0:
@treeowl The tests are defined with a CPP macro "STRICT" acting as a switch, the switch was set the wrong way but otherwise everything's consistent, so it was just testing the lazy version consistently.

PR #775, opened 2021-04-17
fix type naming consistency

infinity0:


treeowl:
What does this commit have to do with type naming? What is it for?

infinity0:
The file being changed contains some code for strict and lazy versions of IntMap, but mixes up invocation of functions from strict vs lazy modules. This commit fixes that.

treeowl:
Could you improve the commit message? Otherwise, looks good.

infinity0:
Sure, how about 75c1e92 ?

sjakobi (approved):
Thanks!

PR #782, opened 2021-06-21
Fix formatting

konsumlamm:
Fixes formatting in `Data.IntMap`/`Data.IntSet` (and related modules).
Due to a line break, the text in quotes was not cursive in the docs, which is fixed by removing the line break.

### DIFF
--- a/containers/src/Data/IntMap.hs
+++ b/containers/src/Data/IntMap.hs
@@ -46,9 +46,8 @@
 --      Workshop on ML, September 1998, pages 77-86,
 --      <http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.37.5452>
 --
---    * D.R. Morrison, \"/PATRICIA -- Practical Algorithm To Retrieve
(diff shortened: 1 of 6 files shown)


sjakobi (approved):
Thanks! This looks like a bug in haddock to me. Do you know whether it has been reported already?

treeowl:
Is there another way to break these lines while preserving the formatting? I'll take this if it's the best we can get.

konsumlamm:
> Thanks! This looks like a bug in haddock to me. Do you know whether it has been reported already?

I found https://github.com/haskell/haddock/issues/126, which comes to the conclusion that it's not worth it.

> Is there another way to break these lines while preserving the formatting? I'll take this if it's the best we can get.

One could write
```hs
--    * D.R. Morrison,
--      \"/PATRICIA -- Practical Algorithm To Retrieve Information Coded In Alphanumeric/\",
--      Journal of the ACM, 15(4), October 1968, pages 514-534.
```
but I haven't found any way to escape the line break.

konsumlamm:
ping @treeowl

treeowl:
Thanks.

Issue #788, opened 2021-07-22
user larger keys for IntMap tests


jwaldmann:
Prompted by https://github.com/haskell/containers/issues/787#issuecomment-884565646, I replaced the Arbitrary instance with
```
instance Arbitrary a => Arbitrary (IntMap a) where
  arbitrary = fmap (fromList . fmap (\(k,v) -> (getLarge k, v))) arbitrary
```
Good thing: all tests are green. So, no action required. We still might want to replace the instance?

That section of the code has heading ` Arbitrary, reasonably balanced trees`. I think "reasonable" is the wrong goal here -  for correctness tests, I want _all_ sorts of data, including "unreasonable". This might be different from performance tests where we might want _typical_, and that could mean "small keys", as produced by the current instance.

treeowl:
I certainly agree that we want extreme trees! Is `Large` a good way to get that? Or do we need to combine multiple generators? If you think this version is more thorough than what we did before, I'll happily merge.

Quick question: do you have any idea how to up the test count in CI?

jwaldmann:
>  Is `Large` a good way to get that? 

I have no idea, and no experience. `Large` looked to me like one standard way to quickly get Ints that use all bits. I found out about it only after I made the ad-hoc definition `2^x * y`. Quickcheck's generator https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/src/Test.QuickCheck.Arbitrary.html#arbitrarySizedBoundedIntegral  is different (better?) in the sense that it will use all bits (mine would just shift a few-bit pattern). On the other hand, an Int with a long tail of zeroes might be useful for the specific data structures that we want to test.

> If you think this version is more thorough than what we did before, I'll happily merge.

"more thorough" is hard to define/measure. One could fuzz the code and see how fast that's detected? (Nice student project ...)

> Quick question: do you have any idea how to up the test count in CI?

On the command line, I used `stack test --ta '-a 10000'`. [containers CI](.github/workflows/haskell-ci.yml)  uses cabal, I'm sure it has a way to pass options. This seems to work:
```
cabal test intset-properties --test-option="-a 10000"
```

The CI script is generated? Then we need to find a way to pass options through the generator?

treeowl:
I'll merge now. We can investigate more options later.

treeowl:
Oh wait, this isn't a PR. 🤦

jwaldmann:
I can make  a PR.

treeowl:
Thanks.

[...]

PR #790, opened 2021-07-22
Larger keys for intmap and intset

jwaldmann:
as discussed in https://github.com/haskell/containers/issues/788 .

I have no idea on how to best pass the "-a N" option for testing. Does it go in haskell-ci.yml, or in cabal.haskell-ci? or in containers.cabal? 

or in the source? That seems easiest to do, and most stable (independent from build/CI tooling) but it would ignore command-line options, which is bad? https://hackage.haskell.org/package/test-framework-0.8.2.0/docs/Test-Framework-Runners-Console.html#v:defaultMainWithArgs

Adding `--test-option="-a 10000"` does seem to increase CI build times - quite drastically. Or, they weren't very stable to begin with.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -250,7 +250,7 @@ jobs:
           $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --write-ghc-environment-files=always
       - name: tests
         run: |
-          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --test-show-details=direct --test-option="-a 2000"
+          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --test-show-details=direct --test-options="--quickcheck-tests 1000"  --test-options="--quickcheck-max-size 100"  --test-options="--timeout 10s"
       - name: haddock
         run: |
           if [ $((HCNUMVER >= 80200)) -ne 0 ] ; then $CABAL v2-haddock --haddock-all $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all ; fi
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -250,7 +250,7 @@ jobs:
           $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --write-ghc-environment-files=always
       - name: tests
         run: |
-          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --test-show-details=direct --test-option="-a 10000"
+          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --test-show-details=direct --test-option="-a 2000"
       - name: haddock
         run: |
           if [ $((HCNUMVER >= 80200)) -ne 0 ] ; then $CABAL v2-haddock --haddock-all $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all ; fi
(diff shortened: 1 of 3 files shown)


sjakobi:
> I have no idea on how to best pass the "-a N" option for testing. Does it go in haskell-ci.yml, or in cabal.haskell-ci? or in containers.cabal?
> 
> or in the source? That seems easiest to do, and most stable (independent from build/CI tooling) but it would ignore command-line options, which is bad? https://hackage.haskell.org/package/test-framework-0.8.2.0/docs/Test-Framework-Runners-Console.html#v:defaultMainWithArgs

So in `dhall`'s testsuite we have a little helper that we use to specify the number of QuickCheck tests in the code, but that can still be overriden on the CLI with a higher number of tests:

https://github.com/dhall-lang/dhall-haskell/blob/8aed63714def462056847f685f5871d74c12b6cf/dhall/tests/Dhall/Test/QuickCheck.hs#L787-L791

We're using `tasty` though – I don't know whether something similar can be achieved with `test-framework`. 

Since `test-framework` is basically unmaintained, and since `tasty`'s API has a large overlap with `test-framework`, maybe we should migrate to `tasty` anyway?

Bodigrim:
Migration to `tasty` usually boils down to renaming imports and wrapping top-level tests in `testGroup “All”`. The sooner we leave `test-framework` in the past, the better.

sjakobi:
@treeowl How do you feel about migrating to `tasty`?

treeowl:
I know nothing about test frameworks, so I'm very happy to let other people
decide.

treeowl:
Oh yeah, this discussion. Someone came up with something pretty decent (I think) for my little toy `compact-sequences` package. Does that look like something we could reasonably copy, or is it not up to the challenge of a big test suite?

treeowl:
Now that I've merged #796, can you do what you want with test case sizes here? Ideally, I like to run tests three ways:

1. Many small cases.
2. A moderate number of moderately sized cases.
3. A small number of large cases.

The three ways should be tweaked to run in the same amount of time (give or take an order of magnitude or so). This should be almost as good as SmallCheck for the small cases, but also be likely to catch glaring mistakes in large ones.

treeowl:
I'd feel more comfortable about this if we tested with multiple sizes. See https://github.com/treeowl/compact-sequences/commit/585bfac95611b30cf060a4c04ade9f2ee19fa3ed for a commit doing so elsewhere.

jwaldmann:
> tested with multiple sizes

OK, I'll look into this.

jwaldmann:
now the test run takes 25 min in CI (for each ghc version). that's too much?

treeowl:
Ouch. Yeah, that seems excessive. Did you limit that to just `IntMap` and `IntSet`?

jwaldmann:
no, I applied globally. https://github.com/jwaldmann/containers/commit/3a3929dff0b32707e226adb6cb17efdeb644e698

I theory, your method (lots of small, some large, etc.) should be fine for all tests?

treeowl:
Yeah, it should work, but it's pricy. `IntMap` and `IntSet` seem much more likely than the other structures to exhibit special behavior around size extremes.

jwaldmann:
I did restrict the test sets that get larger parameters. It's still expensive, and the programming gets ugly. I am not an expert at tasty's not-quite-regexps, and CI's not-quite-bash.

OTOH, perhaps 20 min (per GHC version) isn't too bad. If it catches bugs? The contributor can run faster tests locally. The maintainer needs days to accept the PR, anyway. Which is fine. (Though, a more clever CI would use that waiting time to run extra tests ... or even write some...)

treeowl:
Looking over the conversation, it's not clear whether everything is resolved. Could you possibly update?

meooow25:
The large keys part of this is obsolete (https://github.com/haskell/containers/issues/788#issuecomment-2253105117), but this PR also bumps the quickcheck args. Not sure if we still want to do that?

meooow25:
Closing for now. Bumping the number of quickcheck tests seems like a good idea, so we can still do that. It will probably need to be configured via the `cabal.haskell-ci` file.

PR #791, opened 2021-07-29
Fix simple typo in Set docs

hawkinsw:
In the overview of Set docs there is a first/third person error. This
patch fixes that problem. The commit message is significantly longer
than the patch.

### DIFF
--- a/containers/docs/set.rst
+++ b/containers/docs/set.rst
@@ -23,9 +23,9 @@ provided by the ``containers`` package: :haddock:`/Data.Set` and
    mechanism.
 
 
-All of these implementations are *immutable* which means that any update
-functions do not modify the set that you passed in, they creates a new set. In
-order to keep the changes you need to assign it to a new variable. For example::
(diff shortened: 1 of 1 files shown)


containers/docs/set.rst
```diff
@@ -24,7 +24,7 @@ provided by the ``containers`` package: :haddock:`/Data.Set` and
 
 
 All of these implementations are *immutable* which means that any update
-functions do not modify the set that you passed in, they creates a new set. In
+functions do not modify the set that you passed in, they create a new set. In
 order to keep the changes you need to assign it to a new variable. For example::
```
treeowl, line 28: While you're at it, that comma should probably be a semicolon, since it begins an independent clause.

hawkinsw:
If you only knew how much I wanted to make it a emdash :-)

Revision on the way.

hawkinsw:
Closing because I am submitting a different version with changes that are slightly grander in scope.

Issue #793, opened 2021-08-14
Add foldlWithKeyM


andrewthad:
The type would be:

    foldlWithKeyM :: Monad m => (a -> k -> b -> m a) -> a -> Map k b -> m a

I'm happy to contribute this if it would be accepted.

treeowl:
Why only the left fold?

andrewthad:
I could add the right variant as well. I haven’t ever needed it, but it would be good to have as well.

treeowl:
I think that's best, since the containers are mostly quite symmetrical.

andrewthad:
See https://github.com/haskell/containers/pull/795 and we can continue discussion there.

PR #795, opened 2021-08-16
Add strict foldlWithKeyM

andrewthad:
Resolves https://github.com/haskell/containers/issues/793

This is just the implementation of a single monadic fold. @treeowl if you confirm that this implementation of this particular function looks good, I can add all of the lazy folds and the right folds as well

### DIFF
--- a/containers/src/Data/Map/Internal.hs
+++ b/containers/src/Data/Map/Internal.hs
@@ -259,6 +259,10 @@ module Data.Map.Internal (
     , foldrWithKey'
     , foldlWithKey'
 
+    -- ** Monadic folds
+    , foldlWithKeyM
+    , foldrWithKeyM
+
     -- * Conversion
     , elems
     , keys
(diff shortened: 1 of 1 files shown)


andrewthad:
@treeowl Do you have any feedback on this?

andrewthad:
I've changed this in the most recent commit. I think I'll have:

* left monadic fold on map
* right monadic fold on map
* left monadic fold on intmap
* right monadic fold on intmap

Anything else you'd like to see in that list.

containers/src/Data/Map/Internal.hs
```diff
@@ -3365,16 +3365,16 @@ foldlWithKey' f z = go z
...
-foldlWithKeyM' :: Monad m => (a -> k -> b -> m a) -> a -> Map k b -> m a
-foldlWithKeyM' f z = go z
+-- | /O(n)/. Monadic variant of 'foldlWithKey'.
+foldlWithKeyM :: Monad m => (a -> k -> b -> m a) -> a -> Map k b -> m a
+foldlWithKeyM f z = go z
   where
-    go !z' Tip             = return z'
+    go z' Tip             = return z'
     go z' (Bin _ kx x l r) = do
       !z'' <- go z' l
```
treeowl, line 3374: This is still strict. I guess it would be okay to have both a lazy and strict version if you want, though it seems a bit redundant.

treeowl:
Yeah, could you get me some pizza? I've been missing the really good stuff.

containers/src/Data/Map/Internal.hs
andrewthad: Sorry, that was an oversight. This should be resolved in the latest commit.

andrewthad:
I live in Atlanta, so all the truly good pizza is a ways up the east coast from me.

treeowl:
Oh darn. Guess I'll have to make you pizza if you come visit (D.C. area).

ocramz:
@treeowl This seems like an uncontroversial (and welcome) addition, how can we move this PR forward?

andrewthad:
Is there anything I can do to move this forward?

treeowl:
I got sick a few weeks ago, and then went on family vacation. I expect to be ready to get back to my GitHub backlog by next Thursday. If this is ready for review, could you mark it as such?

andrewthad:
I've squashed and rebased. This adds `foldlWithKeyM` and `foldrWithKeyM` functions for the `Data.Map` data structure, both of which are lazy in the accumulator. It's ready for review.

PR #796, opened 2021-08-17
Migrate from test-framework to tasty

Bodigrim:
As discussed in #790.

### DIFF
--- a/containers-tests/tests/intmap-properties.hs
+++ b/containers-tests/tests/intmap-properties.hs
@@ -27,18 +27,16 @@ import qualified Prelude (map)
 import Data.List (nub,sort)
 import qualified Data.List as List
 import qualified Data.IntSet as IntSet
-import Test.Framework
-import Test.Framework.Providers.HUnit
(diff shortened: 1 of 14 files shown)


containers-tests/tests/intmap-properties.hs
```diff
@@ -1469,36 +1449,36 @@ prop_deleteMaxModel ys = length ys > 0 ==>
 prop_filter :: Fun Int Bool -> [(Int, Int)] -> Property
 prop_filter p ys = length ys > 0 ==>
   let xs = List.nubBy ((==) `on` fst) ys
-      m  = filter (apply p) (fromList xs)
+      m  = filter (applyFun p) (fromList xs)
```
treeowl, line 1452: Why this change? `apply` isn't deprecated, and it's shorter.

Bodigrim: No problem, let's stick to `apply`.

treeowl:
Thanks!

Issue #794, opened 2021-08-16
IntMap should offer lookup variant for likely successful lookup


Boarders:
Currently lookup is defined to fail early by checking if the prefix mask agrees with the key:
```haskell
lookup :: Key -> IntMap a -> Maybe a
lookup !k = go
  where
    go (Bin p m l r) | nomatch k p m = Nothing
                     | zero k m  = go l
                     | otherwise = go r
    go (Tip kx x) | k == kx   = Just x
                  | otherwise = Nothing
    go Nil = Nothing
 ```
 
 In the original paper they note that if we expect our lookups to be successful then we should instead prioritize testing if the kth bit is zero and failing only when we reach leaves. If I understand correctly, then this would just be the following code:
 ```haskell
 lookupSucc :: Key -> IntMap a -> Maybe a
lookupSucc !k = go
  where
    go (Bin _p m l r)
                     | zero k m  = go l
                     | otherwise = go r    
    go (Tip kx x) | k == kx   = Just x
                  | otherwise = Nothing
    go Nil = Nothing
 ```
 Is that correct? It seems to pass the tests  in any case. This can certainly lead to large speed ups when we know we are doing many successful lookups. For example, in the lookup benchmark we get the following (this is a best case scenario as the benchmark does not test a map with misses which seems like a bad design for a benchmark):
 ```terminal
 benchmarked lookup
time                 172.2 μs   (170.3 μs .. 174.1 μs)
                     0.998 R²   (0.994 R² .. 1.000 R²)
mean                 176.6 μs   (175.3 μs .. 181.2 μs)
std dev              7.148 μs   (2.786 μs .. 15.37 μs)

benchmarked lookupSucc
time                 153.0 μs   (151.6 μs .. 154.0 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 152.2 μs   (151.7 μs .. 152.7 μs)
std dev              1.783 μs   (1.306 μs .. 2.362 μs)
```

Even if this is not quite the correct implementation, containers should probably offer this sort of function to allow users to tune their code to their needs as both variants are worthwhile.

treeowl:
This sounds like an excellent idea. Could you *briefly* raise the question
of naming on the libraries list? It's even possible we'd want to call it
`lookup`, and name the fast-fail version something else. Did you test for
how bad this version is in an often-failing situation?

Boarders:
[...]

PR #800, opened 2021-09-09
Change IntMap.lookup and add new IntMap.query function

Boarders:
- `IntMap.lookup` no longer checks for short circuit failure.
  - Add a new function `IntMap.query` with the old fast-fail behaviour.
  
Addresses https://github.com/haskell/containers/issues/794

### DIFF
--- a/containers-tests/tests/intmap-properties.hs
+++ b/containers-tests/tests/intmap-properties.hs
@@ -295,12 +295,17 @@ test_notMember = do
 
 test_lookup :: Assertion
 test_lookup = do
-    employeeCurrency 1 @?= Just 1
-    employeeCurrency 2 @?= Nothing
+    employeeCurrency 1      @?= Just 1
+    employeeCurrency 2      @?= Just 2
+    employeeCurrency 3      @?= Just 3
+    employeeCurrency 4      @?= Just 4
+    employeeCurrency 5      @?= Nothing
+    employeeCurrency (2^10) @?= Just 42
+    employeeCurrency 6      @?= Nothing
   where
-    employeeDept = fromList([(1,2), (3,1)])
-    deptCountry = fromList([(1,1), (2,2)])
-    countryCurrency = fromList([(1, 2), (2, 1)])
+    employeeDept    = fromList [(1,2), (2, 14), (3, 10), (4, 18), (2^10, 100)]
+    deptCountry     = fromList [(1,1), (14, 14), (10, 10), (18, 18), (100, 100), (2,2)]
+    countryCurrency = fromList [(1, 2), (2, 1), (14, 2), (10, 3), (18, 4), (100, 42)]
     employeeCurrency :: Int -> Maybe Int
     employeeCurrency name = do
         dept <- lookup name employeeDept
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -594,25 +594,22 @@ notMember k m = not $ member k m
 
 -- | /O(min(n,W))/. Lookup the value at a key in the map. See also 'Data.Map.lookup'.
 
--- See Note: Local 'go' functions and capturing]
+-- See Note: Local 'go' functions and capturing
 lookup :: Key -> IntMap a -> Maybe a
 lookup !k = go
   where
-    go (Bin p m l r) | nomatch k p m = Nothing
-                     | zero k m  = go l
-                     | otherwise = go r
+    go (Bin _p m l r) | zero k m  = go l
+                      | otherwise = go r
     go (Tip kx x) | k == kx   = Just x
                   | otherwise = Nothing
     go Nil = Nothing
-
 
 -- See Note: Local 'go' functions and capturing]
 find :: Key -> IntMap a -> a
 find !k = go
   where
-    go (Bin p m l r) | nomatch k p m = not_found
-                     | zero k m  = go l
-                     | otherwise = go r
+    go (Bin _p m l r) | zero k m  = go l
+                      | otherwise = go r
     go (Tip kx x) | k == kx   = x
                   | otherwise = not_found
     go Nil = not_found
@@ -943,8 +940,7 @@ adjust f k m
 -- > adjustWithKey f 7 empty                         == empty
 
 adjustWithKey ::  (Key -> a -> a) -> Key -> IntMap a -> IntMap a
-adjustWithKey f !k t@(Bin p m l r)
-  | nomatch k p m = t
+adjustWithKey f !k (Bin p m l r)
   | zero k m      = Bin p m (adjustWithKey f k l) r
   | otherwise     = Bin p m l (adjustWithKey f k r)
 adjustWithKey f k t@(Tip ky y)
@@ -976,8 +972,7 @@ update f
 -- > updateWithKey f 3 (fromList [(5,"a"), (3,"b")]) == singleton 5 "a"
 
 updateWithKey ::  (Key -> a -> Maybe a) -> Key -> IntMap a -> IntMap a
-updateWithKey f !k t@(Bin p m l r)
-  | nomatch k p m = t
+updateWithKey f !k (Bin p m l r)
   | zero k m      = binCheckLeft p m (updateWithKey f k l) r
   | otherwise     = binCheckRight p m l (updateWithKey f k r)
 updateWithKey f k t@(Tip ky y)
@@ -998,8 +993,7 @@ updateWithKey _ _ Nil = Nil
 -- > updateLookupWithKey f 3 (fromList [(5,"a"), (3,"b")]) == (Just "b", singleton 5 "a")
 
 updateLookupWithKey ::  (Key -> a -> Maybe a) -> Key -> IntMap a -> (Maybe a,IntMap a)
-updateLookupWithKey f !k t@(Bin p m l r)
-  | nomatch k p m = (Nothing,t)
+updateLookupWithKey f !k (Bin p m l r)
   | zero k m      = let !(found,l') = updateLookupWithKey f k l
                     in (found,binCheckLeft p m l' r)
   | otherwise     = let !(found,r') = updateLookupWithKey f k r
--- a/containers-tests/benchmarks/IntMap.hs
+++ b/containers-tests/benchmarks/IntMap.hs
@@ -11,10 +11,18 @@ import Data.Maybe (fromMaybe)
 import Prelude hiding (lookup)
 
 main = do
-    let m = M.fromAscList elems :: M.IntMap Int
-    evaluate $ rnf [m]
+    let m     = M.fromAscList elems_hits   :: M.IntMap Int
+    let m'    = M.fromAscList elems_mid    :: M.IntMap Int
+    let m''   = M.fromAscList elems_most   :: M.IntMap Int
+    let m'''  = M.fromAscList elems_misses :: M.IntMap Int
+    let m'''' = M.fromAscList elems_mixed  :: M.IntMap Int
+    evaluate $ rnf [m, m', m'', m''', m'''']
     defaultMain
-        [ bench "lookup" $ whnf (lookup keys) m
+        [ bench "lookup_hits" $ whnf (lookup keys) m
+        , bench "lookup_half" $ whnf (lookup keys) m'
+        , bench "lookup_most" $ whnf (lookup keys) m''
+        , bench "lookup_misses" $ whnf (lookup keys'') m'''
+        , bench "lookup_mixed" $ whnf (lookup keys) m''''
         , bench "insert" $ whnf (ins elems) M.empty
         , bench "insertWith empty" $ whnf (insWith elems) M.empty
         , bench "insertWith update" $ whnf (insWith elems) m
@@ -44,12 +52,23 @@ main = do
                     (M.fromList $ zip [1..10] [1..10])
         ]
   where
-    elems = zip keys values
+    elems = elems_hits
+    elems_hits   = zip keys values
+    elems_mid    = zip (map (+ (2^12 `div` 2)) keys) values
+    elems_most   = zip (map (+ (2^12 `div` 10)) keys) values
+    elems_misses = zip (map (\x-> x * 2 + 1) keys) values
+    elems_mixed = zip mixedKeys values
+    --------------------------------------------------------
     keys = [1..2^12]
+    keys' = fmap (+ 1000000) keys
+    keys'' = fmap (* 2) [1..2^12]
+    mixedKeys = interleave keys keys'
     values = [1..2^12]
+    --------------------------------------------------------
     sum k v1 v2 = k + v1 + v2
     consPair k v xs = (k, v) : xs
 
+------------------------------------------------------------
 add3 :: Int -> Int -> Int -> Int
 add3 x y z = x + y + z
 {-# INLINE add3 #-}
@@ -95,3 +114,8 @@ alt xs m = foldl' (\m k -> M.alter id k m) m xs
 maybeDel :: Int -> Maybe Int
 maybeDel n | n `mod` 3 == 0 = Nothing
            | otherwise      = Just n
+
+------------------------------------------------------------
+interleave :: [Int] -> [Int] -> [Int]
+interleave [] ys = ys
+interleave (x:xs) (y:ys) = x : y : interleave xs ys


Boarders:
These are the benchmarks I get on my machine (lookup now doesn't fast fail and query is the old behaviour):
```terminal
benchmarked query_hits
time                 276.0 μs   (266.1 μs .. 284.9 μs)
                     0.996 R²   (0.994 R² .. 0.999 R²)
mean                 261.0 μs   (259.0 μs .. 263.7 μs)
std dev              8.073 μs   (5.881 μs .. 11.27 μs)

benchmarked query_half
time                 135.6 μs   (132.5 μs .. 139.1 μs)
                     0.997 R²   (0.995 R² .. 0.999 R²)
mean                 135.9 μs   (134.9 μs .. 137.1 μs)
std dev              3.739 μs   (3.134 μs .. 4.913 μs)

benchmarked query_most
time                 248.7 μs   (243.7 μs .. 254.8 μs)
                     0.993 R²   (0.988 R² .. 0.997 R²)
mean                 245.7 μs   (242.5 μs .. 250.9 μs)
std dev              13.82 μs   (9.804 μs .. 21.46 μs)
variance introduced by outliers: 33% (moderately inflated)

benchmarked query_misses
time                 252.6 μs   (241.0 μs .. 261.1 μs)
                     0.983 R²   (0.971 R² .. 0.992 R²)
mean                 309.3 μs   (298.5 μs .. 319.2 μs)
std dev              37.71 μs   (32.95 μs .. 43.33 μs)
variance introduced by outliers: 71% (severely inflated)

benchmarked query_mixed
time                 35.88 μs   (35.03 μs .. 36.58 μs)
mean                 35.55 μs   (35.31 μs .. 35.78 μs)
std dev              809.9 ns   (685.5 ns .. 1.008 μs)

benchmarked lookup_hits
time                 231.5 μs   (227.3 μs .. 237.0 μs)
                     0.996 R²   (0.993 R² .. 0.998 R²)
mean                 240.3 μs   (238.3 μs .. 243.1 μs)
std dev              8.042 μs   (6.614 μs .. 10.46 μs)
variance introduced by outliers: 15% (moderately inflated)

benchmarked lookup_half
time                 236.0 μs   (227.6 μs .. 251.5 μs)
                     0.985 R²   (0.976 R² .. 0.993 R²)
mean                 232.6 μs   (229.1 μs .. 237.5 μs)
std dev              13.75 μs   (10.51 μs .. 17.64 μs)
variance introduced by outliers: 37% (moderately inflated)

benchmarked lookup_most
time                 220.1 μs   (217.9 μs .. 223.4 μs)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 223.8 μs   (222.0 μs .. 225.7 μs)
std dev              6.178 μs   (5.237 μs .. 7.243 μs)

benchmarked lookup_misses
time                 231.7 μs   (227.4 μs .. 239.1 μs)
                     0.996 R²   (0.991 R² .. 0.999 R²)
mean                 229.9 μs   (228.0 μs .. 232.9 μs)
std dev              7.857 μs   (6.027 μs .. 10.50 μs)
variance introduced by outliers: 16% (moderately inflated)

benchmarked lookup_mixed
time                 20.19 μs   (19.37 μs .. 20.97 μs)
                     0.994 R²   (0.991 R² .. 0.998 R²)
mean                 20.56 μs   (20.31 μs .. 20.81 μs)
std dev              853.2 ns   (692.1 ns .. 1.152 μs)
variance introduced by outliers: 22% (moderately inflated)
```

containers/src/Data/IntMap/Internal.hs
```diff
@@ -598,7 +599,22 @@ notMember k m = not $ member k m
...
-    go (Bin p m l r) | nomatch k p m = Nothing
+    go (Bin _p m l r)
+                     | zero k m  = go l
+                     | otherwise = go r
+    go (Tip kx x) | k == kx   = Just x
+                  | otherwise = Nothing
+    go Nil = Nothing
+
+-- | /O(min(n,W))/. Query has identical behaviour to 'Data.IntMap.Internal.lookup' but
+-- will fail faster in the case that the key does not share  is not present.
```
konsumlamm, line 610:
```suggestion
-- | /O(min(n,W))/. 'query' has identical behaviour to 'lookup', but
-- will fail faster in the case that the key is not present.
```

konsumlamm (changes requested):

treeowl:
And add that it's otherwise slower?

containers/src/Data/IntMap/Internal.hs
Boarders: How embarrassing, thank you!

Boarders:
@treeowl : done.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -598,7 +599,23 @@ notMember k m = not $ member k m
...
+    go (Bin _p m l r)
+                     | zero k m  = go l
+                     | otherwise = go r
+    go (Tip kx x) | k == kx   = Just x
+                  | otherwise = Nothing
+    go Nil = Nothing
+
+-- | /O(min(n,W))/. 'query' has identical behaviour to 'Data.IntMap.Internal.lookup', but
+-- will fail faster in the case that the key is not present. It is otherwise
+-- preferable to use `Data.IntMap.Internal.lookup'.
```
konsumlamm, line 611:
```suggestion
-- | /O(min(n,W))/. 'query' has identical behaviour to 'Data.IntMap.Internal.lookup', but
-- will fail faster if the key is not present. When the key is (likely) present, it is
-- preferable to use 'Data.IntMap.Internal.lookup' instead.
```
This is clearer IMO.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -598,7 +599,23 @@ notMember k m = not $ member k m
...
+
+-- See Note: Local 'go' functions and capturing]
```
konsumlamm, line 613: You apparently just copied this from `lookup`, but there's an unmatched `]`.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -598,7 +599,23 @@ notMember k m = not $ member k m
 lookup :: Key -> IntMap a -> Maybe a
 lookup !k = go
   where
-    go (Bin p m l r) | nomatch k p m = Nothing
+    go (Bin _p m l r)
+                     | zero k m  = go l
+                     | otherwise = go r
```
konsumlamm, line 604:
```suggestion
    go (Bin _p m l r)  | zero k m  = go l
                       | otherwise = go r
```
This doesn't really matter, but it's consistent with the rest of the definition.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -598,7 +599,23 @@ notMember k m = not $ member k m
...
+-- preferable to use `Data.IntMap.Internal.lookup'.
+
+-- See Note: Local 'go' functions and capturing]
+query :: Key -> IntMap a -> Maybe a
+query !k = go
+  where
+    go (Bin p m l r)
+                     | nomatch k p m = Nothing
                      | zero k m  = go l
                      | otherwise = go r
```
konsumlamm, line 620:
```suggestion
    go (Bin p m l r) | nomatch k p m = Nothing
                     | zero k m  = go l
                     | otherwise = go r
```
Same as above.

containers-tests/tests/intmap-properties.hs
```diff
@@ -307,6 +308,20 @@ test_lookup = do
...
         lookup country countryCurrency
 
+test_query :: Assertion
+test_query = do
+    employeeCurrency 1 @?= Just 1
+    employeeCurrency 2 @?= Nothing
+  where
+    employeeDept = fromList([(1,2), (3,1)])
+    deptCountry = fromList([(1,1), (2,2)])
+    countryCurrency = fromList([(1, 2), (2, 1)])
```
konsumlamm, line 318:
```suggestion
    employeeDept = fromList [(1, 2), (3, 1)]
    deptCountry = fromList [(1, 1), (2, 2)]
    countryCurrency = fromList [(1, 2), (2, 1)]
```
Seems you copied this as well, but the parentheses are just unnecessary (Haskell functions are not called like `f(...)` after all).

containers/src/Data/IntMap/Internal.hs
Boarders: Ok, it is worth saying that is is more subtle in reality than the key being present or not present and depends instead on whether the key shares a bit prefix with the current tree keys but that subtlety is hard to capture.

containers-tests/tests/intmap-properties.hs
Boarders: Yes, I just kept what was in place, I'll change both and extend the test as well a small amount.

Boarders: Let me know if you would like any changes to the new test inputs.

sjakobi:
Apologies for the late response!

> * `IntMap.lookup` no longer checks for short circuit failure.

Frankly, I don't think this is a good idea. We don't know whether any users are relying on `lookup` to fail fast. `containers` is used about everywhere, including packages that don't have active maintainers. I don't think we should release a potentially large performance regression to the ecosystem.

> * Add a new function `IntMap.query` with the old fast-fail behaviour.

I don't think this is good naming. Imagine you're a Haskell newbie, starting to use `Data.IntMap`. How do you remember which of `lookup` and `query` have the desired fail-fast or fail-slow behaviour?

I think it would be better to add a more explicit name, such as `lookupFailSlow` (or `lookupFailFast`).

treeowl:
I disagree with you on this one, @sjakobi. While it's true we can't change anything without regressing some things, I think in the vast majority of practical cases, lookups are likely to succeed and we should do what we can to make that the fast path. Change logs, announcements, and documentation should prompt anyone relying on short-circuiting to switch functions.

mitchellwrosen:
Furthermore, from the discussion on the original issue (see https://github.com/haskell/containers/issues/794#issuecomment-916245832 for example), it sounds like it's not as simple as, "if you expect [some vague threshold percentage] of misses, use the fail-fast variant"!

Boarders:
@mitchellwrosen That is correct (and the above benchmarks show it), the `nomatch` function relies on the prefix stored in the tree to differ before the switching bit of what we are looking up. That means it will be good at short circuiting keys that are far away from those in the map but not those that are close (where closeness is in terms of shared bit prefix). I don't think this is actually a common occurrence and so I don't think the short-circuit default is good.

@sjakobi I understand where you are coming from, why don't I try and see if I can test this change on something that makes large use of intmap (I wanted to try ghc) and see how it goes? I think for smaller consumers of intmap it probably won't make a big difference and for large users we can make the necessary patches to those packages that use the current behaviour in critical ways. I am open to changing the name but I don't think it should be optimized for newbies per se who should not be caring about such issues as to the internals of how intmap works.

nomeata:
I am on board with @treeowl here: Most changes will benefit some users, and regress some others. My intuition is that this will benefit most users. Definitely those where all lookups succeed, which might actually be a common case (think of the environment in a compiler, all variables will be present in the map). But likely also those users that have failing lookups, but where the looked up keys and the present values are nicely mixed, so the early failure isn’t very early, and the benefit of fewer branches may more. 

But in the end, only benchmarks can tell. Looking forward to @Boarders’s investigation here (although good luck getting statistically significant results here…)

nomeata (commented):
Shouldn’t the implementatoin `find`of find be changed as well? The type signature of `find` indicates that the user is _expecting_ success, so here we probably benefit especially well from this optimization?

treeowl:
Yes, `find` for sure; that was mentioned early on but maybe lost.

Boarders:
I'll add it shortly.

Boarders:
How about `findWithDefault`?

nomeata:
Not that anyone would be surprised by this, but the Coq formalization proves that this implementation is semantically correct: https://github.com/plclub/hs-to-coq/pull/188

sjakobi:
Apologies! I should have fully read the discussion in #794 before commenting here.

So I read that failing lookups don't _necessarily_ get slower, and some in fact get faster with this patch.

Nevertheless I'm pretty impressed by the massive 6x slowdown of the `lookup_fails` benchmark reported in https://github.com/haskell/containers/issues/794#issuecomment-899191262.

Is this already the worst contrived case possible? I was imagining that there's some weird corner case where we have to perform 63 successive `zero` checks before concluding that the key is not present. What's the slowdown for the absolute worst case?

Boarders:
@sjakobi As I said, these worst case scenarios seem very unlikely to representative and involve lots of lookups of keys with bitpatterns without shared prefixes to anything in the map, that doesn't seem like what the library should be optimizing for.

nomeata:
> What's the slowdown for the absolute worst case?

I think it’d be that one: A map that is maximally deep (already bad – no balanced map will be that deep, so we are already at a “bad” application of IntMap), and then looking up a key that will make `lookup` walk to that maximum depth, when it otherwise could fail imediatelly:
```
Prelude Data.IntMap> let m = unions [ singleton (2^n) () | n <- [0..62]]
Prelude Data.IntMap> Data.IntMap.lookup (2^63) m
Nothing
```

sjakobi:
@Boarders I fully agree that this library shouldn't optimize for worst-case scenarios. Nevertheless I think we ought to avoid shipping perf regressions to users with unusual use cases for `IntMap.lookup`. That's why I'm curious what the "long tail" of possible regressions due to this change looks like.

If we accept that some users may encounter serious perf regressions due to this change, we can for example consider a staged rollout of these changes: We could add a `nomatch`-less `lookup` variant in the next minor release, and wait with changing the default `lookup` until the next major release.

treeowl:
I just don't think what we currently optimize for is likely to be dominant
in any realistic application. "Usually we look up things that are far away
from anything in the map, and we need to fail quickly" sounds like some
kind of strange joke.

Boarders:
Granted, I have no idea what I am doing, but here is what I got from trying to measure this change on GHC building cabal using perf:

original:
```
   777,509,663,120      instructions              #    1.26  insn per cycle         
   617,691,499,581      cycles                                                      

     143.578308984 seconds time elapsed

     142.807996000 seconds user
       1.146751000 seconds sys
```

with this change:
```
   776,453,461,794      instructions              #    1.32  insn per cycle         
   590,402,339,813      cycles                                                      

     134.333660622 seconds time elapsed

     133.517496000 seconds user
       1.228369000 seconds sys
  ```
  
  Doesn't look like there is much in it and I don't really know how variable these sort of measurements are, but definitely doesn't look like any sort of serious regression.

**Comment by Boarders** [comment]
Are there some other serious users of `IntMap` I could run some benchmarks against does anyone know?

**Comment by sjakobi** [comment]
> Are there some other serious users of `IntMap` I could run some benchmarks against does anyone know?

I had the same question with #340. I think Clash seemed interesting, but it also depended on GHC which made it tricky to build.

I believe @jwaldmann has an `IntMap` benchmark that you could try.

**Comment by Boarders** [comment]
Not sure what the deal is with the timed-out CI.

**Comment by treeowl** [comment]
I just restarted CI. We'll see.

**Comment by Boarders** [comment]
Vanessa McHale (@vmchale) kindly informed me that her compiler, [kempe](https://github.com/vmchale/kempe), makes heavy use of IntMaps and so I ran the benchmarks using the main containers branch and this branch and get the following (output courtesy of Ben Gamari's [Criterion cmparison](https://github.com/bgamari/criterion-compare) tool):

image

**Comment by Boarders** [comment]
I tried measuring this against some other packages but the synthetic benchmarks were so variable that nothing valuable seemed to come out of the exercise.

**Comment by treeowl** [comment]
I think we should probably just go ahead. Making performance more
consistently good is just good policy.

**Comment by nomeata** [comment]
What does “go ahead” mean?

If the conclusion is that very likely no serious regressions in real applications are expected, then I would be in favor of changing the implementation of `lookup` and `find`, without extending the API at all. If (and only if) we later learn that there really are users who strongly preferred the old implementations, we can offer both. But I would not complicate the API without hard evidence that someone needs it.

**Comment by treeowl** [comment]
@nomeata , that sounds like a pretty reasonable approach, yes.

**Comment by Boarders** [comment]
Ok, this evening I will remove query and then it looks like we should be good to go on getting this merged.

**Comment by treeowl** [comment]
Yes. I won't delay this any further. But afterwards, we should really look at `alter` and `alterF`. Presumably the same applies to them. We really only want to shortcut for intersection, union, difference, and merge.

**Comment by Boarders** [comment]
Ah, let I didn't realise about alter, I'll make those changes too.

**Comment by Boarders** [comment]
I looked into changing `alter` and did the rather naive thing. The current definition is as follows:
```haskell
alter :: (Maybe a -> Maybe a) -> Key -> IntMap a -> IntMap a
alter f !k t@(Bin p m l r)
  | nomatch k p m = case f Nothing of
                      Nothing -> t
                      Just x -> link k (Tip k x) p t
  | zero k m      = binCheckLeft p m (alter f k l) r
  | otherwise     = binCheckRight p m l (alter f k r)
alter f k t@(Tip ky y)
  | k==ky         = case f (Just y) of
                      Just x -> Tip ky x
                      Nothing -> Nil
  | otherwise     = case f Nothing of
                      Just x -> link k (Tip k x) ky t
                      Nothing -> Tip ky y
alter f k Nil     = case f Nothing of
                      Just x -> Tip k x
                      Nothing -> Nil
```

I changed that to:
```haskell
alter :: (Maybe a -> Maybe a) -> Key -> IntMap a -> IntMap a
alter f !k t@(Bin p m l r)
  | zero k m      = binCheckLeft p m (alter f k l) r
  | otherwise     = binCheckRight p m l (alter f k r)
alter f k t@(Tip ky y)
  | k==ky         = case f (Just y) of
                      Just x -> Tip ky x
                      Nothing -> Nil
  | otherwise     = case f Nothing of
                      Just x -> link k (Tip k x) ky t
                      Nothing -> Tip ky y
alter f k Nil     = case f Nothing of
                      Just x -> Tip k x
                      Nothing -> Nil
```

This definition (strangely to me!) fails the lazy property tests (but _not_ the strict ones) as it fails the `commonPrefix` property. In particular, if we use the original definition of `alter` on an example quickcheck is unhappy with then we get the following:
```haskell
λ: m = fromList [(-5, ()), (-2, ())]
λ: f = \case {Nothing -> Just (); Just () -> Nothing}
λ: m' = alter f 0 m
λ: m'
fromList [(-5,()),(-2,()),(0,())]
```
If on the other hand we remove the `nomatch` test then we get:
```haskell
λ: m = fromList [(-5, ()), (-2, ())]
λ: f = \case {Nothing -> Just (); Just () -> Nothing}
λ: m' = alter f 0 m
λ: m'
fromList [(0,()),(-5,()),(-2,())]
```

I didn't investigate further to figure out why this is a specific problem with `Data.IntMap.Lazy`. If there is a more obviously correct implementation then please do let me know.

treeowl:
Ah, I see. I got things rather mixed up. That won't work. What *should* work is modifying `adjustWithKey` to remove the `nomatch` case.

treeowl:
Also `updateWithKey`, I believe.

treeowl:
(And `updateLookupWithKey`.)

Boarders:
Ok, thanks, should have it by tomorrow.

treeowl:
Oh yeah, and delete....

jwaldmann:
> I believe @jwaldmann has an `IntMap` benchmark that you could try.

https://gitlab.imn.htwk-leipzig.de/waldmann/containers-benchmark

Boarders:
I will get back to finishing this work as soon as I am back from holiday!

Boarders:
@treeowl Apologies for the delay in making these changes but I have finally gotten around to it. Is there anything else to do here?

@jwaldmann When I try to clone your repository I get permission denied for some reason.

treeowl:
Thanks a lot!

PR #806, opened 2021-11-17
Enable TypeOperators to use (~)

int-index:
[GHC Proposal #371](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0371-non-magical-eq.md) requires `TypeOperators` to use type equality `a~b`. The following lines are affected:

```
containers/src/Data/Sequence/Internal.hs
4416:instance a ~ Char => IsString (Seq a) where
```

The fix is to simply enable the extension.


monoidal:
Pinging @treeowl.

treeowl:
Thanks.

PR #813, opened 2022-01-18
Add Lift instances

treeowl:
* `Data.Sequence`: Add `Lift` instances for `Seq`, `ViewL`, and `ViewR`. The `Seq` instance
tries to be a bit clever about the shape of the resulting tree and the
size of the splice; everything else is straightforward.
* Others: Add `Lift` instances for `Data.IntMap.IntMap`, `Data.IntSet.IntSet`, `Data.Map.Map`, `Data.Set.Set`,
`Data.Tree.Tree`, and `Data.Graph.SCC`.

### DIFF
--- a/containers-tests/containers-tests.cabal
+++ b/containers-tests/containers-tests.cabal
@@ -37,6 +37,7 @@ common deps
       array    >=0.4.0.0
     , base     >=4.9.1   && <5
     , deepseq  >=1.2     && <1.5
+    , template-haskell
 
 common test-deps
   import: deps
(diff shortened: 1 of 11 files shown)


treeowl:
This PR is based off the branch in #812.

treeowl:
@phadej, it turned out I needed a bit of tweaking to get those `coerce` calls to typecheck, I think particularly in untyped splices. Could you have a look and see if you think it's all good? I also added some tests for `Data.Sequence`, since that's the only non-derived instance.

treeowl:
One question in the back of my mind is whether there would be any value in normalizing the representations of `Set`s and `Map`s. We'd get a guarantee that `==` sets/maps would produce identical splices, as long as the underlying keys/values made that guarantee as well. But does it matter?

treeowl:
Since no one seems to have any comments, I'll merge this.

PR #832, opened 2022-04-25
Fix mention of `foldl'` in Haddocks of `unions`/`unionsWith`

amesgen:
Tiny documentation fix, was confused by this for a split second today 😆 

Before:

After:

### DIFF
--- a/containers/src/Data/Map/Internal.hs
+++ b/containers/src/Data/Map/Internal.hs
@@ -1773,7 +1773,7 @@ maxView t = case maxViewWithKey t of
   Union.
 --------------------------------------------------------------------}
 -- | The union of a list of maps:
---   (@'unions' == 'Prelude.foldl' 'union' 'empty'@).
+--   (@'unions' == 'Foldable.foldl'' 'union' 'empty'@).
 --
 -- > unions [(fromList [(5, "a"), (3, "b")]), (fromList [(5, "A"), (7, "C")]), (fromList [(5, "A3"), (3, "B3")])]
 -- >     == fromList [(3, "b"), (5, "a"), (7, "C")]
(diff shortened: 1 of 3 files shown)


amesgen:
Another option would be to remove the inlined implementation from the Haddocks all-together, as the up-to-date actual implemenation is just one click away.

treeowl:
The documentation should have everything a user needs to know to understand both the meaning of the function and its performance. In this case, I think the easiest way to do that is to give the actual implementation.

treeowl (approved):

treeowl:
In the future, please refrain from including images in issues or issue discussions unless they contribute substantially; they are generally bad for accessibility. A diagram of an idea or a rendering of a formatting change would be fine, but these images weren't helpful.

containers/src/Data/Map/Internal.hs
```diff
@@ -1773,7 +1773,7 @@ maxView t = case maxViewWithKey t of
   Union.
 --------------------------------------------------------------------}
 -- | The union of a list of maps:
```
konsumlamm, line 1775:
```suggestion
-- | The union of a 'Foldable' of maps:
```

containers/src/Data/Map/Internal.hs
```diff
@@ -1788,7 +1788,7 @@ unions ts
 #endif
 
 -- | The union of a list of maps, with a combining operation:
```
konsumlamm, line 1790:
```suggestion
-- | The union of a 'Foldable' of maps, with a combining operation:
```

konsumlamm (commented):
There's actually another mistake: The functions work on any `Foldable`, not just lists.

amesgen:
@konsumlamm Personally, I have no opinion on whether that is misleading (after all, `Foldable` can be characterized by `toList`). In any case, if this change is desired, it should also be changed in  `Data.Map.Strict.Internal`/ `Data.Set.Internal`/`Data.IntSet.Internal`, which might or might not be in scope of this PR.

containers/src/Data/Map/Internal.hs
treeowl: I would go with "`'Foldable' container`".

PR #847, opened 2022-07-23
Apply the MathJax changes from #830 to the new functions from #817

josephcsible:


treeowl:
Oops, I missed that when merging. Thanks.

Issue #814, opened 2022-01-26
Alternative to keysSet that uses Arg to keep the value


josephcsible:
```haskell
import qualified Data.Map.Internal as Map
import qualified Data.Set.Internal as Set
import Data.Semigroup (Arg(..))

mapToArgSet :: Map.Map k a -> Set.Set (Arg k a)
mapToArgSet Map.Tip = Set.Tip
mapToArgSet (Map.Bin sz k v l r) = Set.Bin sz (Arg k v) (mapToArgSet l) (mapToArgSet r)
```

This function seems safe to me, since `Arg k a`'s `Eq` and `Ord` instances only care about `k` and not `a`. Assuming I'm right about that, would this be worth adding to `Data.Map` (or maybe `Data.Set`), as a more powerful alternative to `keysSet`?

treeowl:
It's safe enough. It could be defeated by overlapping instances, but that won't cause a memory safety problem. I really like this idea. Could you write the libraries mailing list to see if anyone has suggested improvements to the proposed API?

treeowl:
What about going the other way, by the way? And should we have unsafe versions that let the user provide the conversion function?

josephcsible:
> Could you write the libraries mailing list to see if anyone has suggested improvements to the proposed API?

Sure, I'll do this tonight.

> What about going the other way, by the way?

You can already go the other way with the existing public API, like this: `mapKeysMonotonic (\(Arg k _) -> k) . fromSet (\(Arg _ v) -> v)` I'm not opposed to adding this too if you still want it though.

treeowl:
How does that go the other way?

josephcsible:
> And should we have unsafe versions that let the user provide the conversion function?

I'm not sure exactly what you mean here. What's the type of the functions you're asking for?

josephcsible:
> How does that go the other way?

The thing I originally proposed has type `Map.Map k a -> Set.Set (Arg k a)`, and the thing I said you can already do has type `Set.Set (Arg k a) -> Map.Map k a`.

treeowl:
Oh, I see now. Sleepy, so got confused. What I meant about an unsafe version was

```haskell
unsafeToSet :: (k -> a -> b) -> Map k a -> Set b
unsafeToMap :: (b -> (k, a)) -> Set b -> Map k a
```

Probably not by those names though.

treeowl:
(The unsafe things could also live in an `Applicative`.)

josephcsible:
[...]

PR #848, opened 2022-07-25
Fix #818: add unsafe functions for converting between Set and Map

### LINKED ISSUE #818: Add unsafe functions for converting between Set and Map
In #814, @treeowl asked for functions like these:
```haskell
unsafeToSet :: (k -> a -> b) -> Map k a -> Set b
unsafeToMap :: (b -> (k, a)) -> Set b -> Map k a
```
And also ones that live in an `Applicative`, which I assume means ones like these:
```haskell
unsafeToSetA :: Applicative t => (k -> a -> t b) -> Map k a -> t (Set b)
unsafeToMapA :: Applicative t => (b -> t (k, a)) -> Set b -> t (Map k a)
```
Note the lack of `Ord` constraints: these functions are unsafe in that it's the user's responsibility to make sure the `k` and `b` values agree on their ordering, just as it is for `mapMonotonic`. I think those are reasonable functions to add, but a little bit out of scope of that issue, hence this one.

josephcsible:


treeowl:
Can you give some examples of situations where you've wanted this?

josephcsible:
@treeowl I don't have a use of my own for these. I wrote them because you said you wanted them in #814.

treeowl:
My memory really does have the sieve nature. I will attempt to remember what the heck I was thinking and whether it makes sense or not. Argh.

meooow25:
Closing since we don't have a motivation for #818.

PR #855, opened 2022-09-05
Set test

coot:


treeowl:
For what little it's worth, my guess is that GHC changed something about the low-level closure inspection API, and `nothunks` hasn't been updated to take it into account. But I don't know enough about that package to even be able to file an intelligent bug report.

coot:
Yes, that's plausible. I cannot reproduce it with `ghc-9.4.2`, so it seems I need a newer version of `ghc`.  This should make it easier to find what changed in `ghc`.

coot:
It fails with the prerelease `ghc-9.4.0.20220523`, but it runs just fine with `ghc-9.4.1` (and `ghc-9.4.2`).

treeowl:
Ugh, it seems I didn't update the Haskell CI quite right; I tried to add 9.4 but ... didn't. Not sure what I did wrong.

treeowl:
Well, we actually want both of them. And that file shouldn't be edited by hand.

treeowl:
OK, I see where I goofed up. Sorry for wasting everyone's time.

treeowl:
Thanks for your help anyway, @coot. I've stolen your typo fix.

PR #858, opened 2022-09-07
Use less cpp and don't rely on base version for liftA2 comaptibility

googleson78:
Relying on base version should work fine, once `base` is bumped. But it currently isn't, causing warnings (again) when compiling ghc. To this end, we can employ this solution, which both avoids this issue, as well as uses less CPP.

### DIFF
--- a/containers/src/Prelude.hs
+++ b/containers/src/Prelude.hs
@@ -4,19 +4,15 @@
 -- @liftA2@ wasn't previously exported from the standard prelude.
 module Prelude
   ( module Prel
-#if !MIN_VERSION_base(4,18,0)
   , Applicative (..)
-#endif
 #if !MIN_VERSION_base(4,10,0)
   , liftA2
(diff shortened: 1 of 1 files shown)


treeowl:
Good idea. Ping me if you can after it passes CI; I'll merge.

googleson78:
@treeowl ping

treeowl:
@Bodigrim, could you push this to the GHC side?

Bodigrim:
@treeowl I assume @googleson78 has tested it against his GHC branch, so please merge.

treeowl:
@Bodigrim I've already merged it here. I just don't know what needs to happen to push it over to GHC, and I'm pretty busy with other things.

googleson78:
I'm also pushing it on the GHC side, so no need to think about that.

treeowl:
Great, thanks.

PR #860, opened 2022-09-29
Fix build paths in gitignore

alexfmpe:
Current .gitignore isn't matching these dirs
```
containers-tests/dist-newstyle/
containers-tests/dist/
containers/dist-newstyle/
containers/dist/
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,8 @@
 *.swo
 *.tix
 .hpc/
-/dist/*
-/dist-newstyle/*
+**/dist/*
+**/dist-newstyle/*
 GNUmakefile
 dist-install
 ghc.mk


treeowl:
Is there any risk someone might want a directory named `dist` somewhere in the source tree? Or is that too far-fetched?

alexfmpe:
It would be a risky thing to do in a haskell repo and it being in gitignore would likely stand out as when one tried to add it to version control. That said, I'm not opposed to hardcoding `containers` and `containers-tests` instead of `**`. Would that be better?

treeowl:
No, I think it's probably fine. If we run into trouble we'll deal with it then.

treeowl:
Thanks.

PR #869, opened 2022-11-13
Fix some code formatting in docs

meooow25:
Between `@`s Haddock requires escaping of special characters like `\`, `<`, `>`. Some code snippets don't have this and the rendered output can be off, sometimes making it [incorrect (`<$>` -> `$`)](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Lazy.html#v:traverseWithKey). This PR fixes the ones I found.

### DIFF
--- a/containers/src/Data/Map/Internal.hs
+++ b/containers/src/Data/Map/Internal.hs
@@ -1899,7 +1899,7 @@ difference t1 (Bin _ k _ l2 r2) = case split k t1 of
 -- | \(O\bigl(m \log\bigl(\frac{n+1}{m+1}\bigr)\bigr), \; m \leq n\). Remove all keys in a 'Set' from a 'Map'.
 --
 -- @
--- m \`withoutKeys\` s = 'filterWithKey' (\k _ -> k ``Set.notMember`` s) m
+-- m \`withoutKeys\` s = 'filterWithKey' (\\k _ -> k ``Set.notMember`` s) m
 -- m \`withoutKeys\` s = m ``difference`` 'fromSet' (const ()) s
 -- @
 --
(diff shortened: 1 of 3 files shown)


containers/src/Data/Map/Internal.hs
```diff
@@ -2467,10 +2467,8 @@ filterMissing f = WhenMissing
 -- | Filter the entries whose keys are missing from the other map
 -- using some 'Applicative' action.
 --
--- @
--- filterAMissing f = Merge.Lazy.traverseMaybeMissing $
---   \k x -> (\b -> guard b *> Just x) <$> f k x
--- @
+-- > filterAMissing f = Merge.Lazy.traverseMaybeMissing $
+-- >   \k x -> (\b -> guard b *> Just x) <$> f k x
```
treeowl, line 2471: This is a much more radical change. Why?

meooow25: It's cleaner in source code without the escaping, in my opinion. `>` is like `@` but takes the string literally.
The other instances cannot use `>` because they have markup linking to other parts of the documentation.
I could switch this one to `@` if you prefer.

treeowl: Don't we want `Merge.Lazy.traverseMaybeMissing` to be a link?

meooow25: We might, but it would be a little inconsistent since all [adjacent codes](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Internal.html#v:filterMissing) and the corresponding functions [in IntMap](https://hackage.haskell.org/package/containers-0.6.6/docs/Data-IntMap-Internal.html#v:filterAMissing) all don't link to the functions they use.

treeowl:
All right. I guess we can go with it for now.

meooow25:
Does this require more changes?

treeowl:
Nope. Thanks.

PR #872, opened 2022-11-19
Avoid `Data.List.{head,tail}`

Bodigrim:
CLC has approved the proposal to add `{-# WARNING #-}` to `Data.List.{head,tail}` (https://github.com/haskell/core-libraries-committee/issues/87). It means that usage of `head` and `tail` will emit compile-time warnings.

This patch eliminates all usages of `head` and `tail` in `containers`. They are confined to a couple of debug functions, so should not have any effect on performance or semantics.

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -3504,7 +3504,7 @@ showsBars :: [String] -> ShowS
 showsBars bars
   = case bars of
       [] -> id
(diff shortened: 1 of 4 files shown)


treeowl (approved):
These look like improvements all around. Thanks.

Bodigrim:
@treeowl CI is green.

treeowl:
Thanks.

PR #873, opened 2022-11-19
Unboxing and streamlining Map maps

treeowl:
* Use an unboxed-sum version of `Maybe` to implement `mapMaybeWithKey`. This potentially (I suspect usually) allows all the `Maybe`s to be erased.

* Comprehensive rewrite rules for both strict and lazy versions of `map`, `mapWithKey`, `mapMaybeWithKey`, and `filterWithKey` quickly get out of hand. Following `unordered-containers`, tame the mess by implementing both lazy and strict mapping functions in terms of versions that use unboxed results. Rewrite rules on these underlying functions will then apply uniformly. One concern: I found it a bit tricky to get the unfoldings I wanted; lots of things had to be marked `INLINABLE` explicitly.


treeowl:
Hmm.... For compatibility with potential non-GHC implementations, this needs a few changes:

1. Don't use `MagicHash`. Even though it's confusing, we'll need to use something else to distinguish these unboxed things.
2. If it doesn't break performance, use an unlifted newtype like
    ```haskell
    newtype SoloU a = SoloU (# a #)
    ```
    with a pattern synonym. That way, we can implement both a renamed `Maybe#` and `SoloU` either with or without unboxed sums/tuples.

treeowl:
Making progress... still don't know if performance will work. Do we have any map/coerce tests? I forget.

treeowl:
@meooow25 Do you happen to have the bandwidth to see what effect this PR has on the benchmarks?

meooow25:
I'm not very familiar with rewrite rules, but sure I could try running the benchmarks later today.

meooow25:
Here's what I got with `map-benchmarks`, GHC 9.2.5.  
Attaching as a file since it's quite long: [benchmarks.txt](https://github.com/haskell/containers/files/10058174/benchmarks.txt)

treeowl:
Thanks a lot! I'm glad to see that performance does not seem to decrease. Unfortunately, I believe the two apparent improvements are almost certainly benchmarking artifacts. I'm disappointed that there don't seem to be any improvements to `mapMaybe` or `mapMaybeWithKey`. I wonder if that would change for a larger map size, where garbage collection is more expensive.

PR #877, opened 2022-11-22
properly markup `<$>` in documentation of `alterF`

Flupp:
For both `IntMap` variants, the `<` and `>` did not appear in the haddock output.

While at it, also in the documentation for the other `alterF` variants the `<$>` was turned into a hyperlink.


meooow25 (changes requested):
Hi, thanks for fixing the bad Haddocks for IntMap.

> While at it, also in the documentation for the other `alterF` variants the `<$>` was turned into a hyperlink.

I would prefer it if we _did not_ do this, because the link is a bit distracting when it is not the focus of the documentation. Do you mind reverting those?

Flupp:
I thought it made the markup more uniform and might also be helpful for beginners.

Anyways, I now just added escaping where needed.

I also rebased on current master.

meooow25 (approved):
Great, thanks!

PR #879, opened 2022-11-27
Faster IntSet and IntMap split functions

meooow25:
Following up from [#874](https://github.com/haskell/containers/pull/874#discussion_r1027248480).

The main changes are using `bin` everywhere and being strict in `IntSet.splitMember`. Also added one missing property test , and benchmarks to measure the change.

IntSet:
```
  split:dense:        OK (0.20s)
    46.4 ns ± 2.7 ns, 49% less than baseline
  split:sparse:       OK (0.18s)
    84.2 ns ± 5.5 ns, 41% less than baseline
  splitMember:dense:  OK (0.20s)
    46.3 ns ± 3.6 ns, 47% less than baseline
  splitMember:sparse: OK (0.17s)
    82.1 ns ± 5.4 ns, 45% less than baseline
```

IntMap:
```
  split:       OK (0.19s)
    82.3 ns ± 6.4 ns, 43% less than baseline
  splitLookup: OK (0.17s)
    84.5 ns ± 5.9 ns, 43% less than baseline
```

### DIFF
--- a/containers/src/Data/IntSet/Internal.hs
+++ b/containers/src/Data/IntSet/Internal.hs
@@ -933,8 +933,16 @@ splitMember x t =
   where
     go x' t'@(Bin p m l r)
         | nomatch x' p m = if x' < p then (Nil, False, t') else (t', False, Nil)
-        | zero x' m      = case go x' l of (lt, fnd, gt) -> (lt, fnd, bin p m gt r)
-        | otherwise      = case go x' r of (lt, fnd, gt) -> (bin p m l lt, fnd, gt)
+        | zero x' m =
+          case go x' l of
+            (lt, fnd, gt) ->
+              let !gt' = bin p m gt r
+              in (lt, fnd, gt')
+        | otherwise =
+          case go x' r of
+            (lt, fnd, gt) ->
+              let !lt' = bin p m l lt
+              in (lt', fnd, gt)
     go x' t'@(Tip kx' bm)
         | kx' > x'          = (Nil, False, t')
           -- equivalent to kx' > prefixOf x'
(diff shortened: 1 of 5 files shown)


meooow25:
Side note: Antitone functions run much slower than split functions. For example, `spanAntitone` takes 1.6-2.6x more time than `splitMember` on the current benchmarks. Inlining it makes it just as fast, but would we want to do that?

treeowl:
Impressive results! Where did you add strictness in `splitMember`? I can't find it. I don't *think* we want to `INLINE` the general antitone functions, but it would be good to make sure that they get unfoldings and to document that inclining them might be a good idea in hot spots.

meooow25:
I see. I'll check if that works separately.

containers/src/Data/IntSet/Internal.hs
```diff
@@ -913,22 +916,33 @@ split x t =
...
+        | nomatch x' p m = if x' < p then (Nil, False, t') else (t', False, Nil)
+        | zero x' m =
+          case go x' l of
+            (lt, fnd, gt) ->
+              let !gt' = bin p m gt r
+              in (lt, fnd, gt')
+        | otherwise =
+          case go x' r of
+            (lt, fnd, gt) ->
+              let !lt' = bin p m l lt
```
meooow25, line 944: > Where did you add strictness in splitMember?

`gt'` and `lt'` here.

treeowl: Thanks. Now I see it. Should we add a strict triple type alongside our strict pair type? It wouldn't have to be in this PR.

meooow25: Apparently `Data.Map` has one, and uses it for this very purpose: https://github.com/haskell/containers/blob/aebfe5ed88e37b8fc5373ec033c23d6b40faad61/containers/src/Data/Map/Internal.hs#L3856

However `Data.Set` uses bangs in `splitMember` (like this PR), and `Data.IntMap` defines a different strict triple for `splitLookup`. Might be a good idea to move the strict triple to utils and use it in all these places. Probably better in a separate PR though, yes.

meooow25:
Hi @treeowl, anything else to be done here?

PR #880, opened 2022-12-07
Faster Tree folds

### LINKED ISSUE #878: Poor performance of Foldable for Data.Tree.Tree
Folds on `Tree` seem to perform very poorly.

```hs
{-# LANGUAGE BangPatterns #-}
import Data.Foldable
import Data.Monoid
import Data.Tree
import Test.Tasty.Bench

main :: IO ()
main = defaultMain
    [ env (pure tree_) $ \t -> bgroup ""
        [ bench "sum"              $ whnf sum t
        , bench "foldMap'Tree Sum" $ whnf (foldMap'Tree Sum) t
        ]
    ]
  where
    lim = 1000 * 1000
    tree_ = unfoldTree (\x -> (x, filter (<=lim) [2*x, 2*x+1])) 1 :: Tree Int

foldMap'Tree :: Monoid m => (a -> m) -> Tree a -> m
foldMap'Tree f = go mempty
  where go !z (Node x ts) = foldl' go (z <> f x) ts
```
```
    sum:              OK (1.82s)
      249  ms ±  15 ms
    foldMap'Tree Sum: OK (0.31s)
      10.2 ms ± 717 μs
```

meooow25:
Add some `Foldable` functions for `Data.Tree.Tree`.

* `foldMap`, `null`: Already exist
* `foldr`, `foldl'`: Added, basic functions
* `fold`, `elem`: Added, default definitions but `INLINABLE` helps
* `foldr1`, `foldl1`, `maximum`, `minimum`: Added, default definitions introduce `Maybe`s since these can be partial, we can avoid all that since trees are non-empty.
* `foldMap'`, `toList`, `length`, `sum`, `product`: Not added, defaults work well enough
* `foldr'`, `foldl`: Not added or benchmarked, unlikely to be used(?)

Benchmarks:

```
  fold:    OK (0.25s)
    15.5 μs ± 962 ns, 85% less than baseline
  foldMap: OK (0.14s)
    18.0 μs ± 1.7 μs, 82% less than baseline
  foldr:   OK (0.18s)
    21.8 μs ± 1.5 μs, 93% less than baseline
  foldl':  OK (0.30s)
    18.1 μs ± 878 ns, 95% less than baseline
  foldr1:  OK (0.16s)
    19.6 μs ± 1.5 μs, 94% less than baseline
  foldl1:  OK (0.30s)
    18.0 μs ± 1.2 μs, 95% less than baseline
  toList:  OK (0.17s)
    39.9 μs ± 2.9 μs, 31% less than baseline
  elem:    OK (0.27s)
    16.3 μs ± 777 ns, 47% less than baseline
  maximum: OK (0.16s)
    19.6 μs ± 1.5 μs, 94% less than baseline
  sum:     OK (0.30s)
    17.9 μs ± 671 ns, 95% less than baseline
```

Fixes #878.

### DIFF
--- a/containers-tests/tests/tree-properties.hs
+++ b/containers-tests/tests/tree-properties.hs
@@ -5,23 +5,37 @@ import Data.Tree as T
 import Control.Applicative (Const(Const, getConst), pure, (<$>), (<*>), liftA2)
 
 import Test.Tasty
+import Test.Tasty.HUnit
 import Test.Tasty.QuickCheck
 import Test.QuickCheck.Function (apply)
-import Test.QuickCheck.Poly (A, B, C)
+import Test.QuickCheck.Poly (A, B, C, OrdA)
 import Control.Monad.Fix (MonadFix (..))
 import Control.Monad (ap)
+import Data.Foldable (foldl', toList)
+import Data.Traversable (foldMapDefault)
 
 default (Int)
 
 main :: IO ()
 main = defaultMain $ testGroup "tree-properties"
          [
-           testProperty "monad_id1"                prop_monad_id1
+           testCase     "foldr"                    test_foldr
+         , testProperty "monad_id1"                prop_monad_id1
          , testProperty "monad_id2"                prop_monad_id2
          , testProperty "monad_assoc"              prop_monad_assoc
          , testProperty "ap_ap"                    prop_ap_ap
          , testProperty "ap_liftA2"                prop_ap_liftA2
          , testProperty "monadFix_ls"              prop_monadFix_ls
+         , testProperty "toList"                   prop_toList
+         , testProperty "foldMap"                  prop_foldMap
+         , testProperty "foldl'"                   prop_foldl'
+         , testProperty "foldr1"                   prop_foldr1
+         , testProperty "foldl1"                   prop_foldl1
+         , testProperty "foldr_infinite"           prop_foldr_infinite
+         , testProperty "maximum"                  prop_maximum
+         , testProperty "minimum"                  prop_minimum
+         , testProperty "sum"                      prop_sum
+         , testProperty "product"                  prop_product
          ]
 
 {--------------------------------------------------------------------
@@ -53,8 +67,23 @@ instance Arbitrary a => Arbitrary (Tree a) where
 #endif
 
 ----------------------------------------------------------------
+-- Utilities
+----------------------------------------------------------------
+
+data Magma a
+  = Inj a
+  | Magma a :* Magma a
+  deriving (Eq, Show)
+
+----------------------------------------------------------------
 -- Unit tests
 ----------------------------------------------------------------
+
+test_foldr :: Assertion
+test_foldr = do
+  foldr (:) [] (Node 1 []) @?= [1]
+  foldr (:) [] (Node 1 [Node 2 [Node 3 []]]) @?= [1..3]
+  foldr (:) [] (Node 1 [Node 2 [Node 3 [], Node 4 []], Node 5 [Node 6 [], Node 7 []]]) @?= [1..7]
 
 ----------------------------------------------------------------
 -- QuickCheck
@@ -101,3 +130,39 @@ prop_monadFix_ls val ta ti =
     f :: (Int -> Int) -> Int -> Tree (Int -> Int)
     f q y = let t = apply ti y
             in fmap (\w -> fact w q) t
+
+prop_toList :: Tree A -> Property
+prop_toList t = toList t === foldr (:) [] t
+
+prop_foldMap :: Tree A -> Property
+prop_foldMap t =
+  foldMap (:[]) t === toList t .&&.
+  foldMap (:[]) t === foldMapDefault (:[]) t
+
+prop_foldl' :: Tree A -> Property
+prop_foldl' t = foldl' (flip (:)) [] t === reverse (toList t)
+
+prop_foldr1 :: Tree A -> Property
+prop_foldr1 t = foldr1 (:*) (fmap Inj t) === foldr1 (:*) (map Inj (toList t))
+
+prop_foldl1 :: Tree A -> Property
+prop_foldl1 t = foldl1 (:*) (fmap Inj t) === foldl1 (:*) (map Inj (toList t))
+
+prop_foldr_infinite :: NonNegative Int -> Property
+prop_foldr_infinite (NonNegative n) =
+    forAllShow genInf (const "<possibly infinite tree>") $
+        \t -> length (take n (foldr (:) [] t)) <= n
+  where
+    genInf = Node () <$> oneof [listOf genInf, infiniteListOf genInf]
+
+prop_maximum :: Tree OrdA -> Property
+prop_maximum t = maximum t === maximum (toList t)
+
+prop_minimum :: Tree OrdA -> Property
+prop_minimum t = minimum t === minimum (toList t)
+
+prop_sum :: Tree OrdA -> Property
+prop_sum t = sum t === sum (toList t)
+
+prop_product :: Tree OrdA -> Property
+prop_product t = product t === product (toList t)
--- a/containers/src/Data/Tree.hs
+++ b/containers/src/Data/Tree.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE CPP #-}
 #if __GLASGOW_HASKELL__
@@ -52,7 +53,8 @@ module Data.Tree(
 
     ) where
 
-import Data.Foldable (toList)
+import Data.Foldable (fold, foldl', toList)
+import Data.Traversable (foldMapDefault)
 import Control.Monad (liftM)
 import Control.Monad.Fix (MonadFix (..), fix)
 import Data.Sequence (Seq, empty, singleton, (<|), (|>), fromList,
@@ -179,16 +181,59 @@ mfixTree f
                     [0..] children)
 
 instance Traversable Tree where
-    traverse f (Node x ts) = liftA2 Node (f x) (traverse (traverse f) ts)
+  traverse f = go
+    where go (Node x ts) = liftA2 Node (f x) (traverse go ts)
+  {-# INLINE traverse #-}
 
+-- | Folds in preorder
+
+-- See Note [Implemented Foldable Tree functions]
 instance Foldable Tree where
-    foldMap f (Node x ts) = f x `mappend` foldMap (foldMap f) ts
+    fold = foldMap id
+    {-# INLINABLE fold #-}
+
+    foldMap = foldMapDefault
+    {-# INLINE foldMap #-}
+
+    foldr f z = \t -> go t z  -- Use a lambda to allow inlining with two arguments
+      where
+        go (Node x ts) = f x . foldr (\t k -> go t . k) id ts
+        -- This is equivalent to the following simpler definition, but has been found to optimize
+        -- better in benchmarks:
+        -- go (Node x ts) z' = f x (foldr go z' ts)
+    {-# INLINE foldr #-}
+
+    foldl' f = go
+      where go !z (Node x ts) = foldl' go (f z x) ts
+    {-# INLINE foldl' #-}
+
+    foldr1 f = go id
+      where go k (Node x ts) = foldr (\n k' prev -> f prev (go k' n)) k ts x
+    {-# INLINE foldr1 #-}
+
+    foldl1 f (Node x ts) = foldl (foldl f) x ts
 
     null _ = False
     {-# INLINE null #-}
 
-    toList = flatten
-    {-# INLINE toList #-}
+    elem = any . (==)
+    {-# INLINABLE elem #-}
+
+    maximum = foldl1' max
+    {-# INLINABLE maximum #-}
+
+    minimum = foldl1' min
+    {-# INLINABLE minimum #-}
+
+    sum = foldl1' (+)
+    {-# INLINABLE sum #-}
+
+    product = foldl1' (*)
+    {-# INLINABLE product #-}
+
+foldl1' :: (a -> a -> a) -> Tree a -> a
+foldl1' f = \(Node x ts) -> foldl' (foldl' f) x ts
+{-# INLINE foldl1' #-}
 
 instance NFData a => NFData (Tree a) where
     rnf (Node x ts) = rnf x `seq` rnf ts
@@ -262,8 +307,7 @@ draw (Node x ts0) = lines x ++ drawSubTrees ts0
 --
 -- > flatten (Node 1 [Node 2 [], Node 3 []]) == [1,2,3]
 flatten :: Tree a -> [a]
-flatten t = squish t []
-  where squish (Node x ts) xs = x:Prelude.foldr squish xs ts
+flatten = toList
 
 -- | Returns the list of nodes at each level of the tree.
 --
@@ -415,3 +459,23 @@ unfoldForestQ f aQ = case viewl aQ of
     splitOnto as (_:bs) q = case viewr q of
         q' :> a -> splitOnto (a:as) bs q'
         EmptyR -> error "unfoldForestQ"
+
+--------------------------------------------------------------------------------
+
+-- Note [Implemented Foldable Tree functions]
+-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+--
+-- Implemented:
+--
+-- foldMap, foldr, foldl': Basic functions.
+-- fold, elem: Implemented same as the default definition, but INLINABLE to
+-- allow specialization.
+-- foldr1, foldl1, null, maximum, minimum: Implemented more efficiently than
+-- defaults since trees are non-empty.
+-- sum, product: Implemented as strict left folds. Defaults use the lazy foldMap
+-- before base 4.15.1.
+--
+-- Not implemented:
+--
+-- foldMap', toList, length: Defaults perform well.
+-- foldr', foldl: Unlikely to be used.
--- a/containers-tests/benchmarks/Tree.hs
+++ b/containers-tests/benchmarks/Tree.hs
@@ -1,0 +1,53 @@
+module Main where
+
+import Control.DeepSeq (NFData, rnf)
+import Control.Exception (evaluate)
+import Data.Coerce (coerce)
+import Data.Foldable (fold, foldl', toList)
+import Data.Monoid (All(..))
+import Test.Tasty.Bench (Benchmark, Benchmarkable, bench, bgroup, defaultMain, whnf, nf)
+import qualified Data.Tree as T
+
+main :: IO ()
+main = do
+  evaluate $ rnf ts `seq` rnf tsBool
+  defaultMain
+    [ bgroup "fold" $ forTs tsBool $ whnf fold . (coerce :: T.Tree Bool -> T.Tree All)
+    , bgroup "foldMap" $ forTs tsBool $ whnf (foldMap All)
+    , bgroup "foldr_1" $ forTs tsBool $ whnf (foldr (&&) True)
+    , bgroup "foldr_2" $ forTs ts $ whnf (length . foldr (:) [])
+    , bgroup "foldr_3" $ forTs ts $ whnf (\t -> foldr (\x k acc -> if acc < 0 then acc else k $! acc + x) id t 0)
+    , bgroup "foldl'" $ forTs ts $ whnf (foldl' (+) 0)
+    , bgroup "foldr1" $ forTs tsBool $ whnf (foldr1 (&&))
+    , bgroup "foldl1" $ forTs ts $ whnf (foldl1 (+))
+    , bgroup "toList" $ forTs ts $ nf toList
+    , bgroup "elem" $ forTs ts $ whnf (elem 0)
+    , bgroup "maximum" $ forTs ts $ whnf maximum
+    , bgroup "sum" $ forTs ts $ whnf sum
+    ]
+  where
+    ts = [binaryTree, lineTree] <*> [1000, 1000000]
+    tsBool = [t { getT = True <$ getT t } | t <- ts]
+
+forTs :: [Tree a] -> (T.Tree a -> Benchmarkable) -> [Benchmark]
+forTs ts f = [bench label (f t) | Tree label t <- ts]
+
+data Tree a = Tree
+  { getLabel :: String
+  , getT :: T.Tree a
+  }
+
+instance NFData a => NFData (Tree a) where
+  rnf (Tree label t) = rnf label `seq` rnf t
+
+binaryTree :: Int -> Tree Int
+binaryTree n = Tree label t
+  where
+    label = "bin,n=" ++ show n
+    t = T.unfoldTree (\x -> (x, takeWhile (<=n) [2*x, 2*x+1])) 1
+
+lineTree :: Int -> Tree Int
+lineTree n = Tree label t
+  where
+    label = "line,n=" ++ show n
+    t = T.unfoldTree (\x -> (x, [x+1 | x+1 <= n])) 1
--- a/containers-tests/containers-tests.cabal
+++ b/containers-tests/containers-tests.cabal
@@ -152,6 +152,14 @@ benchmark map-benchmarks
   main-is:        Map.hs
   ghc-options:    -O2
 
+benchmark tree-benchmarks
+  import: benchmark-deps
+  default-language: Haskell2010
+  type:           exitcode-stdio-1.0
+  hs-source-dirs: benchmarks
+  main-is:        Tree.hs
+  ghc-options:    -O2
+
 benchmark sequence-benchmarks
   import: benchmark-deps
   default-language: Haskell2010


treeowl:
95% less than? So only 5% of the earlier time? That sounds ... A bit suspicious. Is it right?

meooow25:
Unfortunately yes. Currently it seems to be ~20x slow for something like `sum`. I saw a similar difference when I reported the issue, and here's a comparison vs summing a list of the same size.
```
  sum Tree Int old: OK (0.21s)
    393  μs ±  23 μs
  sum [Int]:        OK (0.23s)
    13.9 μs ± 800 ns
```
```
  sum Tree Int new: OK (0.15s)
    18.2 μs ± 1.4 μs
```

treeowl:
It's apparent that the trees in the test aren't big enough. I don't love comparing microseconds. Could you possibly bump those sizes up some and see what things look like?

meooow25:
What size would you prefer? I chose `2^12` because that's what other container tests are using.

treeowl:
I don't know, really. Ideally we'd have a few sizes, but for now aim for half a second each, unless that makes the benchmarking take too long.

meooow25:
Half a second would need something _very_ large. I've updated it to `2^20`, a little over a million.
```
  fold:    OK (0.76s)
    8.45 ms ± 366 μs, 76% less than baseline
  foldMap: OK (0.40s)
    8.89 ms ± 683 μs, 73% less than baseline
  foldr:   OK (0.42s)
    10.7 ms ± 753 μs, 94% less than baseline
  foldl':  OK (0.42s)
    10.5 ms ± 949 μs, 95% less than baseline
  foldr1:  OK (0.26s)
    10.0 ms ± 989 μs, 95% less than baseline
  foldl1:  OK (0.42s)
    10.5 ms ± 748 μs, 96% less than baseline
  toList:  OK (0.35s)
    30.0 ms ± 1.8 ms,  9% less than baseline
  elem:    OK (0.43s)
    10.3 ms ± 701 μs, 28% less than baseline
  maximum: OK (0.42s)
    10.5 ms ± 788 μs, 96% less than baseline
  sum:     OK (0.26s)
    10.5 ms ± 725 μs, 96% less than baseline
```

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,43 @@ mfixTree f
...
-    foldMap f (Node x ts) = f x `mappend` foldMap (foldMap f) ts
+    fold = foldMap id
+    {-# INLINABLE fold #-}
+
+    foldMap f = go
+      where go (Node x ts) = f x `mappend` foldMap go ts
+    {-# INLINE foldMap #-}
+
+    foldr f z = \t -> go t z
+      where go (Node x ts) = f x . foldr ((.) . go) id ts
```
treeowl, line 195: Is there a reason not to tuck `z` under the lambda?

I imagine you pass `z` through the recursion because that performs better than doing the obvious thing. Have you verified that it's better? If so, a note explaining that would be helpful.

meooow25: I assume you mean "under the `go`"? Yes I had tried a few variations and found it performs well.
Added a comment.

Looking at the core and it does seem to generate better output. I wouldn't be able to explain why though.
For the record, on the `(&&)` benchmark
```hs
go (Node x ts) = f x . foldr ((.) . go) id ts
```
generates core like
```
Rec {
-- RHS size: {terms: 20, types: 19, coercions: 0, joins: 0/0}
go1 :: [Tree Bool] -> Bool -> Bool
[GblId, Arity=2, Str=<1L><ML>, Unf=OtherCon []]
go1
  = \ (ds :: [Tree Bool]) (eta :: Bool) ->
      case ds of {
        [] -> eta;
        : y ys ->
          case y of { Node ww1 ww2 ->
          case ww1 of {
            False -> False;
            True -> go1 ww2 (go1 ys eta)
          }
          }
      }
end Rec }
```
which is ideal, and
```hs
go (Node x ts) z' = f x (foldr go z' ts)
```
generates
```
Rec {
-- RHS size: {terms: 25, types: 27, coercions: 0, joins: 0/1}
$wgo3 :: Bool -> [Tree Bool] -> Bool -> Bool
[GblId, Arity=3, Str=<1L><ML><L>, Unf=OtherCon []]
$wgo3
  = \ (ww :: Bool) (ww1 :: [Tree Bool]) (w :: Bool) ->
      case ww of {
        False -> False;
        True ->
          letrec {
            go1 [Occ=LoopBreaker, Dmd=SCS(L)] :: [Tree Bool] -> Bool
            [LclId, Arity=1, Str=<1L>, Unf=OtherCon []]
            go1
              = \ (ds :: [Tree Bool]) ->
                  case ds of {
                    [] -> w;
                    : y ys -> case y of { Node ww3 ww4 -> $wgo3 ww3 ww4 (go1 ys) }
                  }; } in
          go1 ww1
      }
end Rec }
```

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,46 @@ mfixTree f
...
+    {-# INLINABLE fold #-}
+
+    foldMap f = go
+      where go (Node x ts) = f x `mappend` foldMap go ts
+    {-# INLINE foldMap #-}
+
+    foldr f z = \t -> go t z  -- Use a lambda to allow inlining with two arguments
+      where
+        go (Node x ts) = f x . foldr ((.) . go) id ts
+        -- This has been found to optimize better than the more obvious definition of
```
phadej, line 197: this is a unjustified claim. Microbenchmarks are not enough to establish that. Given that `foldr` is marked `INLINE`, it's optimized at each use case

I'd go for simpler version if the difference in benchmarks small, as GHC optimizer is unrelieable, and perfectly should be tuned to make "normal" code run smarter. I.e. in perfect future, the simple version would optimize as well.

If the difference is big, than it's should be reported to GHC upstream.

I'm also not a fan of having `INLINE` for "performance". It's not sustainable, causes code growth (makes microbenchmarks look better, but real code worse). `INLINEABLE` should be enough for specialization.

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,46 @@ mfixTree f
...
 
     null _ = False
     {-# INLINE null #-}
 
-    toList = flatten
-    {-# INLINE toList #-}
+    elem = any . (==)
+    {-# INLINABLE elem #-}
+
+    maximum = foldl1' max
```
phadej, line 216: You should also implement `sum` and `product`, as you on this.

And `foldMap'` for `base >=4.13`.

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,46 @@ mfixTree f
...
+    {-# INLINABLE maximum #-}
+
+    minimum = foldl1' min
+    {-# INLINABLE minimum #-}
+
+foldl1' :: (a -> a -> a) -> Tree a -> a
+foldl1' f = \(Node x ts) -> foldl' (foldl' f) x ts
+{-# INLINE foldl1' #-}
```
phadej, line 224: GHC will barf on `INLINE` on recursive functions. I.e. it does nothing. `INLINEABLE` for specialization is enough.

containers/src/Data/Tree.hs
phadej: Forgot to mention: `foldr` doesn't need to be "fast", it meant to be lazy (e.g. for lists), and can be for `Tree` as well. 

Benchmarking something like `head (toList t)` is a good benchmark for `foldr`. It should be independent of `t` size.

meooow25: > this is a unjustified claim. Microbenchmarks are not enough to establish that. Given that foldr is marked INLINE, it's optimized at each use case

Fair point, but we have to pick one implementation, and benchmarks help us decide. If you would prefer the comment said something else I'm open to suggestions.

> I'd go for simpler version if the difference in benchmarks small

Current
```
  foldr: OK (0.36s)
    11.1 ms ± 840 μs
```
Simpler
```
  foldr: OK (0.26s)
    18.2 ms ± 1.5 ms
```

> If the difference is big, than it's should be reported to GHC upstream.

That's reasonable, I can open an issue.

> I'm also not a fan of having INLINE for "performance".

It seems necessary, `INLINABLE` doesn't always work as well. Every `foldr` in containers (`Set`, `Map`, `IntSet`, `IntMap`, `FingerTree` for `Sequence`) is `INLINE` for performance.

> Forgot to mention: foldr doesn't need to be "fast", it meant to be lazy

Why not both?

> Benchmarking something like head (toList t) is a good benchmark for foldr. 

`head (toList t)` would finish instantaneously. I don't see the point of that. But I'm happy to add/replace foldr benchmarks with other use cases. There is already `toList` and `&&` is the best I could think of for something reducing to a small value.

containers/src/Data/Tree.hs
meooow25: As I mentioned in the description `sum`, `product` and `foldMap'` are working well by default. The only thing we gain by implementing these is saving one operation with the initial value. Which I can implement, sure, but is that what you're asking for?

containers/src/Data/Tree.hs
meooow25: This is not a recursive function.

containers/src/Data/Tree.hs
phadej: description is not visible in the source nor version control. Why `elem` is expanded, but `foldMap'` isn't? Write the note.

I also doubt that `sum` works well on old GHCs, before it was changed to use `foldMap'`.

containers/src/Data/Tree.hs
phadej: Ah indeed!

containers/src/Data/Tree.hs
phadej: What's the difference between current and simpler implementation if you don't INLINE them.

> It seems necessary, INLINABLE doesn't always work as well. Every foldr in containers (Set, Map, IntSet, IntMap, FingerTree for Sequence) is INLINE for performance.

And that is bad. Don't make things worse.

> head (toList t) would finish instantaneously.

It's a correctness test. `take n (toList infinitelySwawningTree)` should always work.
e.g. `let t = Node 'x' (repeat t)` or just `t = Node 'x' [t]`.

treeowl: @phadej, inclining a fold for a simple structure generally shouldn't increase code size much if at all. What's the problem you see with it?

treeowl: I tend to want two benchmarks for `foldr`, to reflect two use patterns:

1. `length . foldr (:) []`, for the lazy production scenario.
2. Some strict left fold with abort for other common uses. It's too early in the morning for me to come up with a nice one.

phadej: > inlining a fold for a simple structure generally shouldn't increase code size much if at all. What's the problem you see with it?

The little streams make the big rivers.

YMMV.

containers/src/Data/Tree.hs
meooow25: Done, added a note and implemented `sum` and `product`.

containers/src/Data/Tree.hs
meooow25: Added the `length . foldr (:) []` benchmark.

> What's the difference between current and simpler implementation if you don't INLINE them.

|`foldr (&&) True`|`INLINE`|`INLINABLE`|Nothing|
|---|---|---|---|
|Current|11.0 ms ± 546 μs|23.3 ms ± 1.6 ms|23.2 ms ± 1.5 ms|
|Simple|18.2 ms ± 1.4 ms|18.3 ms ± 1.5 ms|30.0 ms ± 1.4 ms|

|`length . foldr (:) []`|`INLINE`|`INLINABLE`|Nothing|
|---|---|---|---|
|Current|17.5 ms ± 1.7 ms|21.5 ms ± 1.6 ms|21.4 ms ± 1.8 ms|
|Simple|31.3 ms ± 1.6 ms|30.0 ms ± 2.4 ms|35.6 ms ± 2.8 ms|

I'm strongly in favor of keeping the current implementation.

> It's a correctness test.

Well then that's a test, not a benchmark. Added.

meooow25: @treeowl what do you think?
And let me know if want me to add an example of 2.

treeowl: I definitely want an example of a left fold with abort. Something like

```haskell
foldr go stop t (0 :: Int)
  where
    go x r acc
      | acc > some_constant = acc
      | otherwise = r (acc + x)
```

I think that's an important `foldr` pattern, and we want to make sure it's optimized reasonably well.

meooow25: Added.

meooow25:
Added benchmarks using a line tree as an example of a deep tree, since a binary tree is rather shallow. Also added a smaller size of 1000.

And I opened a GHC issue about the fold function: https://gitlab.haskell.org/ghc/ghc/-/issues/22659

<details>
<summary>New benchmarks</summary>
<pre>
  fold
    bin,n=1000:     OK (0.96s)
      4.20 μs ± 347 ns
    bin,n=1000000:  OK (0.58s)
      8.86 ms ± 820 μs
    line,n=1000:    OK (1.70s)
      5.30 μs ± 442 ns
    line,n=1000000: OK (0.56s)
      14.8 ms ± 802 μs
  foldMap
    bin,n=1000:     OK (1.74s)
      4.58 μs ± 338 ns
    bin,n=1000000:  OK (0.45s)
      8.93 ms ± 869 μs
    line,n=1000:    OK (1.77s)
      5.79 μs ± 367 ns
    line,n=1000000: OK (0.58s)
      15.2 ms ± 1.1 ms
  foldr_1
    bin,n=1000:     OK (1.75s)
      5.51 μs ± 331 ns
    bin,n=1000000:  OK (0.59s)
      10.7 ms ± 723 μs
    line,n=1000:    OK (1.58s)
      7.34 μs ± 667 ns
    line,n=1000000: OK (3.56s)
      46.1 ms ± 3.3 ms
  foldr_2
    bin,n=1000:     OK (1.59s)
      8.42 μs ± 716 ns
    bin,n=1000000:  OK (0.69s)
      17.0 ms ± 1.2 ms
    line,n=1000:    OK (1.42s)
      9.31 μs ± 757 ns
    line,n=1000000: OK (0.60s)
      34.6 ms ± 2.5 ms
  foldr_3
    bin,n=1000:     OK (1.63s)
      7.88 μs ± 674 ns
    bin,n=1000000:  OK (0.66s)
      15.0 ms ± 904 μs
    line,n=1000:    OK (1.78s)
      7.25 μs ± 365 ns
    line,n=1000000: OK (0.43s)
      28.5 ms ± 1.8 ms
  foldl'
    bin,n=1000:     OK (1.09s)
      4.43 μs ± 335 ns
    bin,n=1000000:  OK (0.61s)
      10.5 ms ± 753 μs
    line,n=1000:    OK (1.12s)
      5.81 μs ± 439 ns
    line,n=1000000: OK (0.78s)
      20.8 ms ± 1.1 ms
  foldr1
    bin,n=1000:     OK (2.08s)
      5.54 μs ± 207 ns
    bin,n=1000000:  OK (0.60s)
      10.1 ms ± 989 μs
    line,n=1000:    OK (1.78s)
      6.69 μs ± 443 ns
    line,n=1000000: OK (0.37s)
      19.6 ms ± 1.4 ms
  foldl1
    bin,n=1000:     OK (1.73s)
      4.35 μs ± 377 ns
    bin,n=1000000:  OK (0.87s)
      10.3 ms ± 419 μs
    line,n=1000:    OK (1.14s)
      6.27 μs ± 609 ns
    line,n=1000000: OK (2.51s)
      30.6 ms ± 1.6 ms
  toList
    bin,n=1000:     OK (1.61s)
      9.13 μs ± 761 ns
    bin,n=1000000:  OK (0.54s)
      29.4 ms ± 1.6 ms
    line,n=1000:    OK (1.61s)
      10.7 μs ± 814 ns
    line,n=1000000: OK (0.38s)
      49.4 ms ± 4.8 ms
  elem
    bin,n=1000:     OK (2.00s)
      4.48 μs ± 288 ns
    bin,n=1000000:  OK (0.88s)
      10.5 ms ± 440 μs
    line,n=1000:    OK (1.02s)
      5.50 μs ± 383 ns
    line,n=1000000: OK (0.37s)
      20.8 ms ± 1.7 ms
  maximum
    bin,n=1000:     OK (1.74s)
      4.52 μs ± 387 ns
    bin,n=1000000:  OK (0.37s)
      10.5 ms ± 974 μs
    line,n=1000:    OK (1.14s)
      6.16 μs ± 342 ns
    line,n=1000000: OK (0.79s)
      20.9 ms ± 722 μs
  sum
    bin,n=1000:     OK (1.99s)
      4.47 μs ± 285 ns
    bin,n=1000000:  OK (0.89s)
      10.3 ms ± 392 μs
    line,n=1000:    OK (1.15s)
      6.36 μs ± 388 ns
    line,n=1000000: OK (0.79s)
      21.0 ms ± 1.7 ms
</pre>
</details>

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,55 @@ mfixTree f
...
+    fold = foldMap id
+    {-# INLINABLE fold #-}
+
+    foldMap f = go
+      where go (Node x ts) = f x `mappend` foldMap go ts
+    {-# INLINE foldMap #-}
+
+    foldr f z = \t -> go t z  -- Use a lambda to allow inlining with two arguments
+      where
+        go (Node x ts) = f x . foldr ((.) . go) id ts
```
phadej, line 198: In https://gitlab.haskell.org/ghc/ghc/-/issues/22659#note_469510 SPJ identifies why this is faster. And sebastian suggest that this variant may be slower for deep trees (i.e. something like `Node 1 $ singleton $ Node 2 $ singleton $ ...`)

I suggest:
- Add benchmarks for "simple" foldr as well, so the difference can be checked later as well
- Add benchmark for deep trees
- Rewrite this using two mutually recursive functions as SPJ shows. That implementation is less magical, relies less on optimizer, and is easy to explain why it behaves better.

meooow25: > sebastian suggest that this variant may be slower for deep trees

That is not what Sebastian is saying. If I understand the comment correctly, Sebastian says that if we had two mutually recursive functions after optimization, we would allocate more. The current code optimizes to a single function.

> Add benchmarks for "simple" foldr as well, so the difference can be checked later as well

I don't see the point of this. Anyone running benchmarks would have to filter out the noise of a benchmark that is not live code. If anyone wants to test the simple version that's easy, they just need to comment out this line and uncomment the definition below.

> Add benchmark for deep trees

I have, please read the comment above.

> Rewrite this using two mutually recursive functions as SPJ shows. That implementation is less magical, relies less on the optimizer, and is easy to explain why it behaves better.

The second function out of SPJ's two mutually recursive functions manually case matches on the list and recurses. We don't need to write code like that when we have folds to do the job for us. And if we use foldr, that is the same as the simple definition, only broken into two, and it optimizes to the same inefficient code.

I honestly don't understand your dislike for this implementation.

phadej: At the very least expand

```haskell
(.) . go
```

it's a pointless point-free form.

phadej: > I don't see the point of this. Anyone running benchmarks would have to filter out the noise of a benchmark that is not live code. If anyone wants to test the simple version that's easy, they just need to comment out this line and uncomment the definition below.

I disagree about "easy". But as I'm not a maintainer, I cannot require you to add it either.

meooow25: > At the very least expand `(.) . go`

Sure thing, done.

phadej: Ok now I undertstand what happens.

```haskell
foldr f z = \(Node x ts) -> f x (foldrForest f ts z)

foldrForest :: (a -> b -> b) -> [Tree a] -> b -> b
foldrForest f = go where 
  go []               z = z
  go (Node x xs : ys) z = f x (go xs (go ys z))
{-# INLINE foldrForest #-}
```

I find this easier to understand that one written using list's `foldr`. YMMV

meooow25: Yes, your function seems to end up in the same form as the current implementation. But I do not find the current one hard to understand. Of course, this is subjective. When something can be written as a fold, I would prefer to write it as a fold instead of explicit recursion unless there is a strong reason to do otherwise.

@treeowl do you have a preference here?
Otherwise, if there's nothing else to add, can we close this?

treeowl:
Sorry, I got distracted with some lens stuff. I like folds, if they're clear and fast. I feel similarly about explicit recursion. Whatever's nicest in context.

meooow25:
I'm happy with the current state of things. If you agree, we can merge it.

meooow25:
Sorry, I feel like I'm missing something. What's the hold up here? Please tell me clearly.

treeowl:
I need to make the time for a final review. I will try to do that tomorrow.

meooow25:
That's understandable, thank you.

meooow25:
Hi, did you get time to review it?

treeowl:
It will happen tonight without fail!

containers-tests/tests/tree-properties.hs
```diff
@@ -101,3 +116,23 @@ prop_monadFix_ls val ta ti =
...
+prop_foldr1 :: Tree Int -> Property
+prop_foldr1 t = foldr1 (++) (fmap (:[]) t) === toList t
+
+prop_foldl1 :: Tree Int -> Property
+prop_foldl1 t = foldl1 (++) (fmap (:[]) t) === toList t
+
+prop_foldr_infinite :: Int -> Property
+prop_foldr_infinite n = take n (foldr (:) [] t) === [1..n]
+  where
+    t = Node 1 [Node x [] | x <- [2..]]
```
treeowl, line 138: This test is really just about termination, right? Anything else it has to say will be captured by finite tests. Wouldn't it be better to use something like this to get more infinite tree shapes?

```haskell
prop_foldr_infinite n = forAll genny (\genny -> length (take n (foldr (:) [] t)) === n)
  where
    genny :: Gen (Tree ())
    genny = Node () <$> oneof [listOf genny, infiniteListOf genny]
```

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,55 @@ mfixTree f
...
+    {-# INLINABLE maximum #-}
+
+    minimum = foldl1' min
+    {-# INLINABLE minimum #-}
+
+    sum = foldl1' (+)
+    {-# INLINABLE sum #-}
+
+    product = foldl1' (*)
+    {-# INLINABLE product #-}
```
treeowl, line 229: I don't see tests for any of these methods. Could you add a sum test and a minimum test?

containers/src/Data/Tree.hs
```diff
@@ -415,3 +456,23 @@ unfoldForestQ f aQ = case viewl aQ of
...
+-- fold, elem: Implemented same as the default definition, but INLINABLE to
+-- allow specialization.
+-- foldr1, foldl1, null, maximum, minimum: Implemented more efficiently than
+-- defaults since trees are non-empty.
+-- sum, product: Implemented as strict left folds. Defaults use the lazy foldMap
+-- before base 4.15.1.
+--
+-- Not implemented:
+--
+-- foldMap', toList, length: Defaults perform well.
```
treeowl, line 477: I don't understand what makes `fold` and `elem` different from `foldMap'` in this regard. Are there issues specializing the first two but not the last? If so, do you have any guess why?

treeowl (changes requested):
I'm sorry for the long delay. I've marked up some questions and requests.

treeowl:
Did you see my review?

treeowl:
For `foldl1` and `foldr1`, I would recommend testing with the operation of a free magma.

```haskell
data Magma a
  = Inj a
  | Magma a :* Magma a
  deriving (Eq, Show)
```

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,55 @@ mfixTree f
...
+
+-- See Note [Implemented Foldable Tree functions]
 instance Foldable Tree where
-    foldMap f (Node x ts) = f x `mappend` foldMap (foldMap f) ts
+    fold = foldMap id
+    {-# INLINABLE fold #-}
+
+    foldMap f = go
+      where go (Node x ts) = f x `mappend` foldMap go ts
+    {-# INLINE foldMap #-}
```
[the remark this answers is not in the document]
meooow25: Looks like it works, done.

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,55 @@ mfixTree f
...
+
+    foldr f z = \t -> go t z  -- Use a lambda to allow inlining with two arguments
+      where
+        go (Node x ts) = f x . foldr (\t k -> go t . k) id ts
+        -- This is equivalent to the following simpler definition, but has been found to optimize
+        -- better in benchmarks. See #880 for a discussion on this.
```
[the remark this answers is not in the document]
meooow25: In that case I think we can skip it. We provide the simple function below and the discussion is a bit noisy and not too important.

containers/src/Data/Tree.hs
meooow25: The default definition `foldMap'` is perhaps small enough for GHC to inline without any help. But I don't know for sure.

containers-tests/tests/tree-properties.hs
```diff
@@ -101,3 +116,23 @@ prop_monadFix_ls val ta ti =
...
+prop_foldMap t = foldMap (:[]) t === toList t
+
+prop_foldl' :: Tree Int -> Property
+prop_foldl' t = foldl' (flip (:)) [] t === reverse (toList t)
+
+prop_foldr1 :: Tree Int -> Property
+prop_foldr1 t = foldr1 (++) (fmap (:[]) t) === toList t
+
+prop_foldl1 :: Tree Int -> Property
+prop_foldl1 t = foldl1 (++) (fmap (:[]) t) === toList t
```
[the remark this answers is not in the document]
meooow25: Done.

containers-tests/tests/tree-properties.hs
```diff
@@ -101,3 +116,23 @@ prop_monadFix_ls val ta ti =
     f :: (Int -> Int) -> Int -> Tree (Int -> Int)
     f q y = let t = apply ti y
             in fmap (\w -> fact w q) t
+
+prop_toList :: Tree Int -> Property
+prop_toList t = toList t === foldr (:) [] t
+
+prop_foldMap :: Tree Int -> Property
+prop_foldMap t = foldMap (:[]) t === toList t
```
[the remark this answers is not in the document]
meooow25: Now that foldMap is defined as foldMapDefault, might be a little odd to test this?

containers-tests/tests/tree-properties.hs
meooow25: Sure, done. This generator can generate a finite tree though (`listOf genny` can be null), so I added an infinite of these at the top level.

containers/src/Data/Tree.hs
meooow25: Added for all four.

meooow25:
> Did you see my review?

Hi yes, was going through your comments.

> For `foldl1` and `foldr1`, I would recommend testing with the operation of a free magma.

Thanks, I was thinking the same but I wasn't sure what to name it. And in fact `foldr1` did give different results for a non-associative function, so thanks for catching this. Fixed now.

containers/src/Data/Tree.hs
```diff
@@ -181,14 +182,55 @@ mfixTree f
...
 instance Foldable Tree where
-    foldMap f (Node x ts) = f x `mappend` foldMap (foldMap f) ts
+    fold = foldMap id
+    {-# INLINABLE fold #-}
+
+    foldMap f = go
+      where go (Node x ts) = f x `mappend` foldMap go ts
+    {-# INLINE foldMap #-}
+
+    foldr f z = \t -> go t z  -- Use a lambda to allow inlining with two arguments
```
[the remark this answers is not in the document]
meooow25: Right, I don't think it would be bad to do that, but all `foldrs` in containers are defined as `foldr f z` (even when z can be eta-reduced, not sure why), so I've kept it consistent.
I also noticed this brought up in [GHC !5259](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5259) for `foldl'`. But it seems that it was not definitively addressed, and in the end they went with `foldl' f z0`.

containers/src/Data/Tree.hs
treeowl: Bleh. GHC inlining pragmas are such a dark art. I hate it.

containers-tests/tests/tree-properties.hs
treeowl: Generating a finite tree sometimes is fine. We're going to be generating a *lot* of trees. I'd keep it simple.

treeowl: Oh, but I forgot that it should therefore be `length ... <= n`!

containers/src/Data/Tree.hs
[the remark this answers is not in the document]
treeowl: OK. There might be/have been a reason. The secrets of the GHC inliner....

containers-tests/tests/tree-properties.hs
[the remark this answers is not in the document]
treeowl: I know it's odd, but that'll catch a mistake if someone later decides to go back to implementing it manually.

containers-tests/tests/tree-properties.hs
meooow25: Oh that's what you meant, done.

containers-tests/tests/tree-properties.hs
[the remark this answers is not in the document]
meooow25: Okay, done.

containers-tests/tests/tree-properties.hs
```diff
@@ -101,3 +130,39 @@ prop_monadFix_ls val ta ti =
     f :: (Int -> Int) -> Int -> Tree (Int -> Int)
     f q y = let t = apply ti y
             in fmap (\w -> fact w q) t
+
+prop_toList :: Tree Int -> Property
```
treeowl, line 134: I'm sorry I didn't notice this earlier, but in this test suite (and most of the rest in the package) we tend to use `Test.QuickCheck.Poly` where applicable. So for this test we should have

```haskell
prop_toList :: Tree A -> Property
```

The same goes for the rest. Yes, it's a bit less efficient (`A`, `B`, `C`, etc., are actually `Integer`), but clear communication is more important than constant factors for these tests. For `prop_maximum`, ..., `prop_product`, you'll need to use `OrdA`, etc., instead (they have `Ord` and `Num` instances).

meooow25: Ah makes sense, done.

treeowl (approved):
Whew! Let's do it.

treeowl:
Could you please squash this into one or more logical commits?

meooow25:
Squashed them all!

PR #881, opened 2022-12-11
Implemement foldr, foldl', sum and product in Foldable Tree

phadej:
Addresses #878

With this patch on my machine:

```
All
  foldMap'Tree Sum: OK (0.31s)
    18.8 ms ± 1.9 ms
  sum:              OK (1.42s)
    18.0 ms ± 761 μs
```

This patch "changes" the strictness of `sum` and `product`, but it has already been *different* depending on which GHC/base version `containers` is compiled with. So this change makes it consistent.


phadej:
Ah, there's already #880

PR #889, opened 2023-01-07
Add/correct some time complexities

meooow25:
Added:
* Set: take, drop, powerSet, disjointUnion
* Map: take, drop, splitMember

Corrected:
* Set: cartesianProduct, showTree, showTreeWith
* Map: showTree, showTreeWith
* IntSet: lookupLT, lookupGT, lookupLE, lookupGE, showTree, showTreeWith
* IntMap: lookupLT, lookupGT, lookupLE, lookupGE, alterF, showTree, showTreeWith

Brief explanations for the non-obvious complexities:

* Set.powerSet (added)  
  The power set is calculated in n steps, where at each step the most expensive part is inserting a minimum element into every set in the power set so far. For the ith step this takes $O(2^i \log i)$ time. Overall it's $O(2^n \log n)$.
* Set.cartesianProduct ( $O(nm)$, not conjectured any more)  
  The product is obtained by mapping every node in a set of size $O(n)$ to a set of size $O(m)$, then merging them all. Building the sets clearly takes $O(nm)$. Merging two sets takes logarithmic time. When merging up a binary tree of height $O(\log n)$ it takes $O(n \log m)$, following a similar reasoning as [building a binary heap in O(n)](https://en.wikipedia.org/wiki/Binary_heap#Building_a_heap). So it's $O(nm)$ overall.
* Set.showTree, Set.showTreeWith, Map.showTree, Map.showTreeWith (changed from $O(n)$ to $O(n \log n)$ )  
  In the output string, the line containing a node has length which is $O(\text{depth of the node})$. A Set/Map has depth $O(\log n)$, so the sum of depths of all nodes is bounded by $O(n \log n)$. Nievergelt and Reingold provide a concrete $O(n \log n)$ bound in their paper linked in the docs (Theorem 2).
* IntSet.showTree, IntSet.showTreeWith, IntMap.showTree, IntMap.showTreeWith (changed from $O(n)$ to $O(n \min (n,W))$ )  
  Similar to Set/Map. The depth of an IntSet/IntMap is $O(\min (n,W))$ so the complexity is $O(n \min (n,W))$.

### DIFF
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -1823,6 +1823,17 @@ splitRoot orig =
 -- @
 --
 -- @since 0.5.11
+
+-- Proof of complexity: step executes n times. At the ith step,
+-- "insertMin x `mapMonotonic` pxs" takes O(2^i log i) time since pxs has size
+-- 2^i - 1 and we insertMin into its elements which are sets of size <= i.
+-- "insertMin (singleton x)" and "`glue` pxs" are cheaper operations that both
+-- take O(i) time. Over n steps, we have a total cost of
+--
+--   O(\sum_{i=1}^{n-1} 2^i log i)
+-- = O(log n * \sum_{i=1}^{n-1} 2^i)
+-- = O(2^n log n)
+
 powerSet :: Set a -> Set (Set a)
 powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
   step x pxs = insertMin (singleton x) (insertMin x `mapMonotonic` pxs) `glue` pxs
@@ -1842,11 +1853,7 @@ powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
 --
 -- @since 0.5.11
 cartesianProduct :: Set a -> Set b -> Set (a, b)
--- I don't know for sure if this implementation (slightly modified from one
--- that Edward Kmett hacked together) is optimal. TODO: try to prove or
--- refute it.
---
--- We could definitely get big-O optimal (O(m * n)) in a rather simple way:
+-- The obvious big-O optimal (O(nm)) implementation would be
 --
 --   cartesianProduct _as Tip = Tip
 --   cartesianProduct as bs = fromDistinctAscList
@@ -1855,8 +1862,31 @@ cartesianProduct :: Set a -> Set b -> Set (a, b)
 -- Unfortunately, this is much slower in practice, at least when the sets are
 -- constructed from ascending lists. I tried doing the same thing using a
 -- known-length (perfect balancing) variant of fromDistinctAscList, but it
--- still didn't come close to the performance of Kmett's version in my very
--- informal tests.
+-- still didn't come close to the performance of the implementation we use in my
+-- very informal tests.
+--
+-- The implementation we use (slightly modified from one that Edward Kmett
+-- hacked together) is also optimal but performs better in practice. We map
+-- each element a in as to a set made up of (a,b) for every element b in bs,
+-- taking O(nm) overall. Then we merge these sets up the tree of as, which takes
+-- O(n log m). A brief sketch of proof for the latter:
+--
+-- Consider all nodes in the tree at the same distance from the root to be at
+-- the same "level". The nodes farthest from the root are at level 0, with
+-- levels increasing by 1 towards the root. Being a balanced tree, there are
+-- O(n/2^i) nodes at level i. At every node at level i, we merge the merged left
+-- set, current set, and merged right set into a set of size O(2^i*m) in
+-- O(log (2^i*m)) = O(i + log m) time. Over all levels, we do a total work of
+--
+--   O(\sum_{i=0}^{root_level} n * (i + log m) / 2^i)
+-- = O(  \sum_{i=0}^{root_level} n * i / 2^i
+--     + \sum_{i=0}^{root_level} n * log m / 2^i)
+-- = O(  n * \sum_{i=0}^{root_level} i/2^i
+--     + n * log m * \sum_{i=0}^{root_level} 1/2^i)
+-- = O(  n * \sum_{i=0}^{inf} i/2^i
+--     + n * log m * \sum_{i=0}^{inf} 1/2^i)
+--
+-- The sum terms converge, and we get O(n log m).
 
 -- When the second argument has at most one element, we can be a little
 -- clever.
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -1407,7 +1407,7 @@ deleteAt !i t =
       where
         sizeL = size l
 
--- | Take a given number of elements in order, beginning
+-- | \(O(\log n)\). Take a given number of elements in order, beginning
 -- with the smallest ones.
 --
 -- @
@@ -1428,7 +1428,7 @@ take i0 m0 = go i0 m0
         EQ -> l
       where sizeL = size l
 
--- | Drop a given number of elements in order, beginning
+-- | \(O(\log n)\). Drop a given number of elements in order, beginning
 -- with the smallest ones.
 --
 -- @
@@ -1809,7 +1809,7 @@ splitRoot orig =
 {-# INLINE splitRoot #-}
 
 
--- | Calculate the power set of a set: the set of all its subsets.
+-- | \(O(2^n \log n)\). Calculate the power set of a set: the set of all its subsets.
 --
 -- @
 -- t ``member`` powerSet s == t ``isSubsetOf`` s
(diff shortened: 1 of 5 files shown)


containers/src/Data/Set/Internal.hs
```diff
@@ -1827,7 +1827,7 @@ powerSet :: Set a -> Set (Set a)
 powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
   step x pxs = insertMin (singleton x) (insertMin x `mapMonotonic` pxs) `glue` pxs
 
--- | \(O(mn)\) (conjectured). Calculate the Cartesian product of two sets.
+-- | \(O(nm)\). Calculate the Cartesian product of two sets.
```
meooow25, line 1830: Forgot to mention, the comment below about uncertainty of the implementation being optimal needs to adjusted. Let me know if we want to remove it or keep some part of it.

treeowl: I think the part about conjecture should be replaced by your proof. The simple alternative proving to be slow seems like valuable information for someone in the future who wonders why we do the complicated thing, so it should stay.

meooow25: Done. I didn't want to be too handwavy, so I added a little more detail. I hope it's not too verbose now. Sadly, math too doesn't look very nice in a comment.

containers/src/Data/Set/Internal.hs
```diff
@@ -1809,7 +1809,7 @@ splitRoot orig =
 {-# INLINE splitRoot #-}
 
 
--- | Calculate the power set of a set: the set of all its subsets.
+-- | \(O(2^n \log n)\). Calculate the power set of a set: the set of all its subsets.
```
treeowl, line 1812: Would you please include a proof?

containers/src/Data/Set/Internal.hs
```diff
@@ -1842,10 +1842,6 @@ powerSet xs0 = insertMin empty (foldr' step Tip xs0) where
 --
 -- @since 0.5.11
 cartesianProduct :: Set a -> Set b -> Set (a, b)
--- I don't know for sure if this implementation (slightly modified from one
--- that Edward Kmett hacked together) is optimal. TODO: try to prove or
--- refute it.
---
 -- We could definitely get big-O optimal (O(m * n)) in a rather simple way:
```
treeowl, line 1845:
```suggestion
-- The obvious big-O optimal (O(m * n)) implementation would be
```

containers/src/Data/Set/Internal.hs
```diff
@@ -1855,8 +1851,25 @@ cartesianProduct :: Set a -> Set b -> Set (a, b)
...
+--
+-- The implementation we use (slightly modified from one that Edward Kmett
+-- hacked together) is also optimal but performs better in practice. We map
+-- each element a in as to a set made up of (a,b) for every element b in bs,
+-- taking O(nm) overall. Then we merge these sets up the tree of as, which takes
+-- O(n log m). A brief sketch of proof for the latter:
+--
+-- There are O(n/2^i) nodes at level i of the tree. At a node at level i, we
+-- merge the merged left set, current set, and merged right set into a set of
+-- size O(2^i*m). This is done in a logarithmic time of O(i + log m). Over all
```
treeowl, line 1865: Where does `i + log m` come from?

containers/src/Data/Set/Internal.hs
```diff
@@ -1897,14 +1910,14 @@ disjointUnion as bs = merge (mapMonotonic Left as) (mapMonotonic Right bs)
 {--------------------------------------------------------------------
   Debugging
 --------------------------------------------------------------------}
--- | \(O(n)\). Show the tree that implements the set. The tree is shown
+-- | \(O(n \log n)\). Show the tree that implements the set. The tree is shown
```
treeowl, line 1913: Nice catch. Those spaces and bars add up!

treeowl (changes requested):

containers/src/Data/Set/Internal.hs
meooow25: Sure, doesn't hurt to have it.

containers/src/Data/Set/Internal.hs
meooow25: $\log_2(2^i \cdot m) = i  + \log_2 m$

treeowl: Could you please specify exactly what you mean by "level $i$"? That seems to refer to the distance from a leaf node in some fashion, but readers shouldn't have to calculate things to figure that out.

containers/src/Data/Set/Internal.hs
```diff
@@ -1855,8 +1862,25 @@ cartesianProduct :: Set a -> Set b -> Set (a, b)
...
+-- O(n log m). A brief sketch of proof for the latter:
+--
+-- There are O(n/2^i) nodes at level i of the tree. At a node at level i, we
+-- merge the merged left set, current set, and merged right set into a set of
+-- size O(2^i*m). This is done in a logarithmic time of O(i + log m). Over all
+-- levels, we do a total work of
+--
+--    O(\sum_{i=1}^{height} n * (i + log m) / 2^i)
+--  = O(  n * \sum_{i=1}^{inf} i/2^i
+--      + n * log m * \sum_{i=1}^{inf} 1/2^i)
```
treeowl, line 1881: Could you add an extra step? Switching the limits of summation and distributing are unrelated.

treeowl (changes requested):

containers/src/Data/Set/Internal.hs
meooow25: Ah yes, done. I think usually "level" goes the opposite direction, so we really should define it here.

containers/src/Data/Set/Internal.hs
meooow25: Done.

containers/src/Data/Set/Internal.hs
```diff
@@ -1823,11 +1823,22 @@ splitRoot orig =
...
+
+-- Proof of complexity: step executes n times. At the ith step,
+-- "insertMin x `mapMonotonic` pxs" takes O(2^i log i) time since pxs has size
+-- 2^i - 1 and we insertMin into its elements which are sets of size <= i.
+-- "insertMin (singleton x)" and "`glue` pxs" are cheaper operations that both
+-- take O(i) time. Over n steps, we have a total cost of
+--
+--   O(\sum_{i=0}^{n-1} 2^i log i)
+-- = O(log n * \sum_{i=0}^{n-1} 2^i)
+-- = O(2^n log n)
```
treeowl, line 1835: `O(\sum_{i=0}^{n-1} 2^i log i)` can't possibly be right, because it has `log 0` in it.

meooow25: Yeah, that's a little icky. Updated to count from 1, step 0 simply runs in O(1).

containers/src/Data/Set/Internal.hs
treeowl: I don't like the phrasing "a logarithmic time". How about something like "This takes time O(log (2^i*m)) = O(i + log m)"?

meooow25: Done.

treeowl:
I believe your analysis of `powerSet` is correct and tight. I am rather curious whether it might be possible to change the algorithm to improve the bound a bit with additional sharing. We know the problem is $O(2^n \log n)$, and it's obviously $\Omega (2^n)$. Can we find an $O(2^n)$ algorithm by improving sharing? We don't currently do anything particularly clever about that. Suppose we have `Bin s _ l r`. We can stick anything between `maximum l` and `minimum r` in the blank while sharing `l` and `r`. In fact, we can link `l` to `r` with a whole sequence of elements. We can choose all sorts of ways to balance trees if it helps improve sharing, though I have no concrete ideas for that.

containers/src/Data/Set/Internal.hs
```diff
@@ -1855,8 +1862,31 @@ cartesianProduct :: Set a -> Set b -> Set (a, b)
...
+-- each element a in as to a set made up of (a,b) for every element b in bs,
+-- taking O(nm) overall. Then we merge these sets up the tree of as, which takes
+-- O(n log m). A brief sketch of proof for the latter:
+--
+-- Consider all nodes in the tree at the same distance from the root to be at
+-- the same "level". The nodes farthest from the root are at level 0, with
+-- levels increasing by 1 towards the root. Being a balanced tree, there are
+-- O(n/2^i) nodes at level i. At every node at level i, we merge the merged left
+-- set, current set, and merged right set into a set of size O(2^i*m) in
+-- O(log (2^i*m)) = O(i + log m) time. Over all levels, we do a total work of
```
treeowl, line 1879: If I'm not mistaken, we have $i \le \lceil \log m \rceil$, so we can drop $i$ and simplify the sums below.

meooow25: Ah no, the tree is `as` which as size n, `bs` has size m.

meooow25:
About powerSet, yes I've thought about a bit too, without success. I see we have #890, I'll check your comments there.

containers/src/Data/Set/Internal.hs
treeowl: Oh, oops. OK.

treeowl (approved):
Awesome work.

PR #892, opened 2023-01-09
Improve the test for powerSet

treeowl:
1. Check for validity of the elements of the result, not just the result.
2. Generate sets in the size range we want, rather than potentially splitting a larger one. This is more efficient and also gives us a better shape distribution.
3. Don't bother comparing to a list-based implementation. Once we've performed all the validity checks (which we should have anyway), a size test is sufficient.


treeowl:
@meooow25, your review would be appreciated.

meooow25 (approved):
Seems alright to me.

Issue #896, opened 2023-01-19
benchmarks for Data.Set.powerSet?


jwaldmann:
`Data.Set.powerSet` has a `testProperty` in containers-tests/tests/set-properties.hs, but no benchmarks. Do we want some?

This is related to #890 but I'm making a separate issue because that thread is already quite long, and mainly on implementation details.

Candidates (I can make a PR)
* for construction:
```
whnf S.powerSet (S.fromList[1..w])
```
* for lookup:
```
whnf (\ s -> all (flip S.member s) s) (S.powerSet (S.fromList [1..w]))
```

What should `w` be? For `w = 20`, I get 
```
  powerSet:                  OK (7.22s)
    471  ms ±  32 ms
  member.powerSet:           OK (18.77s)
    2.561 s ± 106 ms
```
that's a lot (w.r.t. to the other benchmarks)

NB: with my suggested implementation (d516927c8b7f5d5803e0cd4a7c335c01b4dbab70), these numbers become (but that really belongs in the other issue)

```
  powerSet:                  OK (0.79s)
    260  ms ± 2.9 ms
  member.powerSet:           OK (15.90s)
    2.216 s ± 1.9 ms
```

treeowl:
The two benchmarks should have different sizes. Make the construction one a bit larger. For the lookup one, I'd like to see an even and and odd size, both a bit smaller.

treeowl:
Closed in #897 .

PR #897, opened 2023-01-19
add benchmarks for Data.Set.powerSet

jwaldmann:
see #896

### DIFF
--- a/containers-tests/benchmarks/Set.hs
+++ b/containers-tests/benchmarks/Set.hs
@@ -45,8 +45,12 @@ main = do
         , bench "alterF:four:strings" $ whnf (alterF_four strings) strings_s
         , bench "alterF_naive:four" $ whnf (alterF_naive_four elems) s
         , bench "alterF_naive:four:strings" $ whnf (alterF_naive_four strings) strings_s
-        , bench "powerSet" $ whnf S.powerSet (S.fromList[1..20])
-        , bench "member.powerSet" $ whnf (\ s -> all (flip S.member s) s) (S.powerSet (S.fromList [1..20]))
+        , bench "powerSet (19)" $ whnf S.powerSet (S.fromList[1..19])
+        , bench "powerSet (20)" $ whnf S.powerSet (S.fromList[1..20])
+        , bench "powerSet (21)" $ whnf S.powerSet (S.fromList[1..21])
+        , bench "member.powerSet (16)" $ whnf (\ s -> all (flip S.member s) s) (S.powerSet (S.fromList [1..16]))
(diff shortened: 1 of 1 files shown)


containers-tests/benchmarks/Set.hs
```diff
@@ -45,6 +45,8 @@ main = do
         , bench "alterF:four:strings" $ whnf (alterF_four strings) strings_s
         , bench "alterF_naive:four" $ whnf (alterF_naive_four elems) s
         , bench "alterF_naive:four:strings" $ whnf (alterF_naive_four strings) strings_s
+        , bench "powerSet" $ whnf S.powerSet (S.fromList[1..20])
```
treeowl, line 48: 20 seems a bit small.

containers-tests/benchmarks/Set.hs
```diff
@@ -45,6 +45,8 @@ main = do
...
+        , bench "member.powerSet" $ whnf (\ s -> all (flip S.member s) s) (S.powerSet (S.fromList [1..20]))
```
treeowl, line 49: 20 seems a bit large, and if want to check both odd and even sizes.

containers-tests/benchmarks/Set.hs
treeowl: I mean I want 21 instead of 20. Or even better, 19, 20, and 21. Just to bump things up to around a second. Seeing a few gives some sense of actual growth.

containers-tests/benchmarks/Set.hs
treeowl: I was thinking to check on the "alignment" of subset size and depth in different circumstances. But now that I think about it, it's probably better to get rid of the lookup benchmark and instead add a property test checking that. Do you think you can formulate one?

treeowl: I wrote that thinking only about correctness, and not performance. If we can formulate a shape property that will ensure good performance for random lookups, I think that would be valuable too. If someone later finds a substantially faster construction that violates that property, we can evaluate how bad it really is. Checking shape and sizes is a lot cheaper than inspecting the actual contents of each subset, which is what my original property tests used to do (I guess that wasn't my smartest day).

treeowl: I'm fine with having the benchmark, and I won't hold things up for the property test. That said, I don't promise not to add one in the future. While we can't say for certain that a particular property will give better performance for all/any applications, it seems clear that both tree balance and the arrangement of element sizes will have *some* impact on performance, so future changes that modify those should require some thought into whether that's likely a good or bad change. I agree that pure lookups into a power set are silly; it would pretty much always make more sense to keep the underlying set around and do a subset query against that. But a power set that's been modified by relatively few insertions and/or deletions will maintain much of the shape it started with.

treeowl:
Thanks!

PR #904, opened 2023-01-22
Improve powerset

jwaldmann:


treeowl:
No, we don't want to use `vector`. We could pull in some bits of `primitive` if we need to, but I'd rather get those bits shifted into `base`.

treeowl:
Do you want to beat this into mergeable shape by tomorrow so it can go in the next release? I would love to reduce the memory requirements ASAP.

PR #907, opened 2023-01-24
Remove unused NoThunks.hs

meooow25:
This is not used anywhere. There is a copy of this in `tests` which _is_ used.
Going by [this comment](https://github.com/haskell/containers/pull/750/files#r595538743) on the original PR, the author might have forgotten to remove this.


treeowl (approved):
Good catch.

PR #921, opened 2023-01-31
Fix foldMap1 for Data.Tree

### LINKED ISSUE #919: Make foldMap1 for Data.Tree follow the structure
`foldMap1` is implemented in terms of `foldrMap1`, which is ... not great. `foldMap1` should (usually) follow the natural structure of a type. Consider the case where someone decides to use `foldMap1` to implement their own copy of `last`:

```haskell
import Data.Semigroup (Last (..))
myLast :: Tree a -> a
myLast = getLast . foldMap1 Last
```

As currently implemented, this will walk the entire tree in pre-order, which could run us into trouble in an infinite subtree, or slow us down with a large one.

treeowl:
Make `foldMap1` follow the tree structure.

Fixes #919


treeowl:
@meooow25, do you have any better ideas?

treeowl:
This really wants a property test for `Last`. A simple one:

```haskell
forAll $ \n -> getLast (foldMap1 Data.Semigroup.Last (Node undefined (replicate n undefined ++ ["success"]))) === "success"
```

treeowl:
No, the test above won't even typecheck. *Sigh*. We want to have some number of tree levels, each of which looks like `replicate some_number undefined ++ [next]`, where `next` is the next level or a node without children.

treeowl:
@meooow25 I'm curious how much faster the custom `last` was than the default one. Looking at the code, I'd expect them to be very close.

meooow25:
With the current code I see
```
  Foldable1
    fold1
      bin,n=1000:     OK (1.70s)
        5.23 μs ± 357 ns
      bin,n=1000000:  OK (0.58s)
        9.63 ms ± 823 μs
      line,n=1000:    OK (3.53s)
        6.86 μs ±  98 ns
      line,n=1000000: OK (0.80s)
        24.7 ms ± 792 μs
    foldMap1
      bin,n=1000:     OK (1.72s)
        5.54 μs ± 361 ns
      bin,n=1000000:  OK (0.59s)
        10.1 ms ± 706 μs
      line,n=1000:    OK (1.74s)
        6.20 μs ± 341 ns
      line,n=1000000: OK (0.47s)
        19.3 ms ± 1.4 ms
    last
      bin,n=1000:     OK (2.31s)
        76.8 ns ± 6.1 ns
      bin,n=1000000:  OK (2.11s)
        140  ns ±  10 ns
      line,n=1000:    OK (1.65s)
        4.45 μs ± 333 ns
      line,n=1000000: OK (0.59s)
        8.82 ms ± 783 μs
```

But there seems to be an inefficiency with `foldMap1` for `NonEmpty`. It is defined as
```hs
foldMap1 f (x :| xs) = go (f x) xs where
    go y [] = y
    go y (z : zs) = y <> go (f z) zs
```

But using a definition like
```hs
foldMap1 f (x0 :| xs0) = go x0 xs0 where
    go x [] = f x
    go x (x':xs) = f x <> go x' xs
```

The benchmarks improve:
```
  Foldable1
    fold1
      bin,n=1000:     OK (1.71s)
        6.70 μs ± 413 ns
      bin,n=1000000:  OK (0.59s)
        11.6 ms ± 810 μs
      line,n=1000:    OK (2.42s)
        5.85 μs ±  90 ns
      line,n=1000000: OK (0.71s)
        6.08 ms ± 350 μs
    foldMap1
      bin,n=1000:     OK (1.77s)
        7.55 μs ± 372 ns
      bin,n=1000000:  OK (0.61s)
        12.1 ms ± 1.1 ms
      line,n=1000:    OK (1.68s)
        6.23 μs ± 410 ns
      line,n=1000000: OK (1.05s)
        6.07 ms ± 177 μs
    last
      bin,n=1000:     OK (2.43s)
        64.5 ns ± 3.1 ns
      bin,n=1000000:  OK (3.22s)
        117  ns ± 1.4 ns
      line,n=1000:    OK (1.61s)
        3.96 μs ± 348 ns
      line,n=1000000: OK (0.80s)
        7.94 ms ± 606 μs
```

`base` should probably fix this, but it seems we can sidestep it by using `NonEmpty`'s `foldrMap1` instead of `foldMap1` to get the same good performance:

```hs
foldMap1 f = go
  where
    go (Node x []) = f x
    go (Node x (t : ts)) = f x <> Foldable1.foldrMap1 go (\t' z -> go t' <> z) (t :| ts)
```

And you're right about `last`, the default seems just as good as our definition now so we can remove it 🎉

meooow25:
About the tests, maybe we can simply generate an ordinary tree and replace all the leaves except for the last with `undefined`?
Another thing that seems a good idea, since this is about the structure of the tree, is to test with `Magma` against `foldTree`.

treeowl:
Good efficiency catch, and the replacement testing idea sounds okay. `Magma` won't work, though, because it's not a semigroup. We want the free semigroup over a set, so `NonEmpty A`. What property should we test with that?

meooow25:
What I mean is we can make `Magma` a semigroup for the test, so that we can test the structure.

```hs
-- unlawful, not associative!
instance Semigroup (Magma a) where
  (<>) = (:*)

prop_foldMap1 :: Tree A -> Property
prop_foldMap1 t = foldMap1 Inj t === foldTree (\x xs -> fold1 (Inj x :| xs)) t
```

treeowl:
Hmmm.... So to test the exact folding pattern? That test would need documentation, since it's a stricter standard than one might expect.

treeowl:
Would you like to open a GHC ticket to change the `base` definition?

meooow25:
I'll open a ticket 👍

Issue #157, opened 2015-06-11
Data.Graph: document complexities O(...)


andreasabel:
The documentation of Data.Graph misses complexity information.

foxik:
Any pull requested with the missing complexity information would be appreciated :-)

andreasabel:
Yes, of course.  I won't prepare it, though, as I have 2^n Agda bugs chasing me... ;-)

stefan-j:
Well in the study they indicate their DFS empirically as linear - O (V + E)

> Using these models, (see for example Sands (1993)) we believe our implementation of the DFS algorithms to be linear, but because these models have not been fully tested, we also ran empirical tests ourselves

Strongly connected components is two DFS so its linear as well

Issue #648, opened 2019-06-24
improve documentation for Data.Graph


jwaldmann:
(see messages in this thread https://mail.haskell.org/pipermail/libraries/2019-June/029674.html)

we could
* add a sentence (in the module header) stating that the purpose is to compute SCCs quickly (we can even copy the O(V+E) claim from the paper?)
* add complexity information to some functions

Example for non-obvious behaviour: http://hackage.haskell.org/package/containers-0.6.1.1/docs/src/Data.Graph.html#graphFromEdges will sort the edge list (so you can't build the graph in linear time?) and the `key_vertex` function it returns, will take logarithmic time?

meooow25:
The documentation is better today: https://hackage.haskell.org/package/containers-0.7/docs/Data-Graph.html
@jwaldmann do you think there is anything else that should be added?

meooow25:
Do reopen if there is more that can be improved.

PR #923, opened 2023-01-31
Update Data.Graph documentation

meooow25:
*  Add time complexities for `Data.Graph` functions.
*  Avoid the `Forest` synonym as it can be confusing. We stopped using it in `Data.Tree` in #665.
* Document expected input graph properties for `topSort`, `reverseTopSort` and `bcc`.

### DIFF
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -215,8 +215,8 @@ flattenSCC :: SCC vertex -> [vertex]
 flattenSCC (AcyclicSCC v) = [v]
 flattenSCC (CyclicSCC vs) = vs
 
--- | The strongly connected components of a directed graph, reverse topologically
--- sorted.
+-- | \(O((V+E) \log V)\). The strongly connected components of a directed graph,
+-- reverse topologically sorted.
 --
 -- ==== __Examples__
 --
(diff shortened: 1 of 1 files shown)


meooow25:
One more for the release 😄

containers/src/Data/Graph.hs
```diff
@@ -383,8 +383,8 @@ graphFromEdges' x = (a,b) where
     (a,b,_) = graphFromEdges x
 {-# INLINABLE graphFromEdges' #-}
 
--- | Build a graph from a list of nodes uniquely identified by keys,
--- with a list of keys of nodes this node should have edges to.
+-- | \(O((V+E) \log V)\). Build a graph from a list of nodes uniquely identified
```
treeowl, line 386: Isn't it $O(V + E \log V)$?

containers/src/Data/Graph.hs
```diff
@@ -610,19 +612,25 @@ preArr bnds      = tabulate bnds . preorderF
...
 postOrd g = postorderF (dff g) []
 
--- | A topological sort of the graph.
+-- | \(O(V+E)\). A topological sort of the graph.
 -- The order is partially specified by the condition that a vertex /i/
 -- precedes /j/ whenever /j/ is reachable from /i/ but not vice versa.
+--
+-- Note: A topological sort exists only when there are no cycles in the graph.
+-- If the graph has cycles, the output of this function will not be a
+-- topological sort. In such a case consider using 'scc'.
```
treeowl, line 627: It would be good to mention that `reverseTopSort` is (currently?) faster.

treeowl:
I'm going to merge this for the release, but I think my comments should be considered for the next one.

containers/src/Data/Graph.hs
meooow25: The sort step is $O(V \log V)$, then the graph takes $O(V + E \log V)$ to build, so the total seems right to me.

containers/src/Data/Graph.hs
meooow25: Because of the `reverse`? That seems avoidable. We build the list from the tree so we should also be able to build the reverse of the list directly. I can make a PR later.

containers/src/Data/Graph.hs
treeowl: You're absolutely right. I got my edges and vertices crossed for a moment.

meooow25:
I think we can close #157 and #648 from this PR, @treeowl?

PR #927, opened 2023-02-01
Minor documentation fixes

treeowl:
* Add `@since` annotations for the `MonadZip` instances for `Data.Sequence` and `Data.Tree`.

* Fix Haddock markup.


treeowl:
I've already pushed these changes to Hackage.

PR #933, opened 2023-02-12
Update links for the `OverloadedLists` extension

### LINKED ISSUE #932: Different Links to OverloadedLists Extension
The documentation for [sequences](https://github.com/haskell/containers/blob/fe4aa23b67cb14a6966c778f66acaecd0ef99f4f/containers/docs/sequence.rst) and the documentation for [maps](https://github.com/haskell/containers/blob/fe4aa23b67cb14a6966c778f66acaecd0ef99f4f/containers/docs/map.rst) both have a link to the Overloaded Lists Extension but they are different and the one in sequences goes to a 404. Perhaps https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/overloaded_lists.html could be used for both links as the page seems more likely to be updated than the [link used for maps](https://gitlab.haskell.org/ghc/ghc/-/wikis/overloaded-lists) which goes to some page on the gitlab wiki.

konsumlamm:
Fixes #932.

### DIFF
--- a/containers/docs/map.rst
+++ b/containers/docs/map.rst
@@ -77,11 +77,11 @@ The following GHCi session shows some of the basic map functionality::
     > fromList [(1,"one"),(2,"two"),(3,"new three"),(4,"newer four")]
 
 .. TIP:: You can use the `OverloadedLists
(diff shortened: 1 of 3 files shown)


treeowl:
Thanks.

PR #934, opened 2023-02-16
More foldMap related Tree tests

### LINKED ISSUE #924: Add laziness tests for Tree foldMap1
#921 fixed a bug where `foldMap1` was too strict. It needs a regression test. We may even want to test for the specific folding shape, as @meooow25 suggested in a comment there.

meooow25:
* Test that last is lazy enough to only look at the path down to the last leaf
* Test that foldMap and foldMap1 fold according to the structure of the tree

Closes #924

### DIFF
--- a/containers-tests/tests/tree-properties.hs
+++ b/containers-tests/tests/tree-properties.hs
@@ -43,6 +43,7 @@ main = defaultMain $ testGroup "tree-properties"
 #if MIN_VERSION_base(4,18,0)
          , testProperty "toNonEmpty"               prop_toNonEmpty
          , testProperty "last"                     prop_last
+         , testProperty "last_path"                prop_last_path
          , testProperty "foldrMap1"                prop_foldrMap1
          , testProperty "foldlMap1'"               prop_foldlMap1'
          , testProperty "foldlMap1"                prop_foldlMap1
@@ -184,6 +185,18 @@ prop_toNonEmpty t = Foldable1.toNonEmpty t === NE.fromList (toList t)
 
 prop_last :: Tree A -> Property
 prop_last t = Foldable1.last t === NE.last (Foldable1.toNonEmpty t)
+
+-- Tests that last only looks at the path going down to the last leaf.
+prop_last_path :: Tree A -> Property
+prop_last_path t = Foldable1.last (replace t) === Foldable1.last t
+  where
+    -- Replace all trees with bottom except for the last one.
+    replace :: Tree a -> Tree a
+    replace (Node x ts) = Node x (replaces ts)
+    replaces :: [Tree a] -> [Tree a]
+    replaces [] = []
+    replaces [t] = [replace t]
+    replaces (t:ts) = error "error tree" : replaces ts
 
 prop_foldrMap1 :: Tree A -> Property
 prop_foldrMap1 t =
--- a/containers-tests/tests/tree-properties.hs
+++ b/containers-tests/tests/tree-properties.hs
@@ -11,9 +11,13 @@ import Test.QuickCheck.Function (apply)
 import Test.QuickCheck.Poly (A, B, C, OrdA)
 import Control.Monad.Fix (MonadFix (..))
 import Control.Monad (ap)
-import Data.Foldable (foldl', toList)
+import Data.Foldable (fold, foldl', toList)
 import Data.Traversable (foldMapDefault)
+#if !(MIN_VERSION_base(4,11,0))
+import Data.Semigroup (Semigroup ((<>)))
+#endif
 #if MIN_VERSION_base(4,18,0)
+import Data.List.NonEmpty (NonEmpty (..))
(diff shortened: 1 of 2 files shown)


containers-tests/tests/tree-properties.hs
```diff
@@ -150,6 +176,12 @@ prop_foldMap t =
   foldMap (:[]) t === toList t .&&.
   foldMap (:[]) t === foldMapDefault (:[]) t
 
+-- We use UnitalMagma with foldMap to test that the structure of the fold
+-- follows that of the tree. This is a desirable property to have.
```
treeowl, line 180: It's not *always* a desirable property, but it's a desirable property for `Tree`. I would love to be able to articulate exactly why, but I'm struggling. I'll give you a few examples of yes/no, and maybe you can figure out a concise way to say what needs to be said?

* `Data.Tree`: the structure is lazy and exposed concretely, with no external (or fancy-type-induced) data structure invariants—the fold should follow the structure.
* `Data.Set`: the structure is strict and has a size annotation invariant we can use easily to avoid unnecessarily appending `mempty`s at the leaves—the fold should not precisely follow the structure.
* A scheduled banker's queue: A lazy fold is expected to require $O(1)$ actual time per element; following the structure would lead to an $O(n)$ pause to reverse the rear, so the fold must not follow the structure. (A strict fold is technically allowed to follow the structure, but doing so might be bad for caching or concurrency and should probably be avoided.)

Perhaps it would be best to avoid generalities, and just try to explain why following the structure of the tree is appropriate for this structure?

containers-tests/tests/tree-properties.hs
```diff
@@ -179,12 +211,28 @@ prop_product :: Tree OrdA -> Property
 prop_product t = product t === product (toList t)
 
 #if MIN_VERSION_base(4,18,0)
+-- We use Magma with foldMap1 to test that the structure of the fold follows
+-- that of the tree. This is a desirable property to have.
```
treeowl, line 215: Same here.

containers-tests/tests/tree-properties.hs
```diff
@@ -179,12 +211,28 @@ prop_product :: Tree OrdA -> Property
...
 
+-- Tests that last only looks at the path going down to the last leaf.
+prop_last_path :: Tree A -> Property
+prop_last_path t = Foldable1.last (replace t) === Foldable1.last t
+  where
+    -- Replace all trees with bottom except for the last one.
+    replace (Node x (t:ts)) = Node x ts'
+      where
+        ts' = Foldable1.foldrMap1 ((:[]) . replace) (const (error "error tree" :)) (t :| ts)
+    replace t@(Node _ []) = t
```
treeowl, line 234: I can't read this. Maybe that's just me, but I suspect we can be clearer. The following will be slightly less efficient, but I don't think we have to worry about that too much for this test.

```haskell
replace (Node x ts) = Node x (replaces ts)
replaces :: [Tree a] -> [Tree a]
replaces [] = []
replaces [t] = [replace t]
replaces (t : ts) = error "error tree" : replaces ts
```

treeowl (changes requested):

containers-tests/tests/tree-properties.hs
meooow25: Sure, done.

containers-tests/tests/tree-properties.hs
treeowl: Too strict.

```haskell
foldMap (:[]) (Node x undefined) = x : _|_
foldMap1 (:[]) (Node x undefined) = _|_
```

and I don't think the latter can be avoided.

meooow25: Ah those pesky bottoms.

> Perhaps it would be best to avoid generalities, and just try to explain why following the structure of the tree is appropriate for this structure?

Do you have an example of a `Monoid` where this implementation is preferable? I'm finding it hard to come up with one. For `Semigroup` we have `Last`.

treeowl: Well, `First a` is a simple example, or `[a]` as I demonstrated already. I'm not sure what you're asking.

meooow25: I mean a `Monoid` where `foldMap` according to the tree structure is preferable to the default that uses `foldr`.

treeowl: Well, there's `Maybe (Last a)`, AKA `Data.Monoid.Last a`.

meooow25: I guess that works. Updated, how does it look?

containers-tests/tests/tree-properties.hs
meooow25: Updated.

meooow25:
I forgot about this, @treeowl does this look good?

treeowl:
Thanks! Sorry that took so long.

Issue #908, opened 2023-01-25
What code is used to generate the appendTree0 & friends in Data.Sequence.Internal?


SamuelSchlesinger:
Just browsing the code to learn and saw
```haskell
-- The appendTree/addDigits gunk below is machine generated
```
in Data.Sequence.Internal. Wondering what code was used to generate that stuff?

treeowl:
You'd have to ask @RossPaterson or [Ralf Hinze](https://pl.cs.uni-kl.de/homepage/en/staff/RalfHinze/). I don't know if that code was preserved somewhere outside the repo.

RossPaterson:
Here it is: [mkappend.hs](https://github.com/haskell/containers/files/10498738/mkappend.txt). This code is 18 years old, and it seems the generated code has had some strictness annotations added to it since,

treeowl:
@RossPaterson , yes, the amortization analysis lets us force the spines, and doing so seems quite beneficial for that operation.

treeowl:
@RossPaterson And thanks!

RossPaterson:
@treeowl Have you done an amortization analysis? I see you consistently force the new middle tree in the `Deep`-`Deep` case of `appendTree`. You also force the collected nodes between the trees in some cases of `appendTree` and `addDigits`, but I couldn't see a consistent pattern to it.

treeowl:
@RossPaterson A fixed number of debits are allowed on each `Deep` node, and none elsewhere, so we're allowed to force along the spine as deep as the shallower tree, which is what we do (unless I made a mistake somewhere). The question of exactly how far off to force was guided by benchmarks; I think it's a matter of the cost of thunk creation vs. other calculation. If you think I've made a mistake somewhere, or could do better, *please* let me know.

SamuelSchlesinger:
Thanks much.

treeowl:
I think we should add the code to the repo for historical reference and so people can learn from it.

treeowl:
@RossPaterson Are you satisfied that I haven't broken the amortized bounds? This conversation got stuck in the back of my mind.

PR #945, opened 2023-04-19
added mkappend.hs

SamuelSchlesinger:
for #908

let me know if I should put the file elsewhere


containers/containers.cabal
```diff
@@ -24,6 +24,7 @@ cabal-version:  >=1.10
 extra-source-files:
     include/containers.h
     changelog.md
+    mkappend.hs
```
treeowl, line 27: I'm not familiar with this section of the Cabal file. What is the purpose of adding the file there?

containers/src/Data/Sequence/Internal.hs
```diff
@@ -1888,7 +1888,7 @@ snocTree' (Deep s pr m (One a)) b =
 (><)            :: Seq a -> Seq a -> Seq a
 Seq xs >< Seq ys = Seq (appendTree0 xs ys)
 
--- The appendTree/addDigits gunk below is machine generated
+-- The appendTree/addDigits gunk below is machine generated via mkappend.hs
```
treeowl, line 1891: Please point out that while it was originally generated thus, it has since been edited to add strictness annotations.

SamuelSchlesinger: good call

containers/containers.cabal
SamuelSchlesinger: It is a list of additional files to be included in source distributions -- figured it made sense to include here, though won't really effect many users if we didn't include it

treeowl: Ah, okay. So ... stuff that you can download using `cabal-install`?

SamuelSchlesinger: I believe it'll cause it to be included in the tarball uploaded to/downloaded from Hackage, yes. Tbh my confidence is somewhat low on my understanding of this, just bumped up against it a few times when messing with my own packages a while ago.

treeowl: I guess that's also means people will be able to browse it on Hackage, which is good.

treeowl (approved):

containers/containers.cabal
SamuelSchlesinger: Actually, not sure you can actually view this list of files in the UI. Would probably make sense for Hackage to show them, but an example I used it for was a compiled frontend for a web package, so maybe there are some packages for which having the whole index wouldn't render very nicely.

treeowl: Won't they show up with all the rest of the files in [this listing](https://hackage.haskell.org/package/containers-0.6.7/src/)?

SamuelSchlesinger: Oh neat, never saw that. Thanks!

treeowl: Search for "Cabal source package" on the main Hackage page of any package and click on "browse". It's kind of buried, but it's there, and useful.

PR #948, opened 2023-05-20
Future-proof against potential Prelude.foldl'

Bodigrim:
See https://github.com/haskell/core-libraries-committee/issues/167

This is, of course, a bit speculative at the moment, but given that the change does not involve CPP, it should not hurt to merge it early to simplify further impact assessment.


Bodigrim:
@treeowl how does it look?

treeowl:
Looks boring. I like it.

Issue #952, opened 2023-06-26
NonEmpty for CyclicSCC


Bodigrim:
https://github.com/haskell/containers/blob/48196fba49386a228939824066f2215f933ce302/containers/src/Data/Graph.hs#L146-L149

How about changing this to `CyclicSCC (NonEmpty vertex)`? This could shave a number of unsafe `Data.List.head` in GHC and other compilers. 

Not sure about backward compatibility / migration schedule, probably a pattern synonym can help to ease the breakage.

treeowl:
Yes, let's do something of this sort. Your second representation doesn't seem to distinguish properly between vertices with no edges and ones with a single edge to themselves.

```haskell
import Data.Graph

main :: IO ()
main =  do
  print $ stronglyConnComp [((), (), [()])]
  print $ stronglyConnComp [((), (), [])]
```

This gives

```
[CyclicSCC [()]]
[AcyclicSCC ()]
```

As I understand it, your `SCC` would represent these two the same.

treeowl:
I must admit to some trepidation about committing to `NonEmpty` given the current lack of a plan for improving its laziness story. If we do use it, I imagine it should be strict and unpacked.

treeowl:
I meant if you want `NonEmpty` it should probably be `{-# UNPACK #-} !(NonEmpty vertex)`. But unpacking by hand wouldn't be unreasonable either.

treeowl:
I'd like to see how this affects the code, yeah.

meooow25:
Can we maybe skip introducing the partial `CyclicSCC` -> `CyclicSCC'` pattern? It's unlikely someone would be making their own `SCC`s. And even with the pattern it's a breaking change if they do `CyclicSCC []`.

treeowl:
@meooow25 , we can just make it unidirectional, right?

meooow25:
Yes, that's what I meant, should have stated explicitly.  
Also this is only a suggestion, it comes down to whether we really don't want to break some code out there that is constructing `CyclicSCC`s and always with non-empty lists.

PR #953, opened 2023-06-27
Data.Graph.SCC: store mutually reachable vertices in a non-empty list

Bodigrim:
As discussed in #952. 

CC @meooow25

### DIFF
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -5,8 +5,11 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveLift #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE Safe #-}
+{-# LANGUAGE TemplateHaskellQuotes #-}
+{-# LANGUAGE ViewPatterns #-}
 #endif
 
 #include "containers.h"
@@ -74,7 +77,11 @@ module Data.Graph (
 
 
     -- * Strongly Connected Components
-    , SCC(..)
+    , SCC(..
+#ifdef __GLASGOW_HASKELL__
+      , CyclicSCC
+#endif
+      )
 
     -- ** Construction
     , stronglyConnComp
@@ -107,6 +114,9 @@ import Data.Tree (Tree(Node), Forest)
 
 -- std interfaces
 import Data.Foldable as F
+#if MIN_VERSION_base(4,18,0)
+import qualified Data.Foldable1 as F1
+#endif
 import Control.DeepSeq (NFData(rnf))
 import Data.Maybe
 import Data.Array
@@ -117,6 +127,8 @@ import Data.Array.Unboxed ( UArray )
 import qualified Data.Array as UA
 #endif
 import qualified Data.List as L
+import Data.List.NonEmpty (NonEmpty(..))
+import qualified Data.List.NonEmpty as NE
 import Data.Functor.Classes
 #if !MIN_VERSION_base(4,11,0)
 import Data.Semigroup (Semigroup (..))
@@ -124,7 +136,7 @@ import Data.Semigroup (Semigroup (..))
 #ifdef __GLASGOW_HASKELL__
 import GHC.Generics (Generic, Generic1)
 import Data.Data (Data)
-import Language.Haskell.TH.Syntax (Lift)
+import Language.Haskell.TH.Syntax (Lift(..))
 -- See Note [ Template Haskell Dependencies ]
 import Language.Haskell.TH ()
 #endif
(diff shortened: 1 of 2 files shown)


containers/src/Data/Graph.hs
```diff
@@ -215,7 +241,7 @@ flattenSCCs = concatMap flattenSCC
 -- | The vertices of a strongly connected component.
 flattenSCC :: SCC vertex -> [vertex]
 flattenSCC (AcyclicSCC v) = [v]
-flattenSCC (CyclicSCC vs) = vs
+flattenSCC (CyclicSCC' vs) = NE.toList vs
```
treeowl, line 244: Surely we should add a `NonEmpty` version....

containers/src/Data/Graph.hs
```diff
@@ -233,11 +259,7 @@ stronglyConnComp
                 -- nodes of the graph; such edges are ignored.
         -> [SCC node]
 
-stronglyConnComp edges0
-  = map get_node (stronglyConnCompR edges0)
-  where
-    get_node (AcyclicSCC (n, _, _)) = AcyclicSCC n
-    get_node (CyclicSCC triples)     = CyclicSCC [n | (n,_,_) <- triples]
+stronglyConnComp = map (fmap (\(n, _, _) -> n)) . stronglyConnCompR
```
treeowl, line 262: This is lazier than what we had before. That inner `fmap` isn't nice.

treeowl (changes requested):

containers/src/Data/Graph.hs
treeowl: `[n | (n,_,_) <- triples]` is more eager than `List.map (\(n,_,_) -> n) triples`. In some other contexts, list fusion and inlining will turn the latter into the former, but I really don't think that's the case here.

containers/src/Data/Graph.hs
treeowl: Oh wait, there isn't a `NonEmpty` version. Oops. Never mind me....

containers/src/Data/Graph.hs
```diff
@@ -166,47 +195,72 @@ deriving instance Generic1 SCC
...
+  lift (AcyclicSCC v) = AppE <$> [| AcyclicSCC |] <*> lift v
+  lift (CyclicSCC' vs) = AppE <$> [| CyclicSCC' |] <*> liftNonEmpty vs
+
+liftNonEmpty :: Lift vertex => NonEmpty vertex -> Q Exp
+liftNonEmpty (x :| xs) = do
+  let nonemptyName = mkNameG DataName "base" "GHC.Base" ":|"
+  x' <- lift x
+  xs' <- lift xs
+  return (InfixE (Just x') (ConE nonemptyName) (Just xs'))
+#endif
```
treeowl, line 213: This looks way too complicated. `TemplateHaskellQuotes` has been around since 8.0; shouldn't you be able to use that here instead of all this low-level fiddling?

treeowl:
There seem to be two issues remaining:

1. Constructor name
2. CI failure on 8.0

I don't really like the primed name, but I'm willing to go along with it if no one can come up with anything better. For the CI failure, I'm okay with dropping support for 8.0, but that will definitely need its own commit, and should probably be a separate PR merged before this one.

treeowl:
Could you please rebase this?

treeowl:
> > I don't really like the primed name, but I'm willing to go along with it if no one can come up with anything better.
> 
> `NECyclicSCC`? `CyclicSCC1`?

They're all horrible (and my own attempts were worse), but I think I dislike `NECyclicSCC` the least.

treeowl:
Yes, please!

treeowl (approved):

treeowl:
Thanks!!!

containers/src/Data/Graph.hs
```diff
@@ -139,15 +151,26 @@ default ()
...
+  -- ^ A maximal set of mutually reachable vertices.
+  --
+  -- @since 0.7.0
   deriving ( Eq   -- ^ @since 0.5.9
            , Show -- ^ @since 0.5.9
            , Read -- ^ @since 0.5.9
            )
 
+-- | Partial pattern synonym for backward compatibility with @containers < 0.7@.
+pattern CyclicSCC :: [vertex] -> SCC vertex
```
alexfmpe, line 167: Doesn't the pattern synonym definition also need to be guarded by `#ifdef __GLASGOW_HASKELL__` ? 
I'm always confused by the portability story. Tangentially related, how do we even test this given CI only uses ghc?

treeowl: Yes it does. The portability story kind of sucks right now. The main goal as I see it is to make sure it won't be too big a pain to add support for a future Report-style Haskell (Purescript, sadly, is just a bit too far away to be a practical target for this). I think we should stop using `__GLASGOW_HASKELL__` directly and instead use a few of our own macros. We should have a build job that pretends were not on GHC; getting that to do testing right will take a bit of fussing, but it's *probably* doable?

PR #954, opened 2023-06-28
Drop support of GHC 8.0 and base-4.9

Bodigrim:
As discussed in https://github.com/haskell/containers/pull/953#issuecomment-1610428490.


containers/src/Utils/Containers/Internal/Prelude.hs
treeowl, line 1: I *think* we still need this to export `liftA2` *as a class method*. Could you update the comment? Do we still need `Foldable` here? I don't remember why that was necessary.

treeowl (approved):
Looks good. Thanks!

PR #956, opened 2023-07-06
Clarify replicateM Haddock

tomjaguarpaw:
"`replicateM` is a sequence counterpart of `replicateM`" doesn't really make much sense, so this PR clarifies it.

### Before

screenshot1

### After

screenshot1

### DIFF
--- a/containers/src/Data/Sequence/Internal.hs
+++ b/containers/src/Data/Sequence/Internal.hs
@@ -1707,7 +1707,8 @@ replicateA n x
   | otherwise   = error "replicateA takes a nonnegative integer argument"
 {-# SPECIALIZE replicateA :: Int -> State a b -> State a (Seq b) #-}
 
--- | 'replicateM' is a sequence counterpart of 'Control.Monad.replicateM'.
+-- | 'replicateM' is the @Seq@ counterpart of
+-- @Control.Monad.'Control.Monad.replicateM'@.
 --
 -- > replicateM n x = sequence (replicate n x)
 --


tomjaguarpaw:
The CI error is `Error: retrieving gpg key timed out.`

containers/src/Data/Sequence/Internal.hs
```diff
@@ -1707,7 +1707,8 @@ replicateA n x
   | otherwise   = error "replicateA takes a nonnegative integer argument"
 {-# SPECIALIZE replicateA :: Int -> State a b -> State a (Seq b) #-}
 
--- | 'replicateM' is a sequence counterpart of 'Control.Monad.replicateM'.
+-- | 'replicateM' is the @Seq@ counterpart of
+-- @Control.Monad.@'Control.Monad.replicateM'.
```
treeowl, line 1711:
```suggestion
-- @"Control.Monad".'Control.Monad.replicateM'@.
```

I'm pretty sure that's how we usually do it, and it feels better.

tomjaguarpaw: OK, as you prefer.

tomjaguarpaw: (FWIW I think it's less clear to link to the `Control.Monad` module *as well*. It distracts from point of the link, which is just to `replicateM`.)

treeowl: Ah, in that case drop the double quotes, but I still want the ampersand shifted to the end.

tomjaguarpaw: Makes sense. Done.

treeowl:
Thanks! I'll merge once CI completes.

tomjaguarpaw:
I believe this is ready for merge.

treeowl:
Sorry I'm being sluggish. I've been sick with COVID for a week, which explains some of it....

tomjaguarpaw:
That's OK. Hope you feel better soon!

PR #957, opened 2023-07-06
Elaborate on asymptotics of IntMap

Bodigrim:
Recently I had a long discussion elsewhere comparing asymptotics of `IntMap` to balanced binary trees. I think it's worth writing down in the documentation.

### DIFF
--- a/containers/src/Data/IntMap.hs
+++ b/containers/src/Data/IntMap.hs
@@ -50,10 +50,29 @@
 --
 -- Operation comments contain the operation time complexity in
 -- the Big-O notation <http://en.wikipedia.org/wiki/Big_O_notation>.
+--
 -- Many operations have a worst-case complexity of \(O(\min(n,W))\).
 -- This means that the operation can become linear in the number of
 -- elements with a maximum of \(W\) -- the number of bits in an 'Int'
--- (32 or 64).
+-- (32 or 64). These peculiar asymptotics are determined by the depth
+-- of the Patricia trees:
+--
+-- * even for an extremely unbalanced tree, the depth cannot be larger than
+--   the number of elements \(n\),
+-- * each level of a Patricia tree determines at least one more bit
+--   shared by all subelements, so there could not be more
+--   than \(W\) levels.
+--
+-- If all \(n\) keys in the tree are between 0 and \(N\) (or, say, between \(-N\) and \(N\)),
+-- the estimate can be refined to \(O(\min(n, \log N))\). If the set of keys
+-- is sufficiently "dense", this becomes \(O(\min(n, \log n))\) or simply
+-- the familiar \(O(\log n)\), matching balanced binary trees.
+--
+-- The most performant scenario for 'IntMap' are keys from a contiguous subset,
+-- in which case the complexity is proportional to \(\log n\), capped by \(W\).
+-- The worst scenario are exponentially growing keys \(1,2,4,\ldots,2^n\),
+-- for which complexity grows as fast as \(n\) but again is capped by \(W\).
+--
 -----------------------------------------------------------------------------
 
 module Data.IntMap


containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,24 @@
 --
 -- Operation comments contain the operation time complexity in
 -- the Big-O notation <http://en.wikipedia.org/wiki/Big_O_notation>.
+--
 -- Many operations have a worst-case complexity of \(O(\min(n,W))\).
 -- This means that the operation can become linear in the number of
 -- elements with a maximum of \(W\) -- the number of bits in an 'Int'
--- (32 or 64).
+-- (32 or 64). This peculiar asymptotics is determined by the depth
+-- of Patricia trees:
```
konsumlamm, line 58:
```suggestion
-- (32 or 64). These peculiar asymptotics are determined by the depth
-- of the Patricia trees:
```

containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,24 @@
...
+-- * even for extremely unbalanced tree the depth cannot be larger than
+--   the number of elements \(n\),
+-- * each level of a Patricia tree determines at least one more bit
+--   shared by all subelements, so there could not be more
+--   than \(W\) levels.
+--
+-- If all \(n\) keys in the tree are less than \(N\),
+-- the estimate can be refined to \(O(\min(n, \log N))\). If the set of keys
+-- is sufficiently "dense", this becomes \(O(\min(n, \log n))\) or simply
+-- familiar \(O(\log n)\), matching balanced binary trees.
```
konsumlamm, line 69:
```suggestion
-- the familiar \(O(\log n)\), matching balanced binary trees.
```

containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,24 @@
...
+-- (32 or 64). This peculiar asymptotics is determined by the depth
+-- of Patricia trees:
+--
+-- * even for extremely unbalanced tree the depth cannot be larger than
+--   the number of elements \(n\),
+-- * each level of a Patricia tree determines at least one more bit
+--   shared by all subelements, so there could not be more
+--   than \(W\) levels.
+--
+-- If all \(n\) keys in the tree are less than \(N\),
```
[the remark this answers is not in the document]
meooow25: Fair enough, I'm only seeing if the situation can be generalized to give the reader more information. But it might be good enough to address only the typical scenario.

containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,24 @@
...
 -- elements with a maximum of \(W\) -- the number of bits in an 'Int'
--- (32 or 64).
+-- (32 or 64). This peculiar asymptotics is determined by the depth
+-- of Patricia trees:
+--
+-- * even for extremely unbalanced tree the depth cannot be larger than
+--   the number of elements \(n\),
+-- * each level of a Patricia tree determines at least one more bit
+--   shared by all subelements, so there could not be more
+--   than \(W\) levels.
```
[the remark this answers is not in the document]
meooow25: > I find it helpful to bring a quick exposition without forcing people to look into the paper.

I disagree... I think documentation should either explain well or point to something which does. The two points about the tree only raise more questions. But I'll leave it to treeowl.

>  It's important to explain that it's more like "it's normally more like log N, which is capped by W, but in unbalanced edge cases can grow as fast as n, but again capped by W"

This is a fair summary that could be documented.

treeowl:
I don't really think there's anything very logarithmic going on here in *general*. The bounds we give are conservative approximations. If we can give tighter, but still conservative, approximations, all the better. But vague "probably this good unless your data aren't what we like" doesn't seem so valuable.

Lysxia:
The case where your keys are all in a small interval seems like a common enough situation that it's worth pointing out  the logarithmic behavior in that case. min(n,64) doesn't actually tell you much because in practice maps contain less than 2^64 elements, so log(n) < 64. There's still an order of magnitude between a balanced tree of depth 10 and a Patricia tree of depth 64. So it's good to point out that cases where an `IntMap` has depth 64 are actually quite pathological.

treeowl:
Agreed. In many cases, we can give bounds in terms of the maximum minus the minimum, right? But that is also pretty conservative, I think, since a map with keys like `[minBound .. minBound + 10000] ++ [maxBound - 10000 .. maxBound]` should be relatively efficient too, if I'm not mistaken. Is there a way we can talk about tree depth that's tighter while still being tractable?

Bodigrim:
@treeowl I rebased and extended the description, how does it look now? It's purely a documentation change, I'd love to get it decided on one way or another.

containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,29 @@
...
+--
 -- Many operations have a worst-case complexity of \(O(\min(n,W))\).
 -- This means that the operation can become linear in the number of
 -- elements with a maximum of \(W\) -- the number of bits in an 'Int'
--- (32 or 64).
+-- (32 or 64). These peculiar asymptotics are determined by the depth
+-- of the Patricia trees:
+--
+-- * even for extremely unbalanced tree the depth cannot be larger than
+--   the number of elements \(n\),
```
konsumlamm, line 61:
```suggestion
-- * even for an extremely unbalanced tree, the depth cannot be larger than
--   the number of elements \(n\),
```

Bodigrim: Thanks!

Lysxia:
That looks pretty good to me!

Bodigrim:
@treeowl just another reminder to take a look at the proposed documentation change.

treeowl:
I'll try to have a look tonight!

Bodigrim:
@treeowl one more ping.

Bodigrim:
@treeowl just a gentle reminder to review this PR.

containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,29 @@
...
+-- * even for an extremely unbalanced tree, the depth cannot be larger than
+--   the number of elements \(n\),
+-- * each level of a Patricia tree determines at least one more bit
+--   shared by all subelements, so there could not be more
+--   than \(W\) levels.
+--
+-- If all \(n\) keys in the tree are between 0 and \(N\),
+-- the estimate can be refined to \(O(\min(n, \log N))\). If the set of keys
+-- is sufficiently "dense", this becomes \(O(\min(n, \log n))\) or simply
+-- the familiar \(O(\log n)\), matching balanced binary trees.
```
treeowl, line 69: What if there are negative keys? Can we give a similarly concise refinement in that case?

containers/src/Data/IntMap.hs
```diff
@@ -50,10 +50,29 @@
...
+--
+-- The most performant scenario for 'IntMap' are keys from a continuous subset,
```
treeowl, line 71: Is "continuous" the right word here? Maybe "contiguous"?

treeowl (commented):
I'm sorry it took so long to review this. I have just a couple questions above.

Bodigrim:
@treeowl thanks, updated and rebased. Good to go?

containers/src/Data/IntMap.hs
treeowl: I think we can probably improve further, but we don't have to do it now.

treeowl (approved):

treeowl:
Thanks, and thanks for your patience.

Issue #870, opened 2022-11-13
Change in complexity of Set/Map operations


meooow25:
Why does #830 change the documented complexity of Set/Map operations from $O\bigl(m \log \bigl(\frac{n}{m} + 1 \bigr)\bigr)$ to $O\bigl(m \log\bigl(\frac{n+1}{m+1}\bigr)\bigr)$?

treeowl:
Edge cases. When $m=0$, it's not nice to blow up. If you have a better way to suggest, please go ahead.

meooow25:
I don't think that's a concern in big-O notation.
$O(\log n)$ is a common complexity and $\log 0$ is undefined.

treeowl:
The tricky thing is that there are two variables. Readers will want to know what happens as $n$ increases without bound when $m=0$.

meooow25:
> Readers will want to know what happens as $n$ increases without bound when $m = 0$

That's not something the big-O complexity can be used for. It is only a bound on how fast the function grows with respect to its inputs as the inputs grow arbitrarily large.

A more reasonable thing a reader might want, perhaps, is to check the complexity for two equal sized sets. With $n = m$, the old complexity gives the value of $O(n)$, and the new one incorrectly gives $O(1)$.

In any case, I think it would be best for documentation to stick to the complexity published in the paper it cites as the source for it. I hope you'll consider changing it back.

tomsmeding:
For $n = m$, the current formula, which is $O(m \log(\tfrac{n+1}{m+1}))$, actually yields $O(0)$, not $O(1)$, which is even worse in a way: $O(1)$ means "constant-time", whereas $O(0)$ means "no time at all".

Admittedly the original formula, namely $O(m \log(\tfrac nm + 1))$, blew up at $m = 0$, which is as sensible an input as $m = n$, but this formula has the advantage of being what the original paper has.

At the very least, the side condition that $m \leq n$ in the current docs should be changed to $m < n$ if the $\tfrac{n+1}{m+1}$ formula is kept &mdash; surely the formula is wrong for $m = n$. I would vote for " $O\bigl(m \log\bigl(\tfrac nm + 1\bigr)\bigr)$, $0 < m \leq n$ " because that at least is valid in the domain that it claims to be valid on. :)

@meooow25:

> > Readers will want to know what happens as increases without bound when $m = 0$
>
[...]

PR #958, opened 2023-07-09
Revert union, difference, etc. complexity changes

tomsmeding:
As discussed in #870, the complexities for union, difference, etc. on Set and Map were changed in #830 to fix some partiality in the expressions, but along the way new partiality was introduced, and useful special cases like m = 1 get incorrect complexity values from the new formulas.

The original formula was as stated in the original paper:
  https://dl.acm.org/doi/10.1145/322123.322127
and this holds for 0 < m <= n, which seems sufficient to me. (The m=0 case is excluded, but for m=0 nothing needs to be done anyway. This contrary to the m=1 case, in which useful work with very specific complexity (namely, O(log(n))) needs to be done.)

This commit reverts all occurrences of the modified complexity formula back to the original one.

NOTE: I have not checked that all functions for which I changed the complexity specification are indeed instances of the algorithms in the original paper. However, due to the quite specific complexity formula that is unlikely to occur anywhere else, I think this is correct anyway.

### DIFF
--- a/containers/src/Data/Map/Internal.hs
+++ b/containers/src/Data/Map/Internal.hs
@@ -1811,7 +1811,7 @@ unionsWith f ts
 {-# INLINABLE unionsWith #-}
 #endif
 
--- | \(O\bigl(m \log\bigl(\frac{n+1}{m+1}\bigr)\bigr), \; m \leq n\).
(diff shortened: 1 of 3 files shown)


treeowl:
For now, we'll go with it. I would rather have something more complete.

PR #959, opened 2023-07-09
Move pattern synonym under cpp guard

alexfmpe:
Should address https://github.com/haskell/containers/pull/953#discussion_r1257348718


meooow25 (approved):

meooow25:
Thanks!

PR #961, opened 2023-07-10
Map/IntMap: Document the most common O(n²) performance mistake

nh2:
This documents the single most common performance mistake when dealing with `containers`.

I have seen professional Haskellers, get it wrong or wonder about it countless times, and the only hint in the entire module space on how the function `f :: a -> a -> a` is applied is on the docs of `insertWith`, which is not where anybody looks when checking the docs of any of these functions.

### DIFF
--- a/containers/src/Data/IntMap/Strict/Internal.hs
+++ b/containers/src/Data/IntMap/Strict/Internal.hs
@@ -425,6 +425,8 @@ insert !k !x t =
 -- > insertWith (++) 5 "xxx" (fromList [(5,"a"), (3,"b")]) == fromList [(3, "b"), (5, "xxxa")]
 -- > insertWith (++) 7 "xxx" (fromList [(5,"a"), (3,"b")]) == fromList [(3, "b"), (5, "a"), (7, "xxx")]
 -- > insertWith (++) 5 "xxx" empty                         == singleton 5 "xxx"
+--
+-- Also see the performance note on 'fromListWith'.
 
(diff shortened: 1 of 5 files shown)


containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -1095,10 +1107,41 @@ fromList xs
...
+-- > fromListWith (++) [(5,"a"), (5,"b"), (3,"x"), (5,"c")] == fromList [(3, "x"), (5, "cba")]
 -- > fromListWith (++) [] == empty
+--
+-- Note the reverse ordering of @"cba"@ in the example.
+--
+-- The symmetric combining function @f@ is applied in a left-fold over the list, as @f new old@.
+--
+-- === Performance
+--
+-- You better ensure that the given @f@ is fast with this order of arguments.
```
treeowl, line 1121: This language looks a bit casual.

treeowl (changes requested):

containers/src/Data/IntMap/Strict/Internal.hs
nh2: What would be preferred, `You should ensure`, `The caller should ensure`?

treeowl: Either of those would be okay.

nh2: Done.

nh2:
Please have another look -- I would like to merge it before any merge conflicts arise.

treeowl:
Sorry this took so long. Thanks.

PR #962, opened 2023-07-14
Tests and benchmarks for {Set,Map}.fromAscList and friends

meooow25:
I wrote these up while working on making `fromAscList` fuse well. But I'm seeing some odd behavior there, so I thought I would split this straightforward PR out and work on the rest afterwards.

Current benchmarks on GHC 9.2.5
Set:
```
  fromAscList:  OK (0.21s)
    50.1 μs ± 2.7 μs, 240 KB allocated, 2.4 KB copied, 8.0 MB peak memory
  fromDescList: OK (0.21s)
    49.1 μs ± 4.1 μs, 240 KB allocated, 2.4 KB copied, 8.0 MB peak memory
```
Map:
```
  fromAscList:         OK
    39.6 μs ± 2.7 μs, 343 KB allocated, 5.2 KB copied,  10 MB peak memory
  fromAscListWithKey:  OK
    47.0 μs ± 4.3 μs, 455 KB allocated,  12 KB copied,  10 MB peak memory
  fromDescList:        OK
    39.6 μs ± 2.6 μs, 343 KB allocated, 5.5 KB copied,  10 MB peak memory
  fromDescListWithKey: OK
    48.1 μs ± 3.2 μs, 455 KB allocated,  13 KB copied,  10 MB peak memory
```

### DIFF
--- a/containers-tests/tests/map-properties.hs
+++ b/containers-tests/tests/map-properties.hs
@@ -28,6 +28,8 @@ import qualified Prelude
 
 import Data.List (nub,sort)
 import qualified Data.List as List
+import Data.List.NonEmpty (NonEmpty(..))
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Set as Set
 import Test.Tasty
 import Test.Tasty.HUnit
@@ -185,9 +187,13 @@ main = defaultMain $ testGroup "map-properties"
          , testProperty "unionWithKeyMerge"   prop_unionWithKeyMerge
          , testProperty "mergeWithKey model"   prop_mergeWithKeyModel
          , testProperty "mergeA effects"       prop_mergeA_effects
-         , testProperty "fromAscList"          prop_ordered
+         , testProperty "fromAscList"          prop_fromAscList
+         , testProperty "fromAscListWith"      prop_fromAscListWith
+         , testProperty "fromAscListWithKey"   prop_fromAscListWithKey
          , testProperty "fromDistinctAscList"  prop_fromDistinctAscList
-         , testProperty "fromDescList"         prop_rev_ordered
+         , testProperty "fromDescList"         prop_fromDescList
+         , testProperty "fromDescListWith"     prop_fromDescListWith
+         , testProperty "fromDescListWithKey"  prop_fromDescListWithKey
          , testProperty "fromDistinctDescList" prop_fromDistinctDescList
          , testProperty "fromList then toList" prop_list
          , testProperty "toDescList"           prop_descList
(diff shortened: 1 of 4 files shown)


containers-tests/tests/map-properties.hs
treeowl, line 1: Some explanations are needed for these changes.

meooow25: Which ones are you referring to?

treeowl: I see a bunch of new code. I don't know what it does/what it tests. What's all that gunk with `head` do? Is there a cleaner way to express whatever that is?

meooow25: The tests match the current implementations against the equivalent sequence we would get if we did the job using `List.groupBy`. The only usage of `head` is on lists out of `groupBy` (which are known to be non-empty).

meooow25: Updated with an explanatory comment and switched to `NonEmpty.groupBy` and `NonEmpty.head`. Is this better?

meooow25:
I would like to pick this and related PRs back up and try to get it over the finish line. @treeowl, hope you are well. Would you be available to review these?

treeowl:
I will be available to do it on Monday. Please ping me then if you remember.

meooow25:
Rebased and updated. @treeowl please take a look.

containers-tests/tests/map-properties.hs
```diff
@@ -1222,31 +1228,68 @@ prop_mergeA_effects xs ys
...
+    List.map (\ys -> (fst (NE.head ys), NE.map snd ys)) .
+    NE.groupBy ((==) `on` fst)
 
 prop_list :: [Int] -> Bool
 prop_list xs = (sort (nub xs) == [x | (x,()) <- toList (fromList [(x,()) | x <- xs])])
 
 prop_descList :: [Int] -> Bool
 prop_descList xs = (reverse (sort (nub xs)) == [x | (x,()) <- toDescList (fromList [(x,()) | x <- xs])])
 
+prop_fromDescList :: [(Int, A)] -> Property
```
treeowl, line 1256: It's not at all clear to me what this is testing. Could you explain? Is there a simpler way to express the idea?

treeowl (changes requested):

containers-tests/tests/map-properties.hs
meooow25: This tests that given a list in decreasing order (`down_sort_xs`), the sequence we get from `fromDescList` is the same as the reverse of the sequence we would get if we group by the key and take the last key-value pair in each group.

I wouldn't mind changing the code if there is a simpler way to express this.

treeowl: Mostly need comments to explain what is being tested. Also: what's with the `()`s above? Those look mysterious.

meooow25: Added some comments.

> Also: what's with the `()`s above?

I haven't introduced those tests. It seems they want to test that `toList . fromList` sorts and nubs the keys.

treeowl: Oh, sorry, misread.

meooow25: Is it better with the comments?

meooow25: I'm going to rework the map tests to just test against `fromList*`.  
Currently the test is independent of other map functions, which will no longer be true, but it should be easier to understand what is being tested.

meooow25: Done. @treeowl let me know if you like this better.

treeowl:
Why drop `Magma`? This whole PR slipped under my radar. I'll try to take a look shortly.

containers-tests/tests/map-properties.hs
```diff
@@ -1272,31 +1278,46 @@ instance Arbitrary WhenMatchedSpec where
...
-prop_fromDistinctDescList xs =
+prop_fromDistinctDescList kxs =
     valid t .&&.
-    toList t === nub_sort_xs
+    toList t === reverse nubDownSortedKxs
   where
-    t = fromDistinctDescList (reverse nub_sort_xs)
-    nub_sort_xs = List.map List.head $ List.groupBy ((==) `on` fst) $ List.sortBy (comparing fst) xs
+    nubDownSortedKxs =
+      List.map NE.head $
```
treeowl, line 1317: I know you didn't do it, but `map head q` is bletcherous. Why not `[x | x : _ <- q]` instead?

meooow25:
> Why drop `Magma`?

Switched to arbitrary functions, makes things simpler.

It had also accumulated a bunch of edits, so I squashed a whole thing. You can take a fresh look.

containers-tests/tests/map-properties.hs
meooow25: What's wrong with that? It's `NonEmpty`'s `head` btw, not list.

treeowl: Ah... sorry ....

meooow25:
I'm planning to merge this soon so I can implement https://github.com/haskell/containers/pull/963#issuecomment-2323026386 which will use these.

Issue #951, opened 2023-06-18
Fusible IntSet.fromDistinctAscList definition


meooow25:
Sister issue to #949. I adapted the old algorithm from #658 into a fold to see if it improves things for IntSet like it did for Set.

```hs
fromDistinctAscList :: [Key] -> IntSet
fromDistinctAscList = finish . Foldable.foldl' next StateEmpty
  where
    next StateEmpty z = State (prefixOf z) (bitmapOf z) Nada
    next (State kx bm stk) z
      | kx == pz  = State pz (bm .|. bmz) stk
      | otherwise = State pz bmz (linkTop (branchMask z kx) kx (Tip kx bm) stk)
      where
        pz = prefixOf z
        bmz = bitmapOf z
    finish StateEmpty = Nil
    finish (State kx bm stk0) = linkAll kx (Tip kx bm) stk0
{-# INLINE fromDistinctAscList #-}

linkTop :: Mask -> Prefix -> IntSet -> Stack -> Stack
linkTop !m !px !tx (Push py ty stk)
  | shorter m mxy = linkTop m pxy (Bin pxy mxy ty tx) stk
  where
    mxy = branchMask px py
    pxy = mask px mxy
linkTop _ px tx stk = Push px tx stk
{-# INLINABLE linkTop #-}

linkAll :: Prefix -> IntSet -> Stack -> IntSet
linkAll !_ tx Nada = tx
linkAll px tx (Push py ty stk)
  | m < 0     = Bin p m tx ty
  | otherwise = linkAll p (Bin p m ty tx) stk
  where
    m = branchMask px py
    p = mask px m
{-# INLINABLE linkAll #-}

data FromDistinctMonoState = State {-# UNPACK #-} !Prefix {-# UNPACK #-} !BitMap !Stack | StateEmpty
data Stack = Push {-# UNPACK #-} !Prefix !IntSet !Stack | Nada
```

-----

Benchmarking 4 scenarios

```hs
  bench "fromDistinctAscList:dense" $ whnf IS.fromDistinctAscList elems -- elems = [1..n]
, bench "fromDistinctAscList:dense:fusion" $
    whnf (\n -> IS.fromDistinctAscList [1..n]) bound
, bench "fromDistinctAscList:sparse" $ whnf IS.fromDistinctAscList elems_sparse -- elems_sparse = map (*64) [1..n]
, bench "fromDistinctAscList:sparse:fusion" $
[...]

PR #965, opened 2023-07-18
Add Data.IntSet.fromRange

### LINKED ISSUE #632: IntSet range construction
`fromDistinctAscList [a,b..c]` takes time proportional to the number of elements. We should be able to write a specialized function that does better when `b-a` is less than 58 or so, and significantly better when it is very small. The idea is to fill up a whole `Tip` in one go.

meooow25:
For #632

Benchmark on GHC 9.2.5:
```
  fromRange:       OK (0.20s)
    376  ns ±  22 ns, 4.0 KB allocated,   2 B  copied, 7.0 MB peak memory
  fromRange:small: OK (0.18s)
    10.0 ns ± 872 ps, 119 B  allocated,   0 B  copied, 7.0 MB peak memory
```

Comparing `fromDistinctAscList [1..2^12]` from #951 against `fromRange (1,2^12)`, currently it takes `24.3 μs` (~62x the time) and with fusion it would still take `4.49 μs` (~11x the time).

### DIFF
--- a/containers/src/Data/IntSet/Internal.hs
+++ b/containers/src/Data/IntSet/Internal.hs
@@ -123,6 +123,7 @@ module Data.IntSet.Internal (
     -- * Construction
     , empty
     , singleton
+    , fromRange
     , insert
     , delete
     , alterF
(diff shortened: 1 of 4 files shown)


containers/src/Data/IntSet/Internal.hs
```diff
@@ -1215,6 +1216,48 @@ fromList xs
   where
     ins t x  = insert x t
 
+-- | \(O(n / W)\). Create a set from a range of integers.
+--
+-- > fromRange (l,h) == fromList [l..h]
```
konsumlamm, line 1221:
```suggestion
-- | \(O(n / W)\). Create a set from a range of integers.
--
-- > fromRange (low, high) == fromList [low..high]
--
-- If @low > high@, @fromRange (low, high)@ is empty.
```

meooow25: Changed to `low`, `high` 👍 

> `If @low > high@, @fromRange (low, high)@ is empty`

I don't think this adds value though because this is already conveyed by the line above. `[low..high]` is empty if `low` > `high`.

konsumlamm: > > `If @low > high@, @fromRange (low, high)@ is empty`
> 
> I don't think this adds value though because this is already conveyed by the line above. `[low..high]` is empty if `low` > `high`.

I thought it might be good to point out the special case, even if it's redundant, but I have no strong feelings on this.

meooow25:
Note: I've made some changes to the algorithm. If you didn't see the previous version you can ignore this comment. I made the change because I realized that we could be forced to do $O(W)$ work even when $n$ is very small, making the complexity $O(n/W + W)$. I changed `goL` and `goR` to avoid this, so it's correctly $O(n/W)$ now. Also fixed a bug where I used `bitmapOfSuffix` instead of `bitmapOf`, the tests ran fine because on x86 the shift value is masked to 5/6 bits anyway.

treeowl:
Cool!

PR #969, opened 2023-09-06
Future-proof testsuite against potential Prelude.foldl' 

TeofilC:
This is a follow-up to https://github.com/haskell/containers/pull/948.
That only dealt with the library. This also makes the test suites compatible


treeowl:
Thanks!

PR #970, opened 2023-09-09
Import `Data.Array.ST` instead of `Data.Array.ST.Safe`

konsumlamm:
`Data.Array.ST` exports the same things as `Data.Array.ST.Safe`, the latter was just added as a migration strategy. The `*.Safe` modules in `array` might get deprecated in the future, see https://gitlab.haskell.org/ghc/packages/array/-/issues/11.

### DIFF
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -108,11 +108,11 @@ import Utils.Containers.Internal.Prelude
 import Prelude ()
 #if USE_ST_MONAD
 import Control.Monad.ST
-import Data.Array.ST.Safe (newArray, readArray, writeArray)
+import Data.Array.ST (newArray, readArray, writeArray)
 # if USE_UNBOXED_ARRAYS
-import Data.Array.ST.Safe (STUArray)
+import Data.Array.ST (STUArray)
 # else
-import Data.Array.ST.Safe (STArray)
+import Data.Array.ST (STArray)
 # endif
 #else
 import Data.IntSet (IntSet)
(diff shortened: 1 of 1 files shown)


treeowl:
There seems to be an incorrect assumption somewhere; CI is failing.

meooow25:
On Hackage I can see that `Data.Array.ST` is "Safe-Inferred" [0.5.5.0](https://hackage.haskell.org/package/array-0.5.5.0/docs/Data-Array-ST.html) onwards. This is [included in GHC 9.6](https://downloads.haskell.org/ghc/9.6.1/docs/users_guide/9.6.1-notes.html#included-libraries). Doesn't that mean it should compile fine without any changes with GHC 9.6 and above?

meooow25:
Hi @konsumlamm, I'm wondering if we want to finish this.

> The next step would be to mark `Data.Array.ST` as `Trustworthy`.

Alternately we can just mark `Data.Graph` as `Trustworthy` and resolve it here on the `containers` side. I think that would be fine.

konsumlamm:
I marked `Data.Graph` as trustworthy for `array < 0.5.7.0` (for `array >= 0.5.7.0`, it's inferred as safe).

meooow25:
Thanks! I would be fine with not having the CPP and making it always Trustworthy. Or does GHC complain if that's the case?

konsumlamm:
> Thanks! I would be fine with not having the CPP and making it always Trustworthy. Or does GHC complain if that's the case?

That makes the CI fail due to `-Werror`, since it's inferred as safe on newer versions.

meooow25 (approved):
Ah that's a bit annoying, let's keep it like this then.

I think you need to mark the PR as non-draft before I can merge it.

meooow25:
Ah, I am able to do that too.

@konsumlamm just checking if you plan to make any more changes here? Seems good to merge to me.

konsumlamm:
Feel free to merge.

meooow25:
Thanks!

Issue #866, opened 2022-11-03
Add `filterKeys`


emilypi:
Hi All,

A more general form of `restrictKeys` would be `filterKeys`, implemented thusly: 

```haskell
filterKeys :: (k -> Bool) -> Map k a -> Map k a 
filterKeys p = filterWithKey (const . p)
```

Which would allow `restrictKeys` to be implemented simply as 

```haskell
restrictKeys :: Set k -> Map k a -> Map k a 
restrictKeys s = filterKeys (`Set.member` s)
```

Is there an appetite for this kind of change? `filterKeys` is a pretty generally useful function, and `restrictKeys` tends to be a little too specialized too suit my needs. Since the change would arguably improve the existing functions, I think it's worth a thought.

treeowl:
`restrictKeys` exists because it can be particularly efficiently.

emilypi:
alright, but what about the addition of `filterKeys` then?

treeowl:
I don't feel strongly about it. I guess it adds symmetry, balancing `filter`.

emilypi:
I'll go do the needful then. Additionally, I'm finding somethings are used internally, with full haddocks that would be useful to export, but are not (e.g. `filterWithKeyA`). Can we export those as well?

treeowl:
`filterWithKeyA` would be a good one, yeah. Maybe some more `mergeA` machinery too.

emilypi:
I'll raise separate PRs to not dirty the history

PR #972, opened 2023-09-23
Add filterKeys for Map and IntMap

flip111:
https://github.com/haskell/containers/issues/866


treeowl:
Please don't include your Stack files. While I won't *require* it, I would prefer to see validity tests and also property tests (with validity tests) added both for `filterKeys` and `filterWithKey`. Also, CI is failing.

flip111:
Hello. The stack files i committed accidentally in my last commit. I will try to fix the test again

meooow25:
We have
```hs
filter p m = filterWithKey (\_ x -> p x) m
```
So why not
```hs
filterKeys p m = filterWithKey (\k _ -> p k) m
```
instead of duplicating the definition of `filterWithKey`?

konsumlamm:
I don't see the advantage of this over just using `filterWithKey`?

flip111:
It's the equivalent of `filter` but then for keys

konsumlamm:
Sure, I know what it does, but why should we add it, when we can just use `filterWithKey`, which isn't any slower, instead?

flip111:
I can't speak for the original poster @emilypi but for my point of view is that haskell libraries have many convenience functions that potentially can be expressed in terms of extra functions. With `filterWithKey` you have an additional argument that is not used (when using it like `filterKeys`).

emilypi:
@flip111 that's exactly it. `filterKeys` is something many people do, and so we should optimize for it. The point is convenience. It's very low cost to add it, and i would argue it meets my personal Fairbarn threshold due to the fact that every other container library in existence features it.

konsumlamm:
> every other container library in existence features it.

Really? I don't think I've ever seen it anywhere, could you give some examples?

emilypi:
- [guava](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Maps.html#filterKeys)
- [kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-keys.html)
- [scala](https://www.scala-lang.org/api/2.13.5/scala/collection/MapView$$FilterKeys.html) - alternatively, view the [deprecated version](https://www.scala-lang.org/api/2.13.5/scala/collection/Map.html#filterKeys(p:K=%3EBoolean):scala.collection.MapView[K,V]) which used to act on the map itself.
- [purescript](https://github.com/purescript/purescript-ordered-collections/blob/v1.6.1/src/Data/Map/Internal.purs#L675-L675)
- [clojure](https://github.com/weavejester/medley/blob/master/src/medley/core.cljc#L142) - they have a general policy of only supplying O(1) functions in the core libraries (one would have to issue `(apply dissoc <f> <map>)`, but alternative cores supply this like the ones linked.
- [C++ experimental](https://en.cppreference.com/w/cpp/experimental/map/erase_if) - there are many stack overflow questiosn regarding this function, and the current state of the non-experimental art is to write a for-loop which removes keys iteratively using `std::map::erase`. 

There are many more examples, but I don't want to spend too much time on it. I don't see a valid argument against this other than "why not use `filterWithKey`", which has been addressed, and is largely a subjective for both sides. I've written this function alot, and I'd like to stop the repetition 😄

PPKFS:
This seems like something that:
- is useful
- is no added overhead
- is not some obscurely named and cryptic to understand function
- sure, it can be done already with `filterWithKey` or some equivalent but writing `filterWithKey (f . snd)` feels very strange - or should we be working in favour of removing every function that is just a more specific form of another function?

Melvar:
I wanted this function a couple weeks ago. Of course I just used `filterWithKey`, but I had to check for the absence of `filterKeys` first.

containers-tests/tests/intmap-properties.hs
treeowl, line 1: We seem to be missing property tests for `filterWithKey` and `filter` as well as the new function. The one for `filterWithKey` should look something like this:

```haskell
prop_filterWithKey :: Fun (Int, Int) Bool -> IMap -> Property
prop_filterWithKey fun m =
  valid m' .&&. toList m' === Prelude.filter (apply fun) (toList m)
  where
    m' = filterWithKey (apply2 fun) m
```

The others should be similar, with appropriate tweaks.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2578,6 +2579,17 @@ filter :: (a -> Bool) -> IntMap a -> IntMap a
 filter p m
   = filterWithKey (\_ x -> p x) m
 
+-- | \(O(n)\). Filter all keys that satisfy some predicate.
+--
+-- > filterKeys (> 4) (fromList [(5,"a"), (3,"b")]) == singleton 5 "a"
```
treeowl, line 2584: Please document the relationship to `filterWithKey`.

treeowl (changes requested):
There seem to be enough people who want this to overcome my general reluctance to add semi-unnecessary functions to this package. I want some documentation and testing tweaks for both versions, even though I only added line comments to one.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2578,6 +2579,17 @@ filter :: (a -> Bool) -> IntMap a -> IntMap a
...
+
+filterKeys :: (Key -> Bool) -> IntMap a -> IntMap a
+filterKeys predicate = go
+    where
+    go Nil           = Nil
+    go t@(Tip k _)   = if predicate k then t else Nil
+    go (Bin p m l r) = bin p m (go l) (go r)
```
emilypi, line 2591: I agree with @konsumlamm - do we gain anything here from duplicating the definition of `filterWithKey` here instead of just issuing

```haskell
filterKeys p = filterWithKey (\k _ -> p k)
```

As we do in `filter`?

containers-tests/tests/intmap-properties.hs
emilypi: If we implement `filterKeys` in terms of `filterWithKey`, that'd be great, but if we end up with the duplicate definition we currently have, I'd also want some tests for fidelity between the two a la 

```haskell
prop_filterWithKey :: Fun Int Bool -> IMap -> Property
prop_filterWithKey p m = fwk === fk 
  where
    fwk = filterWithKey (\k _ -> apply p k) m
    fk = filterKeys (apply p) m
```

containers/src/Data/IntMap/Internal.hs
konsumlamm: @meooow25 suggested implementing it using `filterWithKey`, but I agree, I don't see any reason not to use `filterWithKey` here, especially since `filter` does the same. It's also more obviously correct.

treeowl: @emilypi I imagine that depends on inlining, but we should be able to force the matter if necessary. It's a little better not to call the predicate through a partial application object. That said, I dunno how big a difference it will make.

flip111:
@treeowl @emilypi any more comments? I made some changes after your review.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2578,6 +2579,14 @@ filter :: (a -> Bool) -> IntMap a -> IntMap a
 filter p m
   = filterWithKey (\_ x -> p x) m
 
+-- | \(O(n)\). Filter all keys that satisfy some predicate.
+--
+-- > filterKeys (> 4) (fromList [(5,"a"), (3,"b")]) == singleton 5 "a"
+-- > filterKeys (> 4) == filterWithKey (\k _ -> k > 4)
+
+filterKeys :: (Key -> Bool) -> IntMap a -> IntMap a
+filterKeys predicate = filterWithKey (\k _ -> predicate k)
```
emilypi, line 2588: nice :)

emilypi:
Looks fine to me, but it's up to @treeowl

flip111:
@treeowl could you take a look please? Is there anything i can do or something that needs to be done?

containers-tests/tests/intmap-strictness.hs
```diff
@@ -87,6 +87,24 @@ pInsertLookupWithKeyValueStrict f k v m
...
+-- pFilterKeys fun m =
+--   valid m' .&&. toList m' === Prelude.filter (apply fun) (toList m)
+--   where
+--     m' = filterKeys (apply2 fun) m
+
+-- pFilter :: Fun (Int, Int) Bool -> IMap -> Property
+-- pFilter fun m =
+--   valid m' .&&. toList m' === Prelude.filter (apply fun) (toList m)
+--   where
+--     m' = filter (apply2 fun) m
```
meooow25, line 106: Did you mean to add these? Currently they are commented out.

containers-tests/tests/intmap-strictness.hs
```diff
@@ -87,6 +87,24 @@ pInsertLookupWithKeyValueStrict f k v m
                      not (isBottom $ M.insertLookupWithKey (const3 1) k bottom m)
     | otherwise    = isBottom $ M.insertLookupWithKey (apply3 f) k bottom m
 
+pFilterWithKey :: Fun (Int, Int) Bool -> IMap -> Property
+pFilterWithKey fun m =
+  valid m' .&&. toList m' === Prelude.filter (apply fun) (toList m)
+  where
+    m' = filterWithKey (apply2 fun) m
```
meooow25, line 94: Please move property tests to `intmap-properties.hs`.

containers-tests/tests/map-properties.hs
```diff
@@ -820,8 +820,11 @@ test_filter = do
     filter (> "x") (fromList [(5,"a"), (3,"b")]) @?= empty
     filter (< "a") (fromList [(5,"a"), (3,"b")]) @?= empty
 
-test_filteWithKey :: Assertion
-test_filteWithKey = filterWithKey (\k _ -> k > 4) (fromList [(5,"a"), (3,"b")]) @?= singleton 5 "a"
+test_filterKeys :: Assertion
+test_filterKeys = filterKeys (> 4) (fromList [(5,"a"), (3,"b")]) @?= singleton 5 "a"
```
meooow25, line 824: This is not used, it needs to be added to the list of tests above.

```hs
, testCase "filterKeys" test_filterKeys
```

Also, it would be good to have property tests (like for IntMap).

meooow25 (changes requested):

containers/src/Data/Map/Internal.hs
```diff
@@ -2857,6 +2858,13 @@ filter :: (a -> Bool) -> Map k a -> Map k a
 filter p m
   = filterWithKey (\_ x -> p x) m
 
+-- | \(O(n)\). Filter all keys that satisfy the predicate.
+--
+-- > filterKeys (> 4) (fromList [(5,"a"), (3,"b")]) == singleton 5 "a"
+
```
meooow25, line 2864: Please add a `@since FIXME` at the end of the docs. It will be updated by maintainers to the next release version.  
Same for IntMap.

meooow25:
@flip111 would you like to finish the work here?

containers-tests/tests/intmap-strictness.hs
flip111: Yes they are supposed to be added because of @treeowl his comment

> We seem to be missing property tests for filterWithKey and filter as well as the new function

But these function pFilterWithKey, pFilterKeys and pFilter were difficult for me to implement. The proposed implementation up above does not work, possible as simple as missing imports. I'll have another look if i can make something decent here otherwise i need help with these testing functions.

containers-tests/tests/intmap-strictness.hs
flip111: Ok i will move them (and rename to prop_)

containers-tests/tests/map-properties.hs
flip111: Ok

flip111:
Sorry i didn't do this earlier, sometimes things seem like a mountain when it's actually not that hard. It went a bit out of focus, thanks for the reminder @meooow25

flip111:
I guess now, sync with upstream master, rebase and adjust changelog?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2578,6 +2579,14 @@ filter :: (a -> Bool) -> IntMap a -> IntMap a
 filter p m
   = filterWithKey (\_ x -> p x) m
 
+-- | \(O(n)\). Filter all keys that satisfy some predicate.
+--
+-- > filterKeys (> 4) (fromList [(5,"a"), (3,"b")]) == singleton 5 "a"
+-- > filterKeys (> 4) == filterWithKey (\k _ -> k > 4)
```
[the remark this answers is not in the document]
flip111: I had put that there as example for the documentation for the users. Shall i not just keep that? I'm not familiar with `@` in a docblock, is it a test? I can add those tests and at the same time keep `prop_filterKeys` from `intmap-properties` and `map-properties`

containers-tests/tests/intmap-properties.hs
```diff
@@ -1506,6 +1514,18 @@ prop_filter p ys = length ys > 0 ==>
...
+prop_filterKeys p ys = length ys > 0 ==>
+  let xs = List.nubBy ((==) `on` fst) ys
+      m  = fromList xs
+  in  filterKeys (apply p) m == fromList (List.filter (apply p . fst) xs)
+
+prop_filterWithKey :: Fun (Int, Int) Bool -> [(Int, Int)] -> Property
+prop_filterWithKey p ys = length ys > 0 ==>
+  let xs = List.nubBy ((==) `on` fst) ys
+      m  = fromList xs
+  in  filterWithKey (\k v -> apply p (k, v)) m == fromList (List.filter (\(k, v) -> apply p (k, v)) xs)
```
meooow25, line 1527: Why change these tests? They were fine in your previous commit.

containers/src/Data/IntMap/Internal.hs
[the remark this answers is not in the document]
meooow25: No it's not a test, just plain documentation. It will look like this: https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Lazy.html#v:take

Anyway, you've added it below so please remove this line "`> filterKeys (> 4) == filterWithKey (\k _ -> k > 4)`".

meooow25 (changes requested):

meooow25:
The test is broken. It's a small change so I'm just going to go ahead and fix it.

meooow25 (approved):
Alright, I'll squash and merge this soon if there are no other concerns.

flip111:
Thank you @meooow25

meooow25:
Thanks for wrapping this up @flip111!

Issue #971, opened 2023-09-20
Release for GHC 9.8.1


bgamari:
@treeowl, on reviewing GHC's submodules I noticed that the `containers` commit used by GHC 9.8.1 has a few substantive changes relative to the most recent release, 0.6.7. Unfortunately, this means that a Hackage revision will not be sufficient and we will instead need a proper release. Do you suppose this could happen sometime in the next week or so?

If a release is not possible then I can instead bump GHC back to 0.6.7 with a bounds bump revision (namely f61b0c9104a3c436361f56a0974c5eeef40c1b89).

Apologies for noticing this so late; my earlier audit from our discussion in #955 left me thinking that the only changes were in benchmarks and tests.

treeowl:
Yes, we can make a release this week.

treeowl:
I'll be able to make the release tomorrow.

treeowl:
Released!

Bodigrim:
@treeowl something seems wrong here: f5d0b13251291c3bd1ae396f3e6c8b0b9eaf58b0 says "Bump version to 0.7", but actually bumps to 0.6.8 only: 
https://github.com/haskell/containers/blob/f5d0b13251291c3bd1ae396f3e6c8b0b9eaf58b0/containers/containers.cabal#L1-L2

Hackage release seems correct however.

treeowl:
Yes, that commit is broken. The next one fixes it, no?

bgamari:
I very much wish I had noticed this mix-up this morning. I had assumed that the `master` commit was the final release; in hindsight I should have noticed that something was amiss when I noticed that there was no tag uploaded. @treeowl, if there is a next commit I'm afraid it wasn't pushed.

Unfortunately, the GHC release pipeline which I had planned to cut the release candidate from refers to this broken commit and will need to be scrapped. This  implies that we will need to push the release candidate by at least a day. However, I fear that the situation is even worse than this as none of the boot libraries' bounds current accommodate `containers-0.7`. Fixing this would very likely take weeks.

@treeowl, I don't believe a major release of `containers` is feasible at this juncture. At this point I think we should plan to do the bare minimum to get the release out the door (ideally just a Hackage revision). I believe the only thing preventing us from using 0.6.7 is the `deepseq < 1.5` constraint. Would you be willing to lift this in a revision?

Bodigrim:
[...]

PR #973, opened 2023-09-27
Bump version to 0.7

treeowl:
For #971.


containers/containers.cabal
```diff
@@ -1,5 +1,5 @@
 name: containers
-version: 0.6.7
+version: 0.6.8
```
andreasabel, line 2: That does not feel like 0.7. 😉

PR #980, opened 2024-01-11
Perf/lazy insert ptr removal

HuwCampbell:
This is to open the conversation. I haven't looked at the strict version or any other alternatives like `insertR`.

I'm not convinced that the optimisations for identical pointers is worth the complexity, and it hurts performance for inserts of new elements and the replacement of elements compared to this version.

```
Old
  insert absent:    OK (1.46s)
    352  μs ±  27 μs
  insert same:      OK (1.63s)
    194  μs ±  16 μs
  insert alternate: OK (1.45s)
    351  μs ±  28 μs


New
  insert absent:    OK (1.27s)
    312  μs ±  29 μs
  insert same:      OK (2.00s)
    242  μs ±  13 μs
  insert alternate: OK (1.27s)
    308  μs ±  29 μs
```

### DIFF
--- a/containers-tests/benchmarks/Map.hs
+++ b/containers-tests/benchmarks/Map.hs
@@ -35,6 +35,7 @@ main = do
         , bench "alterF no rules lookup present" $ whnf (atLookupNoRules evens) m_even
         , bench "insert absent" $ whnf (ins elems_even) m_odd
         , bench "insert present" $ whnf (ins elems_even) m_even
+        , bench "insert alternate" $ whnf (ins elems_alts) m_even
         , bench "alterF insert absent" $ whnf (atIns elems_even) m_odd
         , bench "alterF insert present" $ whnf (atIns elems_even) m_even
         , bench "alterF no rules insert absent" $ whnf (atInsNoRules elems_even) m_odd
@@ -100,6 +101,7 @@ main = do
     bound = 2^12
     elems = zip keys values
     elems_even = zip evens evens
+    elems_alts = zip evens odds
     elems_odd = zip odds odds
     elems_rev = reverse elems
     keys = [1..bound]
--- a/containers/src/Data/Map/Internal.hs
+++ b/containers/src/Data/Map/Internal.hs
@@ -776,24 +776,24 @@ singleton k x = Bin 1 k x Tip Tip
 -- See Note: Type of local 'go' function
 -- See Note: Avoiding worker/wrapper
 insert :: Ord k => k -> a -> Map k a -> Map k a
-insert kx0 = go kx0 kx0
+insert kx0 ax0 m0 =
+    case go kx0 ax0 m0 of (m :*: _) -> m
   where
     -- Unlike insertR, we only get sharing here
     -- when the inserted value is at the same address
     -- as the present value. We try anyway; this condition
     -- seems particularly likely to occur in 'union'.
-    go :: Ord k => k -> k -> a -> Map k a -> Map k a
-    go orig !_  x Tip = singleton (lazy orig) x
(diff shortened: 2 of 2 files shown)


treeowl:
I'm surprised you're getting good results with explicit continuations. Do you know how GHC is transforming those?

HuwCampbell:
Yeah I couldn't really see any differences in the union benchmarks (maybe one was marginally faster?). From trying an explicit continuation on the other inserts, I expect that it is slower when there is a value inserted.

I haven't looked at the core yet.

I also considered just throwing a sentinel exception if `EQ` is found, then catching and returning the top map. It would probably be fast, but exceptions for control flow are a bit gross.

containers-tests/benchmarks/Map.hs
```diff
@@ -84,6 +86,9 @@ main = do
         , bench "mapMaybeWithKey" $ whnf (M.mapMaybeWithKey (const maybeDel)) m
         , bench "lookupIndex" $ whnf (lookupIndex keys) m
         , bench "union" $ whnf (M.union m_even) m_odd
+        , bench "union_identical" $ whnf (M.union m_even) m_even
+        , bench "union_sparse" $ whnf (M.union m_even) m_sparse
+        , bench "union_into_sparse" $ whnf (M.union m_sparse) m_even
```
HuwCampbell, line 91: Two more with `m_sparse` and `m_odd` would show if there's a slowdown in `insertR`.

treeowl:
I would expect exceptions to be slow. I think I might have tried? I don't
remember for sure. If you try, use unsafeDupablePerformIO to catch the
exception. Also consider using raise# and catch# manually, to allow an
exception type other than SomeException (this may require
reallyUnsafePtrEquality# to determine whether you got the not-present
exception).

HuwCampbell:
I'm not as concerned with `insertR` as `insert` and `insertWith`.

I guess the question is do we want a 10-15% sacrifice on insertion when the key in not already present but a 20-30% improvement when it is?

How would that decision be made?

meooow25:
> I'm not convinced that the (unsafe) optimisations for identical pointers is worth the complexity, and it hurts performance for inserts of new elements and the replacement of elements compared to this version.
 
> So inserting when the key is missing (common) or a new item at the same key (also common), are faster, while only literally inserting the same identical item (i.e., a no-op, probably less common) is slower.

This sounds reasonable. The current code seems optimized for a very rare case. We have a comment acknowledging the same: https://github.com/haskell/containers/blob/41783ed917e5a16f3025728bcdb3148e73ff723d/containers/src/Data/Map/Internal.hs#L785-L788

But why is it likely in `union`? It would happen when `union`ing two maps with a large overlap in shared entries. That again seems like a rare situation.

---

There are a few variants of `insert` we can try:

A. Pointer equality (current)
B. Return a bool which controls rebalancing (this PR)
C. No tricks
D. ~~Both A and B~~ (let's not go there)

Curiously, strict map uses C instead of A: https://github.com/haskell/containers/blob/41783ed917e5a16f3025728bcdb3148e73ff723d/containers/src/Data/Map/Strict/Internal.hs#L503-L506

And there are 3 situations that can occur:

1. It's a new key. C is the best here, A and B waste work.
2. It's an existing key but the value is new. B is the best here. How likely is this? Perhaps not too unlikely, shared key objects might be around.
3. It's an existing key and existing value. C is the best here, and B should also help. This seems rather unlikely.

---

So, there's a bunch of benchmarking involved, if you want to figure this out.  
For `union` benchmarks you can use the `set-operations-map`, it covers various sets of inputs.  
Related to the history of the current code: #315, #416. Though I don't see any new information that is not already documented.

PR #982, opened 2024-01-13
Make IntMap split, splitLookup strict in the key

meooow25:
All IntMap and IntSet functions taking a key are strict in the key. This allows the Int to be unboxed.

This is technically a breaking change. Currently `split undefined empty = (empty, empty)`, but with this change `split undefined empty = undefined`.

### DIFF
--- a/containers/changelog.md
+++ b/containers/changelog.md
@@ -1,5 +1,14 @@
 # Changelog for [`containers` package](http://github.com/haskell/containers)
 
+## Next release
+
+### Breaking changes
+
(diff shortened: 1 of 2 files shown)


meooow25:
> Could you add a change log entry please?

Done 👍 

> What were you doing that led you to notice this?

I have a structure that stores `(Int, Int)` ranges as `IntMap Int`, with some operations defined in terms of `splitLookup`. I checked the core for those operations and noticed that the `Int`s were boxed, which led me to this.

treeowl:
Awesome. Thanks!

PR #983, opened 2024-01-14
Make IntSet splitMember strict in the key

meooow25:
Currently, the key is ignored for an empty set.

Similar to #982. I didn't notice it earlier else I would have included it there.

IntSet.split is already strict.


treeowl:
Thanks!

PR #987, opened 2024-02-02
Add `flattenSCC1 :: SCC a -> NonEmpty a`

### LINKED ISSUE #985: Add `flattenSCC1 : SCC vertex -> Data.List.NonEmpty.NonEmpty vertex`
The `flattenSCC` function throws away the knowledge that a SCC is a non-empty list of vertices:
https://github.com/haskell/containers/blob/a4e8d167c0c0f0001e48f499d33735a61c68d9d4/containers/src/Data/Graph.hs#L248-L251

In course of the new `x-partial` warning of GHC 9.8, this is a bit inconvenient, since GHC will now cry about `head . flattenSCC` and similar.

The more correct type of `flattenSCC` is `SCC vertex -> Data.List.NonEmpty.NonEmpty vertex`, but in order not to break everyone's code I suggest to add a new function `flattenSCC1` instead.

Should this maybe have gone into `containers-0.7`?

andreasabel:
This gives a more precise type to the existing `flattenSCC :: SCC a -> [a]`.
Closes #985.

### DIFF
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -246,9 +246,18 @@ flattenSCCs :: [SCC a] -> [a]
 flattenSCCs = concatMap flattenSCC
 
 -- | The vertices of a strongly connected component.
+--
+-- @flattenSCC = 'Data.List.NonEmpty.toList' . 'flattenSCC1'@.
+--
+-- This function is retained for backward compatibility,
+-- 'flattenSCC1' has the more precise type.
 flattenSCC :: SCC vertex -> [vertex]
-flattenSCC (AcyclicSCC v) = [v]
-flattenSCC (NECyclicSCC vs) = NE.toList vs
+flattenSCC = NE.toList . flattenSCC1
+
+-- | The vertices of a strongly connected component.
+flattenSCC1 :: SCC vertex -> NonEmpty vertex
+flattenSCC1 (AcyclicSCC v) = v :| []
+flattenSCC1 (NECyclicSCC vs) = vs
 
 -- | \(O((V+E) \log V)\). The strongly connected components of a directed graph,
 -- reverse topologically sorted.
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -89,6 +89,7 @@ module Data.Graph (
 
     -- ** Conversion
     , flattenSCC
+    , flattenSCC1
     , flattenSCCs
 
     -- * Trees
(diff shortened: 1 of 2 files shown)


Bodigrim:
This probably needs a changelog entry and `@since` pragma, but I don't know what's the policy here and which version number is supposed to be released next.

andreasabel:
> This probably needs a changelog entry and `@since` pragma, but I don't know what's the policy here and which version number is supposed to be released next.

I can add that here once the decision has been taken to merge this PR!

treeowl:
Let's do it.

Bodigrim (approved):

andreasabel:
Ok, I added a `since` for 0.7.1.

I was a bit surprised to see `master` at 0.6.8 when it already contained the changes for 0.7.
According to the changelog, `master` also contains breaking changes which would require the next `master` release to be 0.8.

containers/src/Data/Graph.hs
```diff
@@ -246,9 +246,20 @@ flattenSCCs :: [SCC a] -> [a]
...
+-- 'flattenSCC1' has the more precise type.
 flattenSCC :: SCC vertex -> [vertex]
-flattenSCC (AcyclicSCC v) = [v]
-flattenSCC (NECyclicSCC vs) = NE.toList vs
+flattenSCC = NE.toList . flattenSCC1
+
+-- | The vertices of a strongly connected component.
+--
+-- @since 0.7.1
+flattenSCC1 :: SCC vertex -> NonEmpty vertex
```
meooow25, line 260: `flattenSCC1` needs to be exported.

containers/src/Data/Graph.hs
```diff
@@ -246,9 +246,20 @@ flattenSCCs :: [SCC a] -> [a]
 flattenSCCs = concatMap flattenSCC
```
meooow25, line 246: Could we also add `flattenSCCs1 :: [SCC a] -> NonEmpty a`? It seems appropriate.

meooow25 (changes requested):

containers/src/Data/Graph.hs
Bodigrim: Did you mean `flattenSCCs1 :: NonEmpty (SCC a) -> NonEmpty a`?

meooow25: Oh right, it doesn't work out. Scratch that.  
`NonEmpty (SCC a) -> NonEmpty a` could be added, but I doubt it will be useful because `flattenSCCs` is usually applied to the output of `stronglyConnComp`/`stronglyConnCompR`, which has to return `[SCC a]`.

containers/src/Data/Graph.hs
andreasabel: Thanks, this is fixed now.

meooow25 (approved):

meooow25:
Thanks!

PR #993, opened 2024-03-08
Fix changelog, version

treeowl:
I missed some important contributions and contributors.

Also update the version number to what is on Hackage.


treeowl:
Some git mistakes, it seems. Oops.

treeowl:
Nope; this is all b0rked....

Issue #991, opened 2024-02-07
Potential memory optimization for IntMap and IntSet


meooow25:
This isn't particularly novel, so maybe it has been proposed before. Let me know if it has. I only found #340 in my search, which is a bigger idea.

### Current definition
Today, `Bin` for `IntMap` and `IntSet` is represented as https://github.com/haskell/containers/blob/3c13e0b03bde9aab4ce1ea7b23be1c713b89d32b/containers/src/Data/IntMap/Internal.hs#L355-L362

### Potential new definition
The prefix and the mask can be merged so that we save one word per `Bin`.

```hs
data IntMap a = Bin {-# UNPACK #-} !Int -- (current Prefix + current Mask)
                    !(IntMap a)
                    !(IntMap a)
```

The mask bit is always zero in the prefix, so this isn't throwing away any information. The lowest set bit of the new int is the current mask and the rest of it is the current prefix.

### Current branching on `Bin`

Branching on `Bin` is currently done like this: https://github.com/haskell/containers/blob/3c13e0b03bde9aab4ce1ea7b23be1c713b89d32b/containers/src/Data/IntMap/Internal.hs#L813-L817
https://github.com/haskell/containers/blob/3c13e0b03bde9aab4ce1ea7b23be1c713b89d32b/containers/src/Data/IntMap/Internal.hs#L3527-L3528
https://github.com/haskell/containers/blob/3c13e0b03bde9aab4ce1ea7b23be1c713b89d32b/containers/src/Data/IntMap/Internal.hs#L3519-L3520

### New branching on `Bin`

```hs
insert :: Key -> a -> IntMap a -> IntMap a 
insert !k x t@(Bin pm l r) 
  | nomatch k pm = link k (Tip k x) p t 
  | left k pm    = Bin pm (insert k x l) r 
  | otherwise    = Bin pm l (insert k x r)
 
nomatch :: Int -> Int -> Bool
nomatch k pm = mask i pm /= pm .&. (pm-1)

left :: Int -> Int -> Bool
left k pm = int2word k < int2word pm 
```

### Performance impact

I don't know yet, I need to make the change and benchmark. And it will involve changing _every_ function, so it will be take a while.
Memory is certainly saved. `nomatch` gets a little more expensive, but `left` is cheaper than `zero`, so I'm hoping there is zero or positive overall effect.

What do you think? Is it worth checking out how this will fare? And is there any bad consequence of this representation, that I didn't think of?

meooow25:
[...]

PR #994, opened 2024-03-08
Merge 0.7

treeowl:

PR #995, opened 2024-03-10
Optimize IntMap.Bin

meooow25:
Replaces the separate `Prefix` and `Mask` fields in the `Bin` constructor with a single field that contains both.

This reduces the memory required by a `Bin` from 5 to 4 words, at the cost of more computations (which are cheap bitwise ops) being necessary for certain operations.

Implements #991 for `IntMap` only. The same can be done for `IntSet` if this is accepted.

### Memory

Concretely, this reduces the memory required by an `IntMap` by ~12.5% (including the keys and values).

Calculations: For a map with `n` key-values, there are `n` `Tip`s costing 3 words each and `n-1` `Bin`s costing 5 words each before this change and 4 words after. So we save about 1 out of 8 words per key-value.

### Compatibility

This PR, in its current state, makes breaking changes to IntMap internals which is reflected in the exports of the internal module `Data.IntMap.Internal`. There is no change in the exports of any other module.

### Benchmarks

Benchmarks done with GHC 9.6.3.
Last updated: 7e2ca1c
The output is  from a hacky script I put together that compares tasty-bench's csv files.

Benchmark command: `cabal run <target> -- --csv <csv> +RTS -T`

`intmap-benchmarks`

```
Name                          Time - - - - - - - -    Allocated - - - - -     Copied - - - - - - -
                                   A       B     %         A       B     %         A       B     %
alter                         472 μs  421 μs  -10%    2.2 MB  1.8 MB  -18%     71 KB   50 KB  -30%
delete                        103 μs   94 μs   -9%    956 KB  763 KB  -20%    150 B   116 B   -22%
foldlWithKey                  434 μs  438 μs   +0%    2.1 MB  1.7 MB  -19%     64 KB   49 KB  -23%
foldlWithKey'                  12 μs   12 μs   -2%      0 B     0 B             0 B     0 B
foldrWithKey                   23 ns   24 ns   +3%    463 B   463 B    +0%      0 B     0 B
fromAscList                    49 μs   46 μs   -6%    256 KB  223 KB  -12%     14 KB  6.1 KB  -56%
fromDistinctAscList            46 μs   43 μs   -5%    256 KB  223 KB  -12%     14 KB  6.1 KB  -56%
fromList                      114 μs   97 μs  -15%    1.0 MB  861 KB  -18%     36 KB   26 KB  -28%
insert                        111 μs   99 μs  -10%    1.0 MB  861 KB  -18%     36 KB   25 KB  -28%
insertLookupWithKey empty     506 μs  527 μs   +4%    4.0 MB  3.7 MB   -8%    153 KB  112 KB  -26%
insertLookupWithKey update    1.7 ms  1.7 ms   +2%    9.6 MB  8.8 MB   -8%    1.1 MB  1.0 MB   -9%
insertWith empty              116 μs  106 μs   -8%    1.0 MB  863 KB  -18%     36 KB   26 KB  -28%
insertWith update             483 μs  444 μs   -8%    2.3 MB  1.9 MB  -17%    133 KB   93 KB  -29%
insertWith' empty             123 μs  113 μs   -8%    1.0 MB  861 KB  -18%     36 KB   25 KB  -29%
insertWith' update            487 μs  445 μs   -8%    2.2 MB  1.8 MB  -17%     90 KB   65 KB  -28%
insertWithKey empty           118 μs  106 μs  -10%    1.0 MB  863 KB  -17%     36 KB   26 KB  -27%
insertWithKey update          477 μs  446 μs   -6%    2.3 MB  1.9 MB  -17%    120 KB   96 KB  -20%
insertWithKey' empty          124 μs  110 μs  -11%    1.0 MB  861 KB  -18%     36 KB   26 KB  -27%
insertWithKey' update         490 μs  447 μs   -8%    2.2 MB  1.8 MB  -17%     90 KB   67 KB  -24%
lookup_half                   164 μs  110 μs  -33%      0 B     0 B             0 B     0 B
lookup_hits                   165 μs  161 μs   -2%      0 B     0 B             0 B     0 B
lookup_misses                 164 μs  162 μs   -1%      0 B     0 B             0 B     0 B
lookup_mixed                   17 μs   22 μs  +30%      0 B     0 B             0 B     0 B
lookup_most                   157 μs  149 μs   -5%      0 B     0 B             0 B     0 B
map                            31 μs   29 μs   -5%    351 KB  319 KB   -9%     15 KB   13 KB  -16%
mapMaybe                       55 μs   56 μs   +2%    275 KB  255 KB   -6%    5.9 KB  4.9 KB  -16%
mapMaybeWithKey                54 μs   56 μs   +3%    277 KB  255 KB   -7%    5.8 KB  5.0 KB  -15%
mapWithKey                     37 μs   35 μs   -4%    414 KB  383 KB   -7%     20 KB   18 KB  -12%
minView                        45 ns   32 ns  -30%    119 B   103 B   -13%      0 B     0 B
spanAntitone                  112 ns  125 ns  +11%    749 B   655 B   -12%      0 B     0 B
split                          47 ns   45 ns   -3%    486 B   399 B   -17%      0 B     0 B
splitLookup                    50 ns   48 ns   -3%    511 B   422 B   -17%      0 B     0 B
update                        441 μs  420 μs   -4%    2.1 MB  1.7 MB  -17%     68 KB   48 KB  -28%
updateLookupWithKey           901 μs  826 μs   -8%    4.4 MB  3.6 MB  -16%    167 KB  100 KB  -40%
```

`set-operations-intmap`

```
Name                           Time - - - - - - - -    Allocated - - - - -     Copied - - - - - - -
                                    A       B     %         A       B     %         A       B     %
difference-block_nn             25 μs   22 μs  -10%     88 KB   71 KB  -19%    1.0 KB  666 B   -34%
difference-block_nn_swap        25 μs   22 μs   -9%     88 KB   71 KB  -19%    977 B   660 B   -32%
difference-block_ns            2.2 μs  2.0 μs   -8%     13 KB   11 KB  -20%     28 B    16 B   -42%
difference-block_sn_swap       1.9 μs  1.7 μs   -7%    8.8 KB  7.1 KB  -19%     12 B     8 B   -33%
difference-common_nn           1.1 ms  903 μs  -14%    1.9 MB  1.5 MB  -19%    416 KB  297 KB  -28%
difference-common_nn_swap      465 μs  437 μs   -6%      0 B     0 B             0 B     0 B
difference-common_ns           361 μs  313 μs  -13%    1.2 MB  941 KB  -21%    173 KB  107 KB  -38%
difference-common_nt            33 μs   28 μs  -13%    102 KB   81 KB  -20%    1.3 KB  841 B   -36%
difference-common_sn_swap      153 μs  134 μs  -12%      0 B     0 B             0 B     0 B
difference-common_tn_swap       14 μs   12 μs  -19%     44 B    54 B   +22%      0 B     0 B
difference-disj_nn              57 ns   55 ns   -3%    294 B   247 B   -15%      0 B     0 B
difference-disj_nn_swap         64 ns   64 ns   +0%    471 B   383 B   -18%      0 B     0 B
difference-disj_ns              53 ns   51 ns   -4%    294 B   247 B   -15%      0 B     0 B
difference-disj_nt              49 ns   45 ns   -7%    294 B   247 B   -15%      0 B     0 B
difference-disj_sn_swap         58 ns   56 ns   -2%    390 B   319 B   -18%      0 B     0 B
difference-disj_tn_swap         47 ns   47 ns   +0%    271 B   223 B   -17%      0 B     0 B
difference-mix_nn              6.1 ms  4.0 ms  -35%    6.1 MB  5.2 MB  -14%     10 MB  5.3 MB  -47%
difference-mix_nn_swap         5.8 ms  4.6 ms  -19%    6.0 MB  5.3 MB  -11%    9.6 MB  6.6 MB  -31%
difference-mix_ns              322 μs  279 μs  -13%    1.3 MB  1.0 MB  -20%    205 KB  138 KB  -32%
difference-mix_nt               24 μs   21 μs  -14%    102 KB   81 KB  -20%    1.3 KB  883 B   -34%
difference-mix_sn_swap         191 μs  175 μs   -8%    623 KB  543 KB  -12%     47 KB   36 KB  -23%
difference-mix_tn_swap          12 μs  9.2 μs  -23%     19 KB   17 KB  -12%     51 B    41 B   -19%
intersection-block_nn           13 μs   11 μs  -13%      0 B     0 B             0 B     0 B
intersection-block_nn_swap      13 μs   11 μs  -10%      0 B     0 B             0 B     0 B
intersection-block_ns          1.3 μs  1.3 μs   -4%     25 B     0 B   -100%      0 B     0 B
intersection-block_sn_swap     1.2 μs  1.2 μs   -1%      0 B     0 B             0 B     0 B
intersection-common_nn         916 μs  761 μs  -16%    1.9 MB  1.5 MB  -20%    440 KB  299 KB  -31%
intersection-common_nn_swap    880 μs  759 μs  -13%    1.9 MB  1.5 MB  -20%    439 KB  299 KB  -31%
intersection-common_ns         183 μs  176 μs   -3%    354 KB  278 KB  -21%     15 KB  9.9 KB  -32%
intersection-common_nt          13 μs   14 μs  +13%     12 KB  9.7 KB  -18%     22 B    13 B   -40%
intersection-common_sn_swap    186 μs  171 μs   -8%    351 KB  278 KB  -20%     15 KB  9.9 KB  -31%
intersection-common_tn_swap     13 μs   14 μs   +3%     12 KB  8.8 KB  -28%     21 B    13 B   -38%
intersection-disj_nn            40 ns   36 ns   -9%     31 B    31 B    +0%      0 B     0 B
intersection-disj_nn_swap       38 ns   40 ns   +4%     31 B    31 B    +0%      0 B     0 B
intersection-disj_ns            35 ns   33 ns   -5%     31 B    31 B    +0%      0 B     0 B
intersection-disj_nt            32 ns   31 ns   -2%     31 B    31 B    +0%      0 B     0 B
intersection-disj_sn_swap       34 ns   37 ns   +7%     31 B    31 B    +0%      0 B     0 B
intersection-disj_tn_swap       31 ns   32 ns   +3%     31 B    31 B    +0%      0 B     0 B
intersection-mix_nn            857 μs  769 μs  -10%      0 B     0 B             0 B     0 B
intersection-mix_nn_swap       823 μs  794 μs   -3%      0 B     0 B             0 B     0 B
intersection-mix_ns            127 μs  117 μs   -7%      0 B     0 B             0 B     0 B
intersection-mix_nt            7.8 μs  6.3 μs  -19%      0 B    31 B             0 B     0 B
intersection-mix_sn_swap       121 μs  113 μs   -6%      0 B     0 B             0 B     0 B
intersection-mix_tn_swap       8.7 μs  7.7 μs  -11%      0 B     0 B             0 B     0 B
union-block_nn                  33 μs   29 μs  -12%    177 KB  142 KB  -19%    3.9 KB  2.5 KB  -35%
union-block_nn_swap             33 μs   28 μs  -13%    178 KB  142 KB  -20%    3.9 KB  2.5 KB  -35%
union-block_ns                 2.7 μs  2.4 μs  -10%     22 KB   18 KB  -20%     69 B    44 B   -36%
union-block_sn_swap            2.7 μs  2.5 μs  -10%     22 KB   18 KB  -20%     69 B    42 B   -39%
union-common_nn                1.6 ms  1.3 ms  -18%    3.8 MB  3.0 MB  -21%    1.8 MB  1.2 MB  -35%
union-common_nn_swap           1.6 ms  1.3 ms  -16%    3.8 MB  3.0 MB  -21%    1.7 MB  1.2 MB  -32%
union-common_ns                388 μs  353 μs   -8%    1.5 MB  1.2 MB  -19%    288 KB  189 KB  -34%
union-common_nt                 30 μs   27 μs  -11%    114 KB   91 KB  -20%    1.6 KB  1.0 KB  -36%
union-common_sn_swap           387 μs  338 μs  -12%    1.5 MB  1.2 MB  -20%    286 KB  188 KB  -34%
union-common_tn_swap            31 μs   25 μs  -17%    114 KB   91 KB  -20%    1.6 KB  1.0 KB  -36%
union-disj_nn                   80 ns   64 ns  -19%    750 B   607 B   -19%      0 B     0 B
union-disj_nn_swap              78 ns   73 ns   -6%    748 B   607 B   -18%      0 B     0 B
union-disj_ns                   72 ns   57 ns  -19%    671 B   543 B   -19%      0 B     0 B
union-disj_nt                   60 ns   48 ns  -19%    549 B   447 B   -18%      0 B     0 B
union-disj_sn_swap              76 ns   73 ns   -3%    671 B   541 B   -19%      0 B     0 B
union-disj_tn_swap              61 ns   56 ns   -8%    550 B   447 B   -18%      0 B     0 B
union-mix_nn                   9.2 ms  6.5 ms  -28%    7.6 MB  6.1 MB  -19%     17 MB  9.8 MB  -41%
union-mix_nn_swap              9.2 ms  6.3 ms  -31%    7.6 MB  6.0 MB  -20%     17 MB  9.3 MB  -44%
union-mix_ns                   431 μs  384 μs  -10%    1.7 MB  1.3 MB  -19%    352 KB  268 KB  -23%
union-mix_nt                    29 μs   24 μs  -18%    114 KB   91 KB  -20%    1.6 KB  1.0 KB  -37%
union-mix_sn_swap              423 μs  341 μs  -19%    1.7 MB  1.3 MB  -19%    352 KB  247 KB  -29%
union-mix_tn_swap               28 μs   23 μs  -16%    114 KB   91 KB  -20%    1.6 KB  1.0 KB  -35%
```

Allocations have gone down in every benchmark, which is expected but quite pleasant to see.
Runtime has gone down in most benchmarks and gone up in a few.

#### Notable time regressions

The only significant regressions (>10%) I see are in the `union-disj` and `intersection-disj` set operations. However I don't consider this a problem because these operations on the specific disjoint maps used in the tests wrap up in $O(W)$, due to one the left map being `[1..n]` and the right being `[n+1...n+x]`. This is measuring the cost of one set of new bitwise ops vs old bitwise ops, which is unsurprisingly a little slower.  
Edit (58cbc6a): `union-disj` is better now.
Edit (4b708ba): `intersection-disj` is comparable now.

Let me know if anything else stands out as troublesome.

### DIFF
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -21,13 +21,13 @@ nilNeverChildOfBin t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    Bin _ _ l r -> noNilInSet l && noNilInSet r
+    Bin _ l r -> noNilInSet l && noNilInSet r
   where
     noNilInSet t' =
       case t' of
         Nil -> False
         Tip _ _ -> True
-        Bin _ _ l' r' -> noNilInSet l' && noNilInSet r'
+        Bin _ l' r' -> noNilInSet l' && noNilInSet r'
 
 -- Invariant: The Mask is a power of 2. It is the largest bit position at which
 --            two keys of the map differ.
@@ -36,8 +36,8 @@ maskPowerOfTwo t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    Bin _ m l r ->
-      bitcount 0 (fromIntegral m) == 1 && maskPowerOfTwo l && maskPowerOfTwo r
+    Bin p l r ->
+      bitcount 0 (fromIntegral (getMask p)) == 1 && maskPowerOfTwo l && maskPowerOfTwo r
 
 -- Invariant: Prefix is the common high-order bits that all elements share to
 --            the left of the Mask bit.
@@ -46,9 +46,9 @@ commonPrefix t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    b@(Bin p _ l r) -> all (sharedPrefix p) (keys b) && commonPrefix l && commonPrefix r
+    b@(Bin p l r) -> all (sharedPrefix (getPrefix p)) (keys b) && commonPrefix l && commonPrefix r
   where
-    sharedPrefix :: Prefix -> Int -> Bool
+    sharedPrefix :: Int -> Int -> Bool
     sharedPrefix p a = p == p .&. a
 
 -- Invariant: In Bin prefix mask left right, left consists of the elements that
@@ -58,8 +58,8 @@ maskRespected t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    Bin _ binMask l r ->
-      all (\x -> zero x binMask) (keys l) &&
-      all (\x -> not (zero x binMask)) (keys r) &&
+    Bin p l r ->
+      all (\x -> x .&. getMask p == 0) (keys l) &&
+      all (\x -> x .&. getMask p /= 0) (keys r) &&
       maskRespected l &&
       maskRespected r
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -1,11 +1,10 @@
 module IntMapValidity (valid) where
 
-import Data.Bits (xor, (.&.), (.|.), shiftR)
-import Data.List (intercalate)
+import Data.Bits ((.&.), finiteBitSize, testBit)
+import Data.List (intercalate, elemIndex)
 import Data.IntMap.Internal
 import Numeric (showHex)
-import Test.Tasty.QuickCheck (Property, counterexample, property, (.&&.), conjoin)
-import Utils.Containers.Internal.BitUtil (bitcount, shiftRL)
+import Test.Tasty.QuickCheck (Property, counterexample, property, (.&&.))
 
 {--------------------------------------------------------------------
   Assertions
@@ -32,10 +31,8 @@ nilNeverChildOfBin t =
 
 -- Invariants:
 -- * All keys in a Bin start with the Bin's Prefix.
--- * All keys in the Bin's left child start with the Bin's Prefix followed by a
---   zero bit.
--- * All keys in the Bin's right child start with the Bin's Prefix followed by a
---   one bit.
+-- * All keys in the Bin's left child have the Prefix's mask bit unset.
+-- * All keys in the Bin's right child have the Prefix's mask bit set.
 prefixOk :: IntMap a -> Property
 prefixOk t =
   case t of
@@ -44,11 +41,6 @@ prefixOk t =
     Bin p l r ->
       let px = unPrefix p
           m = px .&. (-px)
-          m' = fromIntegral (fromIntegral m `shiftRL` 1)
-          -- pl = px but make the next bit 0 and shift the mask bit right
-          pl = Prefix ((px `xor` m) .|. m')
-          -- pr = px but keep the next bit 1 and shift the mask bit right
-          pr = Prefix (px .|. m')
           keysl = keys l
           keysr = keys r
           debugStr = concat
@@ -57,16 +49,23 @@ prefixOk t =
             , ", keysr=[" ++ intercalate "," (fmap showIntHex keysr) ++ "]"
             ]
       in counterexample debugStr $
-           all (match p) (keysl ++ keysr) &&
-           all (match pl) keysl &&
-           all (match pr) keysr
+           counterexample "mask bit absent" (px /= 0) .&&.
+           counterexample "prefix not shared" (all (`hasPrefix` p) (keysl ++ keysr)) .&&.
+           counterexample "left child, mask found set" (all (\x -> x .&. m == 0) keysl) .&&.
+           counterexample "right child, mask found unset" (all (\x -> x .&. m /= 0) keysr)
 
--- | Whether the @Int@ starts with the given @Prefix@.
-match :: Prefix -> Int -> Bool
-match p k = (k `xor` px) .&. prefixMask == 0
+-- | Inefficient but easily understandable comparison of prefixes
+hasPrefix :: Int -> Prefix -> Bool
+hasPrefix k p = case elemIndex True pbits of
+  Nothing -> error "no mask bit" -- should already be checked
+  Just i -> drop (i+1) kbits == drop (i+1) pbits
   where
-    px = unPrefix p
-    prefixMask = px `xor` (-px)
+    kbits = toBits k
+    pbits = toBits (unPrefix p)
+
+-- | Bits from lowest to highest.
+toBits :: Int -> [Bool]
+toBits x = fmap (testBit x) [0 .. finiteBitSize (0 :: Int) - 1]
 
 showIntHex :: Int -> String
 showIntHex x = "0x" ++ showHex (fromIntegral x :: Word) ""
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -1,6 +1,10 @@
-module IntMapValidity (valid) where
+module IntMapValidity
+  ( valid
+  , hasPrefix
+  , hasPrefixSimple
+  ) where
 
-import Data.Bits (xor, (.&.))
+import Data.Bits (finiteBitSize, testBit, xor, (.&.))
 import Data.List (intercalate, elemIndex)
 import Data.IntMap.Internal
 import Numeric (showHex)
@@ -55,10 +59,19 @@ prefixOk t =
            counterexample "right child, mask found unset" (all (\x -> x .&. m /= 0) keysr)
 
 hasPrefix :: Int -> Prefix -> Bool
-hasPrefix i p = (i `xor` px) .&. prefixMask == 0
+hasPrefix i p = not (nomatch i p)
+
+-- We test that hasPrefix behaves the same as hasPrefixSimple.
+hasPrefixSimple :: Int -> Prefix -> Bool
+hasPrefixSimple k p = case elemIndex True pbits of
+  Nothing -> error "no mask bit" -- should not happen
+  Just i -> drop (i+1) kbits == drop (i+1) pbits
   where
-    px = unPrefix p
-    prefixMask = px `xor` (-px)
+    kbits = toBits k
+    pbits = toBits (unPrefix p)
+
+    -- Bits from lowest to highest.
+    toBits x = fmap (testBit x) [0 .. finiteBitSize (0 :: Int) - 1]
 
 showIntHex :: Int -> String
 showIntHex x = "0x" ++ showHex (fromIntegral x :: Word) ""
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -34,7 +34,7 @@ nilNeverChildOfBin t =
         Bin _ l' r' -> noNilInSet l' && noNilInSet r'
 
 -- Invariants:
--- * All keys in a Bin start with the Bin's Prefix.
+-- * All keys in a Bin start with the Bin's shared prefix.
 -- * All keys in the Bin's left child have the Prefix's mask bit unset.
 -- * All keys in the Bin's right child have the Prefix's mask bit set.
 prefixOk :: IntMap a -> Property
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -1,9 +1,11 @@
 module IntMapValidity (valid) where
 
-import Data.Bits (xor, (.&.))
+import Data.Bits (xor, (.&.), (.|.), shiftR)
+import Data.List (intercalate)
 import Data.IntMap.Internal
-import Test.Tasty.QuickCheck (Property, counterexample, property, (.&&.))
-import Utils.Containers.Internal.BitUtil (bitcount)
+import Numeric (showHex)
+import Test.Tasty.QuickCheck (Property, counterexample, property, (.&&.), conjoin)
+import Utils.Containers.Internal.BitUtil (bitcount, shiftRL)
 
 {--------------------------------------------------------------------
   Assertions
@@ -12,8 +14,7 @@ import Utils.Containers.Internal.BitUtil (bitcount)
 valid :: IntMap a -> Property
 valid t =
   counterexample "nilNeverChildOfBin" (nilNeverChildOfBin t) .&&.
-  counterexample "commonPrefix" (commonPrefix t) .&&.
-  counterexample "maskRespected" (maskRespected t)
+  counterexample "prefixOk" (prefixOk t)
 
 -- Invariant: Nil is never found as a child of Bin.
 nilNeverChildOfBin :: IntMap a  -> Bool
@@ -29,44 +30,43 @@ nilNeverChildOfBin t =
         Tip _ _ -> True
         Bin _ l' r' -> noNilInSet l' && noNilInSet r'
 
--- Invariant: The Mask is a power of 2. It is the largest bit position at which
---            two keys of the map differ.
-maskPowerOfTwo :: IntMap a -> Bool
-maskPowerOfTwo t =
+-- Invariants:
+-- * All keys in a Bin start with the Bin's Prefix.
+-- * All keys in the Bin's left child start with the Bin's Prefix followed by a
+--   zero bit.
+-- * All keys in the Bin's right child start with the Bin's Prefix followed by a
+--   one bit.
+prefixOk :: IntMap a -> Property
+prefixOk t =
   case t of
-    Nil -> True
-    Tip _ _ -> True
+    Nil -> property ()
+    Tip _ _ -> property ()
     Bin p l r ->
       let px = unPrefix p
           m = px .&. (-px)
-      in bitcount 0 (fromIntegral m) == 1 && maskPowerOfTwo l && maskPowerOfTwo r
+          m' = fromIntegral (fromIntegral m `shiftRL` 1)
+          -- pl = px but make the next bit 0 and shift the mask bit right
+          pl = Prefix ((px `xor` m) .|. m')
+          -- pr = px but keep the next bit 1 and shift the mask bit right
+          pr = Prefix (px .|. m')
+          keysl = keys l
+          keysr = keys r
+          debugStr = concat
+            [ "px=" ++ showIntHex px
+            , ", keysl=[" ++ intercalate "," (fmap showIntHex keysl) ++ "]"
+            , ", keysr=[" ++ intercalate "," (fmap showIntHex keysr) ++ "]"
+            ]
+      in counterexample debugStr $
+           all (match p) (keysl ++ keysr) &&
+           all (match pl) keysl &&
+           all (match pr) keysr
 
--- Invariant: Prefix is the common high-order bits that all elements share to
---            the left of the Mask bit.
-commonPrefix :: IntMap a -> Bool
-commonPrefix t =
-  case t of
-    Nil -> True
-    Tip _ _ -> True
-    b@(Bin p l r) ->
-      let px = unPrefix p
-          prefix = px .&. (px-1)
-      in all (sharedPrefix prefix) (keys b) && commonPrefix l && commonPrefix r
+-- | Whether the @Int@ starts with the given @Prefix@.
+match :: Prefix -> Int -> Bool
+match p k = (k `xor` px) .&. prefixMask == 0
   where
-    sharedPrefix :: Int -> Int -> Bool
-    sharedPrefix p a = p == p .&. a
+    px = unPrefix p
+    prefixMask = px `xor` (-px)
 
--- Invariant: In Bin prefix mask left right, left consists of the elements that
---            don't have the mask bit set; right is all the elements that do.
-maskRespected :: IntMap a -> Bool
-maskRespected t =
-  case t of
-    Nil -> True
-    Tip _ _ -> True
-    Bin p l r ->
-      let px = unPrefix p
-          m = px .&. (-px)
-      in all (\x -> x .&. m == 0) (keys l) &&
-         all (\x -> x .&. m /= 0) (keys r) &&
-         maskRespected l &&
-         maskRespected r
+showIntHex :: Int -> String
+showIntHex x = "0x" ++ showHex (fromIntegral x :: Word) ""
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -37,7 +37,9 @@ maskPowerOfTwo t =
     Nil -> True
     Tip _ _ -> True
     Bin p l r ->
-      bitcount 0 (fromIntegral (getMask p)) == 1 && maskPowerOfTwo l && maskPowerOfTwo r
+      let px = unPrefix p
+          m = px .&. (-px)
+      in bitcount 0 (fromIntegral m) == 1 && maskPowerOfTwo l && maskPowerOfTwo r
 
 -- Invariant: Prefix is the common high-order bits that all elements share to
 --            the left of the Mask bit.
@@ -46,7 +48,10 @@ commonPrefix t =
   case t of
     Nil -> True
     Tip _ _ -> True
-    b@(Bin p l r) -> all (sharedPrefix (getPrefix p)) (keys b) && commonPrefix l && commonPrefix r
+    b@(Bin p l r) ->
+      let px = unPrefix p
+          prefix = px .&. (px-1)
+      in all (sharedPrefix prefix) (keys b) && commonPrefix l && commonPrefix r
   where
     sharedPrefix :: Int -> Int -> Bool
     sharedPrefix p a = p == p .&. a
@@ -59,7 +64,9 @@ maskRespected t =
     Nil -> True
     Tip _ _ -> True
     Bin p l r ->
-      all (\x -> x .&. getMask p == 0) (keys l) &&
-      all (\x -> x .&. getMask p /= 0) (keys r) &&
-      maskRespected l &&
-      maskRespected r
+      let px = unPrefix p
+          m = px .&. (-px)
+      in all (\x -> x .&. m == 0) (keys l) &&
+         all (\x -> x .&. m /= 0) (keys r) &&
+         maskRespected l &&
+         maskRespected r
--- a/containers-tests/tests/IntMapValidity.hs
+++ b/containers-tests/tests/IntMapValidity.hs
@@ -1,6 +1,6 @@
 module IntMapValidity (valid) where
 
-import Data.Bits ((.&.), finiteBitSize, testBit)
+import Data.Bits (xor, (.&.))
 import Data.List (intercalate, elemIndex)
 import Data.IntMap.Internal
 import Numeric (showHex)
@@ -54,18 +54,11 @@ prefixOk t =
            counterexample "left child, mask found set" (all (\x -> x .&. m == 0) keysl) .&&.
            counterexample "right child, mask found unset" (all (\x -> x .&. m /= 0) keysr)
 
--- | Inefficient but easily understandable comparison of prefixes
 hasPrefix :: Int -> Prefix -> Bool
-hasPrefix k p = case elemIndex True pbits of
-  Nothing -> error "no mask bit" -- should already be checked
-  Just i -> drop (i+1) kbits == drop (i+1) pbits
+hasPrefix i p = (i `xor` px) .&. prefixMask == 0
   where
-    kbits = toBits k
-    pbits = toBits (unPrefix p)
-
--- | Bits from lowest to highest.
-toBits :: Int -> [Bool]
-toBits x = fmap (testBit x) [0 .. finiteBitSize (0 :: Int) - 1]
+    px = unPrefix p
+    prefixMask = px `xor` (-px)
 
 showIntHex :: Int -> String
 showIntHex x = "0x" ++ showHex (fromIntegral x :: Word) ""
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -368,6 +368,9 @@ data IntMap a = Bin {-# UNPACK #-} !Prefix
 --   All keys in the left child of a Bin have the mask bit unset, and all keys
 --   in the right child have the mask bit set.
 
+-- See Note [Okasaki-Gill] for how the implementation here relates to the one in
+-- Okasaki and Gill's paper.
+
 -- | A @Prefix@ is some prefix of high-order bits of an @Int@.
 --
 -- This is represented by an @Int@ which starts with the prefix bits,
@@ -3725,3 +3728,47 @@ node = "+--"
 withBar, withEmpty :: [String] -> [String]
 withBar bars   = "|  ":bars
 withEmpty bars = "   ":bars
+
+{--------------------------------------------------------------------
+  Notes
+--------------------------------------------------------------------}
+
+-- Note [Okasaki-Gill]
+-- ~~~~~~~~~~~~~~~~~~~
+--
+-- The IntMap structure is based on the map described in the paper "Fast
+-- Mergeable Integer Maps" by Chris Okasaki and Andy Gill, with some
+-- differences.
+--
+-- The paper spends most of its time describing a little-endian tree, where the
+-- branching is done first on low bits then high bits. It then briefly describes
+-- a big-endian tree. The implementation here is big-endian.
+--
+-- The definition of Okasaki and Gill's map would be written in Haskell as
+--
+-- data Dict a
+--   = Empty
+--   | Lf !Int a
+--   | Br !Int !Int !(Dict a) !(Dict a)
+--
+-- Empty is the same as IntMap's Nil, and Lf is the same as Tip.
+--
+-- In Br, the first Int is the shared prefix and the second is the mask bit by
+-- itself. For the big-endian map, the paper suggests that the prefix be the
+-- common prefix, followed by a 0-bit, followed by all 1-bits. This is so that
+-- the prefix value can be used as a point of split for binary search.
+--
+-- IntMap's Bin corresponds to Br, but is different because it has only one
+-- Int (newtyped as Prefix). This describes both prefix and mask, so it is not
+-- necessary to store them separately. This value is, in fact, one plus the
+-- value suggested for the prefix in the paper. This representation is chosen
+-- because it saves one word per Bin without detriment to the efficiency of
+-- operations.
+--
+-- The implementation of operations such as lookup, insert, union, follow
+-- the described implementations on Dict and split into the same cases. For
+-- instance, for insert, the three cases on a Br are whether the key belongs
+-- outside the map, or it belongs in the left child, or it belongs in the
+-- right child. We have the same three cases for a Bin. However, the bitwise
+-- operations we use to determine the case is naturally different due to the
+-- difference in representation.
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -3558,10 +3558,12 @@ nomatch i p = i .&. prefixMask /= px .&. prefixMask
     prefixMask = px `xor` (-px)
 {-# INLINE nomatch #-}
 
--- | Whether the @Int@ is less than the minimum key that may be present in
--- the right child.
+-- | Whether the @Int@ is to the left of the split created by a @Bin@ with this
+-- @Prefix@.
+-- This does not imply that the @Int@ belongs in this @Bin@. That fact is
+-- usually determined first using @nomatch@.
 left :: Int -> Prefix -> Bool
-left i (Prefix p) = natFromInt i < natFromInt p
+left i p = natFromInt i < natFromInt (unPrefix p)
 {-# INLINE left #-}
 
 -- | Whether this Prefix splits a Bin at the sign bit.
@@ -3569,7 +3571,7 @@ left i (Prefix p) = natFromInt i < natFromInt p
 -- If it is true, the left child contains non-negative keys and the right child
 -- contains negative keys.
 signBranch :: Prefix -> Bool
-signBranch (Prefix p) = p == (minBound :: Int)
+signBranch p = unPrefix p == (minBound :: Int)
 {-# INLINE signBranch #-}
 
 -- | The prefix of key @i@ up to (but not including) the switching
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -360,23 +360,39 @@ data IntMap a = Bin {-# UNPACK #-} !Prefix
                     !(IntMap a)
               | Tip {-# UNPACK #-} !Key a
               | Nil
--- IntMap invariants:
+
+-- IntMap structure and invariants:
+--
 -- * Nil is never found as a child of Bin.
--- * The Prefix of a Bin is the common high-order bits that all keys in the
---   Bin share.
--- * We call the bit immediately following the shared prefix the mask bit.
---   All keys in the left child of a Bin have the mask bit unset, and all keys
---   in the right child have the mask bit set.
+--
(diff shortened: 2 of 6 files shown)


treeowl:
This is most impressive. I will do my best to review it promptly. In the mean time, I have a big favor to ask: I was struggling the other day to merge the branch for version 0.7 into master. Do you think you could try to figure out how to do that right?

BurningWitness:
(prefix + mask) is just a part of the key with an additional mask bit set, could this be abused for performance?

With a key and a compatible prefix, `xor` only yields a value at or below the level of the masking bit, whether or not the value is at the masking bit determines the direction (`0` is right).
```
    1800 | 0000011100001000
xor 1807 | 0000011100001111
---------+-----------------
         | 0000000000000111
       
    1800 | 0000011100001000
xor 1795 | 0000011100001000
---------+-----------------
         | 0000000000001011
```

---

So the two comparison patterns could be reduced to the following:

```haskell
-- Single-key operation
trimatch p k =
  let o = xor p k
  in case () of
       () | o .&. p == 0                = goRight
          | unsafeShiftR o 1 .&. p == 0 = goLeft
          | otherwise                   = goFail

-- Merge
quadbranch p1 p2 =
  let o = xor p k `unsafeShiftR` 1
  in case () of
       () | p1 == p2      = goEqual
          | o .&. p1 == 0 = goBelowP1
          | o .&. p2 == 0 = goBelowP2
          | otherwise     = goDiffer
```

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3499,82 @@ instance Read1 IntMap where
...
-linkWithMask m p1 t1 {-p2-} t2
-  | zero p1 m = Bin p m t1 t2
-  | otherwise = Bin p m t2 t1
+linkWithMask :: Int -> Key -> IntMap a -> Key -> IntMap a -> IntMap a
+linkWithMask m k1 t1 k2 t2
+  | natFromInt k1 < natFromInt k2 = Bin p t1 t2
+  | otherwise = Bin p t2 t1
   where
-    p = mask p1 m
+    p = Prefix (mask k1 m + m)
```
BurningWitness, line 3517: I think this addition could be a bitwise or, but that would need to be done before the sign conversion.

meooow25:
@treeowl I assume you mean #994. I don't know much about the CI setup but I'll take a look.

> With a key and a compatible prefix, xor only yields a value at or below the level of the masking bit, whether or not the value is at the masking bit determines the direction (0 is right).

@BurningWitness it is unclear to me how that would work. For instance, `xor p k .&. p == 0`, which is the same as `k .&. p == p`, only tells us whether `p`'s 1-bits agree with `k`. But we also need to know if it the 0-bits agree.  
Probably worth thinking about it though, it would be great to find a more efficient set of bitwise ops to do the job.

containers/src/Data/IntMap/Internal.hs
meooow25: > I think this addition could be a bitwise or

Yes. Do you suggest we use or?

> that would need to be done before the sign conversion.

I don't understand this part. Which sign conversion do you mean?

BurningWitness:
Oh, true, `xor` only has that guarantee if prefix matches the key, so I was wrong.

containers/src/Data/IntMap/Internal.hs
BurningWitness: Sign conversion as in `intFromNat` on `maskW`. Though actually yes, this is setting the lowest bit so the sign doesn't even matter.

> Do you suggest we use or?

In my mind it's a "cleaner" operation because there is no extra carry flag, so it's a "yes, but don't expect it to ever make any difference".

meooow25: That's understandable, updated to use bitwise or.
Note: There are more places in this file where bitwise or could be used (not introduced in this PR), but I'm not touching those for now.

meooow25:
> Oh, true, `xor` only has that guarantee if prefix matches the key, so I was wrong.

Ah well if the prefix already matches, the branch into left or right is done with one comparison so it's hard to imagine something faster than that.  
What I'd like to find a better way for is the `nomatch` stuff.

meooow25:
I believe `nomatch` should be pretty much as fast as the old one now!

Previous: ``i .&. (m `xor` (-m)) == p`` — 1 and, 1 xor, 1 negate, 1 compare
New: ``(i `xor` p) .&. (p `xor` (-p))`` — 1 and, 2 xor, 1 negate

Correct me if I'm wrong, I know little about things at this low a level.

---

Now, if only `shorter` could be faster or avoided in some manner...
I suspect that's the only thing causing minor slowdowns for the `intersection_disj` case.

treeowl:
I doubt the comparison costs significantly over the test, but benchmarking is the way to know for sure.

meooow25:
Okay I found a better way to branch that avoids `shorter`. Almost every benchmark is faster or the same now! 🎉 

@treeowl would you like to review? I think it's in a good shape now.

BurningWitness:
It would be nice to wrap `p .|. (p - 1)` and `p .&. (p - 1)` into functions like `upper` and `lower` respectively for clarity.

---

Also this should mean precise lookups can be expressed as

```haskell
trimatch p k =
  if k < p
    then if k >= lower p
           then goLeft
           else goFail
           
    else if k <= upper p
           then goRight
           else goFail
```

(I think this is faster than the algorithm listed in the paper, four operations in any direction instead of four for failure plus one for picking a side)

containers-tests/tests/IntMapValidity.hs
```diff
@@ -46,9 +48,12 @@ commonPrefix t =
...
+    b@(Bin p l r) ->
+      let px = unPrefix p
+          prefix = px .&. (px-1)
+      in all (sharedPrefix prefix) (keys b) && commonPrefix l && commonPrefix r
   where
-    sharedPrefix :: Prefix -> Int -> Bool
+    sharedPrefix :: Int -> Int -> Bool
     sharedPrefix p a = p == p .&. a
 
 -- Invariant: In Bin prefix mask left right, left consists of the elements that
```
treeowl, line 59: This comment (and any others mentioning the old structure) should be changed. I won't require perfection, but try to give it a once-over.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -353,26 +356,22 @@ intFromNat = fromIntegral
 
 -- See Note: Order of constructors
 data IntMap a = Bin {-# UNPACK #-} !Prefix
-                    {-# UNPACK #-} !Mask
```
treeowl, line 356: Somewhere (probably around here, but it could be in a separate file if you'd prefer) we need an explicit discussion of the relationship between the new representation and the way it's done in the paper.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3469,116 @@ instance Read1 IntMap where
...
   = (mask i m) /= p
-{-# INLINE nomatch #-}
+{-# INLINE nomatchMask #-}
 
--- | Does the key @i@ match the prefix @p@ (up to but not including
--- bit @m@)?
-match i p m
-  = (mask i m) == p
-{-# INLINE match #-}
+-- | Whether the @Int@ does not start with the given @Prefix@.
```
treeowl, line 3555: Could you state this more clearly, and perhaps explain briefly how it's used?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3469,116 @@ instance Read1 IntMap where
...
+-- This does not imply that the @Int@ belongs in this @Bin@. That fact is
+-- usually determined first using @nomatch@.
+left :: Int -> Prefix -> Bool
+left i p = natFromInt i < natFromInt (unPrefix p)
+{-# INLINE left #-}
+
+-- | Whether this Prefix splits a Bin at the sign bit.
+-- This can only be True at the top level.
+-- If it is true, the left child contains non-negative keys and the right child
+-- contains negative keys.
```
treeowl, line 3574: I like the documentation of these two.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3549,19 +3590,20 @@ mask i m
...
   = intFromNat (i .&. ((-m) `xor` m))
 {-# INLINE maskW #-}
 
--- | Does the left switching bit specify a shorter prefix?
-shorter :: Mask -> Mask -> Bool
-shorter m1 m2
-  = (natFromInt m1) > (natFromInt m2)
-{-# INLINE shorter #-}
+-- | Whether the left prefix is shorter than the prefix length indicated by the
+-- right mask.
```
treeowl, line 3599: Would it be easy enough to expand this to suggest why we have it?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -353,26 +356,22 @@ intFromNat = fromIntegral
...
-
+-- | A @Prefix@ stores the most significant bits shared by all keys in a Bin,
+-- immediately followed a single set bit, which we refer to as "mask".
+--
+-- Invariant: All keys in the left child have the mask bit unset, and all keys
+--            in the right child have the mask bit set.
+--
+-- It follows from the invariant that the @Int@ value of the @Prefix@ is the
+-- smallest value that can be present in the right child.
+newtype Prefix = Prefix { unPrefix :: Int }
```
treeowl, line 373: 🎉🎆 I love newtypes like this, as long as they don't make much noise. Awesome.

treeowl (changes requested):
I will need to do a more thorough review of this on Sunday, but I've left a few comments for now. I'm exited about your work here!

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3547,6 +3521,32 @@ binCheckRight _ l Nil = l
...
+
+mapMapBranch :: Prefix -> Prefix -> MapMapBranch
+mapMapBranch p1 p2 = case compare pw1 pw2 of
+  LT | pw2 <= (pw1 .|. (pw1-1)) -> ABR
+     | (pw2 .&. (pw2-1)) <= pw1 -> BAL
+     | otherwise                -> NOM
+  GT | pw1 <= (pw2 .|. (pw2-1)) -> BAR
+     | (pw1 .&. (pw1-1)) <= pw2 -> ABL
+     | otherwise                -> NOM
+  EQ                            -> EQL
```
BurningWitness, line 3538: Moving `ABL` above `BAR` would mean `(pw1 - 1)` is shared between both branches of `pw1 < pw2`, so it can be moved up in the flow.

containers-tests/tests/IntMapValidity.hs
meooow25: These edits were actually a bit silly. It makes sense to redo this validation code, updated now.
I also can't find any more mentions like `Bin prefix mask` (except in the context of `IntSet`) but let me know if I missed any.

meooow25:
@BurningWitness 

> It would be nice to wrap `p .|. (p - 1)` and `p .&. (p - 1)` into functions named `upper` and `lower` respectively for clarity.

If that becomes the predominant way of viewing `Prefix`, then perhaps. Currently we also use `nomatch` so that is not true.

> Also this should mean precise lookups can be expressed as...

I like that, since it uses the same view as `mapMapBranch`, but it makes no difference on benchmarks for me. A small concern (maybe?) is that this duplicates the failure branch.  
So we could switch to it, but I don't really see a good reason to do so. If it is demonstrated to be better in some way, then sure I would favor it.

containers/src/Data/IntMap/Internal.hs
BurningWitness: True, didn't think of that.

BurningWitness:
> A small concern (maybe?) is that this duplicates the failure branch.

Indeed that is the case, the failure branch needs to be let-bound with a `NOINLINE` to avoid it. Probably won't be visible on the benchmarks anyway though.

BurningWitness:
I made a local benchmark on a `Word`-based tree clone and indeed the `nomatch` solution is ahead performance-wise by a remarkably narrow margin, 1-2% difference.

containers/src/Data/IntMap/Internal.hs
meooow25: Done.

meooow25:
Hi @treeowl, please review when you get the time.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3476,127 @@ instance Read1 IntMap where
...
+  where
+    pw1 = natFromInt (unPrefix p1)
+    pw2 = natFromInt (unPrefix p2)
+{-# INLINE mapMapBranch #-}
```
treeowl, line 3551: I'm not great at bit fiddling, so I find this rather intimidating. Do you think you could explain how it works? I don't have a good sense of what sorts of bit pattern relationships it's looking for in each test.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3476,127 @@ instance Read1 IntMap where
...
+-- | Link two @IntMap@s. The maps must not be empty. The @Prefix@es of the two
+-- maps must be different. @k1@ must share the prefix of @t1@. @p2@ must be the
+-- prefix of @t2@.
+linkKey :: Key -> IntMap a -> Prefix -> IntMap a -> IntMap a
+linkKey k1 t1 p2 t2 = link k1 t1 (unPrefix p2) t2
+{-# INLINE linkKey #-}
+
+-- | Link two @IntMap@s. The maps must not be empty. The @Prefix@es of the two
+-- maps must be different. @k1@ must share the prefix of @t1@ and @k2@ must
+-- share the prefix of @t2@.
```
treeowl, line 3489: What are `k1` and `k2`?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3549,19 +3608,20 @@ mask i m
...
+shorterMask :: Prefix -> Int -> Ordering
+shorterMask p1 m2 = natFromInt m2 `compare` natFromInt (px1 .&. (-px1))
+  where px1 = unPrefix p1
+{-# INLINE shorterMask #-}
 
 -- | The first switching bit where the two prefixes disagree.
-branchMask :: Prefix -> Prefix -> Mask
+branchMask :: Int -> Int -> Int
 branchMask p1 p2
   = intFromNat (highestBitMask (natFromInt p1 `xor` natFromInt p2))
```
treeowl, line 3626:
```suggestion
  = intFromNat (highestBitMask (natFromInt (p1 `xor` p2)))
```

treeowl: Actually, I think you can do better; change `highestBitMask` to work with `Int` instead of `Word`. `IntSet` will have to be adjusted, but you're planning to change that anyway.

containers-tests/tests/IntMapValidity.hs
```diff
@@ -12,54 +13,59 @@ import Utils.Containers.Internal.BitUtil (bitcount)
...
-      bitcount 0 (fromIntegral m) == 1 && maskPowerOfTwo l && maskPowerOfTwo r
+    Nil -> property ()
+    Tip _ _ -> property ()
+    Bin p l r ->
+      let px = unPrefix p
+          m = px .&. (-px)
+          keysl = keys l
+          keysr = keys r
+          debugStr = concat
+            [ "px=" ++ showIntHex px
```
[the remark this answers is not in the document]
meooow25: Yes, but binary is just too verbose. For example, we might get a failure like
```
    *** Failed! Falsified (after 8 tests and 9 shrinks):
    -1
    fromList [(0,()),(1,())]
    prefixOk
    px=0b1, keysl=[0b0], keysr=[0b1111111111111111111111111111111111111111111111111111111111111111,0b1]
    prefix not shared
```
With hex, that is
```
    px=0x1, keysl=[0x0], keysr=[0xffffffffffffffff,0x1]
```

containers-tests/tests/IntMapValidity.hs
```diff
@@ -12,54 +13,59 @@ import Utils.Containers.Internal.BitUtil (bitcount)
...
 
--- Invariant: Prefix is the common high-order bits that all elements share to
---            the left of the Mask bit.
-commonPrefix :: IntMap a -> Bool
-commonPrefix t =
-  case t of
-    Nil -> True
-    Tip _ _ -> True
-    b@(Bin p _ l r) -> all (sharedPrefix p) (keys b) && commonPrefix l && commonPrefix r
+-- | Inefficient but easily understandable comparison of prefixes
```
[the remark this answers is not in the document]
meooow25: Switched to an efficient version, but adding tests for tests is a bit too far, don't you think?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -353,26 +356,29 @@ intFromNat = fromIntegral
...
+
+-- See Note [Okasaki-Gill] for how the implementation here relates to the one in
+-- Okasaki and Gill's paper.
+
+-- | A @Prefix@ is some prefix of high-order bits of an @Int@.
+--
+-- This is represented by an @Int@ which starts with the prefix bits,
+-- immediately followed by a set bit. This is the mask bit for a @Bin@. It
+-- follows from the IntMap invariants that this @Int@ value is the smallest
+-- value that can be present in its right child.
```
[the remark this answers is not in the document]
meooow25: This is pretty clear, I like it. Updated with some minor changes.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3476,127 @@ instance Read1 IntMap where
...
-zero i m
-  = (natFromInt i) .&. (natFromInt m) == 0
-{-# INLINE zero #-}
+data MapMapBranch
+  = ABL  -- ^ A contains B in the left child
+  | ABR  -- ^ A contains B in the right child
+  | BAL  -- ^ B contains A in the left child
+  | BAR  -- ^ B contains A in the right child
+  | EQL  -- ^ A and B have equal prefixes
+  | NOM  -- ^ A and B have prefixes that do not match
```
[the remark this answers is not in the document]
meooow25: Added docs.

containers/src/Data/IntMap/Internal.hs
meooow25: How do you mean "better"?
In terms of efficiency, it should not make a difference. Int <-> Word is free.
In terms of clarity, I agree that changing everything to work on Int would be good. I don't know why the Word/Nat stuff was done, maybe it made sense at the time.

containers/src/Data/IntMap/Internal.hs
meooow25: There are cases to consider here, but the bit fiddling is not too complex.

* `pw .|. (pw-1)` sets every bit below the mask bit to 1. This is the largest key the Bin can have.
* `pw .&. (pw-1)` sets the mask bit and every bit below it to 0. This is the smallest key the Bin can have.

First, we compare the prefixes to each other. Then we compare a prefix against the smallest/largest keys of the other prefix. This gives us enough information to tell how they relate to each other.

* If `pw1` from Bin A is less than `pw2` from Bin B, and `pw2` is <= the greatest key in the Bin A, the right child of Bin A covers the entirety of Bin B.
* ...and so on

containers-tests/tests/IntMapValidity.hs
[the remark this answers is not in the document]
flip111: fits in the terminal width with binary no? I'm not sure which one would be better but i sure like that byte chunks would be separated.
```
0b11111111 11111111 11111111 11111111 11111111 11111111 11111111 11111111
or
0xff ff ff ff ff ff ff ff
```

containers/src/Data/IntMap/Internal.hs
flip111: afaik `Word` is clearer when dealing with bits. `Int` is better for signed integers

treeowl: Converting between `Int` and `Word` is free at runtime. But those conversions have to be carried around in Core, and as far as I know they count towards unfolding "size" calculations used to guide inlining. So I don't think they're actually entirely free. Of course, that's probably also true to some extent for `newtype` casts, so maybe it's not worth worrying about?

Was some of the `Word` stuff done to catch type errors, or is it not disciplined enough for that?

containers/src/Data/IntMap/Internal.hs
treeowl: I keep wondering if there's some way to get away with fewer conditional branches here, and whether doing so would help performance or not. We have only six possible results, but can perform up to four tests. Is there anything more "clever" that would be worth trying? Last I checked, GHC's native code generator really didn't like doing branchless stuff, but maybe there's something else to play with? If so, that obviously doesn't have to be done before the merge.

containers/src/Data/IntMap/Internal.hs
treeowl: Anyway, simpler code is better, so streamlining things just to make it easier to read would be a plus, but not urgent.

containers-tests/tests/IntMapValidity.hs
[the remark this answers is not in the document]
treeowl: Nope, I don't think that's too far at all! Maybe I'm weird?

containers-tests/tests/IntMapValidity.hs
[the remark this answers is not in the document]
treeowl: Ugh, yeah, you're right. In an ideal world, it would be great to have a debug build that uses `Int8` and shows things in binary, but that sounds like more trouble than it's worth for the moment.

containers/src/Data/IntMap/Internal.hs
treeowl: Would you mind putting this explanation in code comments? I think it's brief enough, and a great way to guide future contributions into things.

containers-tests/tests/IntMapValidity.hs
```diff
@@ -12,54 +17,61 @@ import Utils.Containers.Internal.BitUtil (bitcount)
...
         Tip _ _ -> True
-        Bin _ _ l' r' -> noNilInSet l' && noNilInSet r'
+        Bin _ l' r' -> noNilInSet l' && noNilInSet r'
 
--- Invariant: The Mask is a power of 2. It is the largest bit position at which
---            two keys of the map differ.
-maskPowerOfTwo :: IntMap a -> Bool
-maskPowerOfTwo t =
+-- Invariants:
+-- * All keys in a Bin start with the Bin's Prefix.
```
treeowl, line 37:
```suggestion
-- * All keys in a Bin start with the Bin's shared prefix.
```

containers/src/Data/IntMap/Internal.hs
meooow25: > afaik `Word` is clearer when dealing with bits. `Int` is better for signed integers

I disagree, there is no difference between `Word` and `Int` when treating them as a pattern of bits.

> ...as far as I know they count towards unfolding "size" calculations used to guide inlining

Interesting if that's true... all of these are marked INLINE though so it should be fine.

> Was some of the Word stuff done to catch type errors, or is it not disciplined enough for that?

I can't tell for sure, and there is no documented reason I can find. The Word stuff has been there since the beginning: https://github.com/haskell/containers/commit/bbbba97cbcf12039810533e3a2daf2eefdefe7f0

> Anyway, simpler code is better, so streamlining things just to make it easier to read would be a plus, but not urgent.

Agreed. If you're fine with it I can try removing the Word stuff after these changes are done to IntMap/IntSet.

containers-tests/tests/IntMapValidity.hs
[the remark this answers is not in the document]
meooow25: Added a property test.

containers/src/Data/IntMap/Internal.hs
BurningWitness: If the concern is that `mapMapBranch` won't inline, then I think the best way forward is to make synonyms for comparisons and unroll the rest of the function at every definition. It's not that much larger than what you have now and the cases are overwhelmingly independent.

```haskell
case compare pA pB of
  EQ                  -> goEQ

  LT | pB <= upper pA -> goABR
     | pA >= lower pB -> goBAL
     | otherwise      -> goANB

  GT | pA <= upper pB -> goBAR
     | pB >= lower pA -> goABL
     | otherwise      -> goBNA
```

There is probably a better way to name the comparisons, though if you know how the tree works they are self-explanatory (`lower`/`upper` are `left`/`right` boundaries respectively and you are querying whether a node is *inside* another node).

BurningWitness: I don't know if there's anything to optimize in the function itself, it's extremely minimal and it's only downside is that it's completely sequential. I guess you could try to force the evaluation of the "lower" comparison before the "upper" one so they're parallel instruction-wise, but I wouldn't be surprised if that optimization is already performed automatically.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3475,69 +3492,148 @@ instance Read1 IntMap where
...
+-- | Calculates how two @Bin@s relate to each other by comparing their
+-- @Prefix@es.
+
+-- Notes:
+-- * pw .|. (pw-1) sets every bit below the mask bit to 1. This is the greatest
+--   key the Bin can have.
+-- * pw .&. (pw-1) sets the mask bit and every bit below it to 0. This is the
+--   smallest key the Bin can have.
+--
+-- First, we compare the prefixes to each other. Then we compare a prefix
```
treeowl, line 3567: What do we learn by comparing the prefixes to each other? That's not exactly an obvious thing to do, and needs documentation. Here's my understanding so far:

`p1` is the smallest permissible value in the right child of its `Bin`. If `pw1 < pw2`, then (according to `left`) `p1` is to the left of the split created by `p2`'s `Bin`.

How exactly does that help?

containers/src/Data/IntMap/Internal.hs
treeowl: Eh, we'll worry about inlining issues later. If there are issues, they're sure to be subtle.

containers/src/Data/IntMap/Internal.hs
flip111: > I disagree, there is no difference between `Word` and `Int` when treating them as a pattern of bits.

When you designate some bytes as `Int` it means that specific bits serve to indicate whether the number is positive or negative usually by [Two's complement](https://en.wikipedia.org/wiki/Two%27s_complement). If you are just using a pattern of bits then the designation `Word` is correct.

Also you should verify that your logic works with a variable amount of bits because the size of `Int` is not guaranteed to be the same in all cases. If your logic depends on specific size then you should use Word8, Word16, Word32 or Word64.

containers/src/Data/IntMap/Internal.hs
meooow25: I don't know how to explain this in a way that conveys more information than the code.

> If `pw1 < pw2`, then (according to `left`) `p1` is to the left of the split created by `p2`'s `Bin`.

That's correct.

> How exactly does that help?

It helps because it narrows down the possibilities from 6 to 3. For instance, if `p1`'s `Bin` is in the right child of `p2`'s `Bin` it implies that `pw1 > pw2`, since `pw2` is the smallest permissible value of the right child of `p2`'s `Bin`. Which is ruled out if `pw1 < pw2` is true.

containers/src/Data/IntMap/Internal.hs
meooow25: > If the concern is that `mapMapBranch` won't inline...

I've marked it `INLINE` and checked the Core to make sure that GHC does its thing and is able to branch on the comparisons directly, so we should be good on that front.

treeowl:
I'm ready to merge, but I want to make sure the introduction of the helpers in your last commit didn't hurt performance. Did you check the performance effect of that, or (alternatively) verify that the Core, unfolding, and unfolding guidance are the same?

meooow25:
Awesome 🎉 
I peeked at the Core for `union` and it seemed alright. But I will run the benchmarks a final time. I also want to go over the PR once to check for bugs, since it touches so many functions. I will update when I'm done, should be some time later today.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -353,26 +356,45 @@ intFromNat = fromIntegral
...
+--
+-- * All bits to the right of the mask bit are set to 0 in a Prefix.
+--
 
+-- See Note [Okasaki-Gill] for how the implementation here relates to the one in
+-- Okasaki and Gill's paper.
+
+-- | A @Prefix@ represents some prefix of high-order bits of an @Int@.
+newtype Prefix = Prefix { unPrefix :: Int }
+  deriving (Eq, Lift)
```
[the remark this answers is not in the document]
treeowl: Yeah, that's altogether fine.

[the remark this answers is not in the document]
treeowl: Deriving `Lift` instances is almost always the right thing. The only exceptions I can think of are:

1. Data structures with really weird things like `IORef`s, `TVar`s, pointer cycles, or infinite/very large lazily generated structures. These require special treatment to lift, when they can be lifted at all.
2. Redundant data structures where a particular canonical form is more desirable. This is the case for `Data.Sequence.Seq`, where the `Lift` instance puts its argument into perfectly balanced left-heavy form. I don't remember what we do for `Map` and `Set`, but we might also have special logic for those.

Note that you don't really have to worry too much about `Lift` breaking abstraction barriers, because Template Haskell already smashes those to smithereens.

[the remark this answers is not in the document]
treeowl: Hmmm.... Actually..... Now you're reminding me of another issue, which is code size of lifted structures. Sometimes it's possible to reduce that (e.g., by consolidating coercions) using a custom instance. In principle, we could write a `Lift` instance for `IntMap` that uses a "smart constructor" instead of `Bin`, to take an `Int` instead of a `Prefix`. We'd have to see if that actually helped in the end or not.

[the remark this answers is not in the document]
treeowl: By code size, I actually meant Core term size, which can affect compilation time and such. It doesn't directly affect the compiled code size.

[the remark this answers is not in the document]
treeowl: tl;dr: the newtype you've introduced will almost certainly make some things slower at compile time (type checking and optimizing splices generated by `lift`). But the only workarounds I can think of will either slow down runtime code or use `unsafeCoerce` in an extremely shady way that I'm guessing would affect profiling. So don't worry about it.

containers/src/Data/IntMap/Internal.hs
meooow25: > When you designate some bytes as `Int` it means that specific bits serve to indicate whether the number is positive or negative usually by [Two's complement](https://en.wikipedia.org/wiki/Two%27s_complement).

This is true, but if we do not care about the _numeric value_, it doesn't matter. It is just some bits.

> Also you should verify that your logic works with a variable amount of bits because the size of `Word`/`Int` is not guaranteed to be the same in all cases. If your logic depends on specific size then you should use Word8, Word16, Word32 or Word64.

`IntMap` works with whatever the size of `Int` is.

Anyway, this is getting a little off-topic for this PR.

containers/src/Data/IntMap/Internal.hs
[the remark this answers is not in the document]
meooow25: Got it 👍

meooow25:
Ok I think this is good to go.

Updated the benchmarks. I found some odd increases in a few `intersection` tests that seemed to have originated with 82f4852, a completely unrelated change. Adding a `-fproc-alignment=64` flag (as suggested by `tasty-bench`) fixed that.

I also added the `"-with-rtsopts=-A32m"` flag from tasty-bench while at it. This greatly reduced the `Copied` stats however, for both before and after. I think that column is less indicative of anything now, but I've kept it for the sake of it.

Let me know if thing look good and I'll squash it into a few commits (benchmark flag, tests, main IntMap changes).

treeowl:
Why would you set `-A32m` here? That seems to change what the benchmarks are measuring quite a lot. The alignment thing seems reasonable, although it's unfortunate we have to do that.

meooow25:
`tasty-bench` claims it makes the individual benchmarks less likely to affect one another. I thought that would be good to have as a general improvement (and not to address by any observed issue).

treeowl:
Let's wait with that one for now. I'd like to hear opinions from others on a separate PR.

meooow25:
Sounds good. Removed the flag and re-ran the benchmarks.

treeowl:
Aaaaand we're merged! Fantastic work!

meooow25:
Thanks for reviewing! Next up... `IntSet` 😄

PR #1005, opened 2024-05-28
Fix Issue 1004: effect order in filterAMissing

j6carey:
The order of Applicative effects in filterAMissing
was incorrect, causing the order of effects to
differ from key order and be influenced by how
the binary tree was balanced.

The fix is to arrange that effects arising from
the key and value at an internal node come after
those in its left branch instead of before.
(Regardless of this fix such effects come before
those effects arising from the right branch.)

This change also expands test coverage
to detect a regression of this fix.


j6carey:
Since I lack permission to initial a continuous integration workflow or assign a reviewer, would someone be so kind as to trigger those steps?

treeowl:
Sorry I missed this.

treeowl:
Thanks!

j6carey:
Thanks for the CI and merge!

PR #1006, opened 2024-05-30
Update doc comments for sorting

konsumlamm:
Explain what stable means and fix `sortOn`/`unstableSortOn` documentation (it was copied from `Data.List.sortOn` and still said "input list"). Also remove the part about "decorate-sort-undecorate paradigm" and "Schwartzian transform", as that doesn't help IMO (it sounds overly technical and I assume most people have never heard of that, at least I haven't).


meooow25 (approved):

meooow25:
Thanks!

PR #1008, opened 2024-05-30
Fix false negatives in IntMap.isProperSubmapOfBy

### LINKED ISSUE #1007: `IntMap` `isProperSubmapOfBy` test failure
I got the following test failure for `isProperSubmapOfBy` in `intmap-properties`:
```
  isProperSubmapOfBy:                    FAIL
    *** Failed! Falsified (after 4 tests and 11 shrinks):
    {_->True}
    fromList [(-3,1),(-1,1)]
    fromList [(-3,1),(-1,1),(0,1)]
    False /= True
    Use --quickcheck-replay=905941 to reproduce.
    Use -p '$0=="intmap-properties.isProperSubmapOfBy"' to rerun this test only.
```

meooow25:
Fixes #1007

### DIFF
--- a/containers-tests/tests/intset-properties.hs
+++ b/containers-tests/tests/intset-properties.hs
@@ -20,6 +20,7 @@ main = defaultMain $ testGroup "intset-properties"
                    , testCase "lookupLE" test_lookupLE
                    , testCase "lookupGE" test_lookupGE
                    , testCase "split" test_split
+                   , testCase "isProperSubsetOf" test_isProperSubsetOf
                    , testProperty "prop_Valid" prop_Valid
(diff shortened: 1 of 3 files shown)


containers-tests/tests/intset-properties.hs
treeowl, line 1: Could you add a property test for sets and maps that catches this?

meooow25: The property test is how konsumlamm caught this (#1007).

But it's random and such cases are apparently unlikely, so it didn't show up when the test was added (in #995).

treeowl:
Thanks!

PR #1009, opened 2024-06-06
Add symmetric difference ops

### LINKED ISSUE #1002: Symmetric difference for sets
`Set` and `IntSet` offer `union`, `intersection`, `difference`, but no [`symmetricDifference`](https://en.wikipedia.org/wiki/Symmetric_difference). I don't see why it should be left out. `Set` even offers `disjointUnion`.

See also: #732. This is about adding a generalized merge, but surely `symmetricDifference` deserves to exist without or without such a merge function.

If `Set` and `IntSet` get these, it would be consistent to add them for `Map` and `IntMap` too.

meooow25:
...for Set, Map, IntSet, IntMap. This joins the set operations already implemented: union, intersection, difference.

Closes #1002.

### DIFF
--- a/containers-tests/benchmarks/SetOperations/SetOperations-IntMap.hs
+++ b/containers-tests/benchmarks/SetOperations/SetOperations-IntMap.hs
@@ -3,4 +3,10 @@ module Main where
 import Data.IntMap as C
 import SetOperations
 
-main = benchmark (\xs -> fromList [(x, x) | x <- xs]) True [("union", C.union), ("difference", C.difference), ("intersection", C.intersection)]
+main :: IO ()
+main = benchmark (\xs -> fromList [(x, x) | x <- xs]) True
+  [ ("union", C.union)
+  , ("difference", C.difference)
+  , ("intersection", C.intersection)
+  , ("symmetricDifference", C.symmetricDifference)
+  ]
--- a/containers-tests/benchmarks/SetOperations/SetOperations-IntSet.hs
+++ b/containers-tests/benchmarks/SetOperations/SetOperations-IntSet.hs
@@ -3,4 +3,10 @@ module Main where
 import Data.IntSet as C
 import SetOperations
 
-main = benchmark fromList True [("union", C.union), ("difference", C.difference), ("intersection", C.intersection)]
+main :: IO ()
+main = benchmark fromList True
+  [ ("union", C.union)
+  , ("difference", C.difference)
+  , ("intersection", C.intersection)
+  , ("symmetricDifference", C.symmetricDifference)
+  ]
--- a/containers-tests/benchmarks/SetOperations/SetOperations-Map.hs
+++ b/containers-tests/benchmarks/SetOperations/SetOperations-Map.hs
@@ -3,4 +3,10 @@ module Main where
 import Data.Map as C
 import SetOperations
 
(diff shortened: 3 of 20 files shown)


meooow25:
@treeowl any concerns about adding these operations?

```hs
symmetricDifference :: Ord a => Set a -> Set a -> Set a
symmetricDifference :: Ord k => Map k a -> Map k a -> Map k a
symmetricDifference :: IntSet -> IntSet -> IntSet
symmetricDifference :: IntMap a -> IntMap a -> IntMap a
```

treeowl:
@jwaldmann , I would look to `mergeA` instead. Is it powerful enough to implement this? I suspect so. Does it perform well enough for this? That's less clear to me.

meooow25:
> It's still an extra maintenance burden.

I don't think that is the case. These are straightforward implementations. We already have `union`, `intersection`, `difference` implemented in this manner, and `symmetricDifference` completes the list.

> Can these functions not be implemented with https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Strict.html#v:mergeWithKey (and similar)? Did you do, did you compare performance?

> I would look to mergeA instead.

These operations would more likely be used for sets (as mentioned in #1002), and there is no `merge`/`mergeA` function for `Set` and `IntSet`.

For `Map` and `IntMap`, yes `symmetricDifference` could be implemented with `merge`/`mergeA`. I have not compared the performance, but from a manual inspection it can only be slower. For instance,

* `Map`'s `mergeA` uses `splitLookup` which allocates and returns a `Maybe a`. This can be avoided in `symmetricDifference` by using `splitMember` because we have no use for that `a`.
  https://github.com/haskell/containers/blob/edc0e149479dbd0d5f41115c109dba699561ee0e/containers/src/Data/Map/Internal.hs#L2694
* `IntMap`'s `mergeA` is quite involved and is much larger than a direct implementation code-size wise. It also has some branches such as
  https://github.com/haskell/containers/blob/edc0e149479dbd0d5f41115c109dba699561ee0e/containers/src/Data/IntMap/Internal.hs#L2072-L2076
  https://github.com/haskell/containers/blob/edc0e149479dbd0d5f41115c109dba699561ee0e/containers/src/Data/IntMap/Internal.hs#L2109-L2110
  which are unnecessary for `symmetricDifference`.

treeowl:
I don't object to adding these. I do want to get `mergeA` for sets, but there's no way I'll get to it myself anytime very soon.

meooow25:
@treeowl should we merge this then?

treeowl:
Yeah, I'd say go ahead.

treeowl:
>     * `IntMap`'s `mergeA` is quite involved and is much larger than a direct implementation code-size wise. It also has some branches such as
>       https://github.com/haskell/containers/blob/edc0e149479dbd0d5f41115c109dba699561ee0e/containers/src/Data/IntMap/Internal.hs#L2072-L2076
>       
>       https://github.com/haskell/containers/blob/edc0e149479dbd0d5f41115c109dba699561ee0e/containers/src/Data/IntMap/Internal.hs#L2109-L2110
>       
>       which are unnecessary for `symmetricDifference`.

If those branches are a performance issue, we should probably rewrite `merge` to avoid them; they're only needed for non-commutative functors.

meooow25:
That's true, we need them in `mergeA` but `merge` could be implemented without them.

PR #1020, opened 2024-08-10
Delete findWithDefault for Strict Map/IntMap

meooow25:
They are redundant because they have the same strictness characteristics as the lazy versions.


treeowl (approved):
LGTM

Issue #1019, opened 2024-08-08
Improve strictness tests


meooow25:
Strictness tests need to be in good shape to avoid bugs like #996

meooow25:
Just started and I'm seeing things like

```hs
λ> import qualified Data.Map.Strict as M
λ> M.fromList [(1,undefined),(1,2)]
fromList *** Exception: Prelude.undefined
λ> M.fromAscList [(1,undefined),(1,2)]
fromList [(1,2)]
```

Hope there aren't too many of these...

treeowl:
Uh oh.... I assume that second one extracted to a `Set`?

meooow25:
I'm not sure what you mean. It's just a consequence of the different implementations.

meooow25:
Looking around at the tests we have today, we seem to be using two libraries:

* `ChasingBottoms` to detect bottom values
* `nothunks` to detect thunks

We can test strictness by detecting bottoms, but the `nothunks` machinery seems massively more powerful than what we need. `nothunks` is intended to find thunks in nested data by exploring it in the manner of `NFData`. We don't have such nested data in containers, it's just about whether maps contain bottom values or whether strict folds indeed fold strictly. I think we could remove this dependency if we wanted.

We can perhaps even inline the bits of `ChasingBottoms` we use and remove that too. But I'll leave that for later.

PR #963, opened 2023-07-16
Improve {Set,Map}.fromAscList and friends

meooow25:
Following up on https://github.com/haskell/containers/issues/949#issuecomment-1574528332

This PR is based off #962, so marking as draft.

This PR improves
* `Set.fromAscList`, `Set.fromDescList`
* `Map.fromAscList`, `Map.fromDescList`, `Map.fromAscListWith`, `Map.fromDescListWith`, `Map.fromAscListWithKey`, `Map.fromDescListWithKey`

by making the combine function they use a good consumer and producer in terms of fusion. This gets rid of the intermediate list always created today, and also fuses with the input list if possible.

---

Benchmarks on GHC 9.2.5:

<details>
<summary>Before</summary>

```
Set
  fromAscList:         OK (0.20s)
    50.1 μs ± 2.9 μs, 240 KB allocated, 2.4 KB copied, 8.0 MB peak memory
  fromAscList:fusion:  OK (0.19s)
    89.5 μs ± 5.7 μs, 528 KB allocated, 6.8 KB copied, 8.0 MB peak memory
  fromDescList:        OK (0.22s)
    51.2 μs ± 2.9 μs, 240 KB allocated, 2.4 KB copied, 8.0 MB peak memory
  fromDescList:fusion: OK (0.18s)
    88.9 μs ± 5.7 μs, 559 KB allocated, 7.3 KB copied, 8.0 MB peak memory

Map
  fromAscList:         OK (0.18s)
    41.9 μs ± 3.6 μs, 343 KB allocated, 5.2 KB copied,  10 MB peak memory
  fromAscList:fusion:  OK (0.13s)
    63.0 μs ± 6.0 μs, 792 KB allocated,  15 KB copied,  10 MB peak memory
  fromDescList:        OK (0.17s)
    40.5 μs ± 3.9 μs, 343 KB allocated, 5.5 KB copied,  10 MB peak memory
  fromDescList:fusion: OK (0.25s)
    63.1 μs ± 5.3 μs, 824 KB allocated,  16 KB copied,  10 MB peak memory
```

</details>

After

```
Set
  fromAscList:         OK (0.18s)
    21.3 μs ± 1.4 μs, 128 KB allocated, 1.4 KB copied, 8.0 MB peak memory, 57% less than baseline
  fromAscList:fusion:  OK (0.89s)
    13.3 μs ± 221 ns, 144 KB allocated, 1.9 KB copied, 8.0 MB peak memory, 85% less than baseline
[...]

PR #1021, opened 2024-08-11
Strictness tests for Map construction

meooow25:
For #1019.

Adds a bunch of property tests that try to create strict and lazy `Map`s with bottom values and check if the result map is bottom.

We also need to do `IntMap` but this PR is already huge, will do that later.

### DIFF
--- a/containers-tests/tests/Utils/Strictness.hs
+++ b/containers-tests/tests/Utils/Strictness.hs
@@ -1,0 +1,116 @@
+module Utils.Strictness
+  ( Bot(..)
+  , Func
+  , applyFunc
+  , Func2
+  , applyFunc2
+  , Func3
+  , applyFunc3
+  ) where
+
+import Test.ChasingBottoms.IsBottom (isBottom)
+import Test.QuickCheck
+
+{--------------------------------------------------------------------
+  Bottom stuff
+--------------------------------------------------------------------}
+
+-- | Arbitrary (Bot a) values may be bottom.
+newtype Bot a = Bot a
+
+instance Show a => Show (Bot a) where
+  show (Bot x) = if isBottom x then "<bottom>" else show x
+
+instance Arbitrary a => Arbitrary (Bot a) where
+  arbitrary = frequency
+    [ (1, pure (error "<bottom>"))
+    , (4, Bot <$> arbitrary)
+    ]
+
+{--------------------------------------------------------------------
+  Lazy functions
+--------------------------------------------------------------------}
+
+-- | A function which may be lazy in its argument.
+--
+-- Either ignores its argument, or uses a QuickCheck Fun (which is always a
+-- strict function).
+data Func a b
+  = FuncLazy b
+  | FuncStrict (Fun a b)
+
+instance (Show a, Show b) => Show (Func a b) where
+  show (FuncLazy x) = "{_lazy->" ++ show x ++ "}"
+  show (FuncStrict fun) = show fun
+
+applyFunc :: Func a b -> a -> b
+applyFunc fun x = case fun of
+  FuncLazy y -> y
+  FuncStrict f -> applyFun f x
+
+instance (CoArbitrary a, Function a, Arbitrary b) => Arbitrary (Func a b) where
+  arbitrary = frequency
+    [ (1, FuncLazy <$> arbitrary)
+    , (4, FuncStrict <$> arbitrary)
+    ]
+
+  shrink fun = case fun of
+    FuncLazy x -> FuncLazy <$> shrink x
+    FuncStrict f -> FuncStrict <$> shrink f
+
+-- | A function which may be lazy in its arguments.
+
+-- Note: We have two separate cases here because we want to generate functions
+-- of type `a -> b -> c` with all possible strictness configurations.
+-- `Func a (Func b c)` is not enough for this, since it cannot generate
+-- functions that are conditionally lazy in the first argument, for instance:
+--
+-- leftLazyOr :: Bool -> Bool -> Bool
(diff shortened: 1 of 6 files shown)


meooow25:
Tests should pass once #473 and #1022 are fixed.

meooow25:
Alright, this is good to go.
Dunno if you want to take a look at this @treeowl.

containers-tests/tests/Utils/Strictness.hs
```diff
@@ -0,0 +1,103 @@
...
+
+  shrink fun = case fun of
+    FuncLazy x -> FuncLazy <$> shrink x
+    FuncStrict f -> FuncStrict <$> shrink f
+
+-- | Function which may be lazy in its arguments
+data Func2 a b c
+  = F2A (Func a (Func b c))
+  | F2B (Func b (Func a c))
+  deriving Show
```
treeowl, line 62: This seems to be designed to increase the probability of the function being lazy in the second argument, over just `Func a (Func b c)`, right? A comment would be a good idea.

containers-tests/tests/map-strictness.hs
```diff
@@ -26,14 +42,887 @@ instance (Arbitrary k, Arbitrary v, Ord k) =>
          Arbitrary (Map k v) where
     arbitrary = M.fromList `fmap` arbitrary
 
+instance (Arbitrary a, Ord a) => Arbitrary (Set a) where
+  arbitrary = Set.fromList <$> arbitrary
+
```
treeowl, line 47: This is no good. For our purposes here, we want to build all `Set` shapes, as we do in `Set` testing, so we can make sure the implemented balance invariants actually match properly. Best to factor out a module for the existing instance, or perhaps to write a more general, less efficient version of it.

treeowl (changes requested):
I haven't read it all yet, but I have a few thoughts.

containers-tests/tests/map-strictness.hs
meooow25: I agree on this and I tried to move the existing instances, but they have a weird `IsInt` constraint which means they cannot generate `OrdA`.

https://github.com/haskell/containers/blob/fbade40798546a7fb7cbabed1922ef7b0010ab1f/containers-tests/tests/set-properties.hs#L219

https://github.com/haskell/containers/blob/fbade40798546a7fb7cbabed1922ef7b0010ab1f/containers-tests/tests/map-properties.hs#L294

containers-tests/tests/map-strictness.hs
```diff
@@ -26,14 +42,887 @@ instance (Arbitrary k, Arbitrary v, Ord k) =>
...
+prop_strictSingleton k (Bot x) = isBottom (M.singleton k x) === isBottom x
+
+prop_lazySingleton :: OrdA -> Bot A -> Property
+prop_lazySingleton k (Bot x) = isNotBottomP (L.singleton k x)
+
+prop_strictFromSet :: Func OrdA (Bot A) -> Set OrdA -> Property
+prop_strictFromSet fun set =
+  isBottom (M.fromSet f set) === any (isBottom . f) (Set.toList set)
+  where
+    f = coerce (applyFunc fun) :: OrdA -> A
```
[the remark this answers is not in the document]
meooow25: > I think the explicit `coerce` here reduces clarity without (at least meaningfully) improving test performance.

In all of the tests with functions, the point is to get some `X -> Y` where `Y` may be bottom. This is done by generating `X -> Bot Y` and using `coerce . applyFunc` on it. It is true that some tests would work without the `coerce`, but I don't see a good reason to deviate from the standard pattern.

> Less importantly: I think `Bot A` is overkill; `Bot ()` should do the trick. That change should improve test performance by reducing the number of random numbers generated.

This is true for this test, but again there is little reason to deviate. The tests run fast and I don't think it is worth trying to optimize tests case-by-case unless it is really causing a problem.

meooow25:
@treeowl do you have more comments on this?

containers-tests/tests/map-strictness.hs
treeowl: We can port the shape generation without the performance-only type restriction. Generate a shape and a list to match.

treeowl: I'm nervous about putting in tests that don't test everything they seem to.

meooow25: Alright, I'll take a look. I haven't attempted to understand the tree generation and the constraint.

By the way, why is it not simply `Arbitrary a => Arbitrary (Set a)`?

meooow25: > We can port the shape generation without the performance-only type restriction. Generate a shape and a list to match.

Done.

> Okay, so ... it totally _could_ be `(Arbitrary a, Ord a) => Arbitrary (Set a)`. The only reason it isn't is that it's faster to generate `Int`s like that than to generate an ordered list of unique arbitrary items. It might be worth seeing how much slower test performance is with a more general instance. Test performance does matter, because the faster the individual tests run the more we can use without driving contributors crazy with CI times.

While I agree that we should care about test performance, this might be worrying out of proportion. If you look at any CI run, say [this one](https://github.com/haskell/containers/actions/runs/10474461204/job/29008781350), you will see that running tests takes only 17s out of the total 9m 50s.

meooow25:
@treeowl are you still reviewing this?

I would like to get the testing PRs submitted soon because they end up blocking other stuff (#1042, #963, similar IntMap tests, etc).

meooow25:
I'll check this once for problems and merge if it seems fine.

meooow25:
Alright, it's looking good.

PR #1025, opened 2024-08-13
Fix discrepancy between mergeWithKey impl and docs

### LINKED ISSUE #979: Data.Map.mergeWithKey doesn't match documentation
The documentation for `Data.Map.Lazy.mergeWithKey` states:

> - a nonempty subtree present only in the first map is passed to `only1` and the output is added to the result;
> - a nonempty subtree present only in the second map is passed to `only2` and the output is added to the result.

Yet this example shows otherwise:
```hs
M.mergeWithKey
  (\k x y -> traceShow ("merge", k, x, y) $ Just x)
  (\m -> traceShow ("left", m) M.empty)
  (\m -> traceShow ("right", m) M.empty)
  M.empty
  M.empty
  `seq` ()
```
```
("right",fromList [])
()
```

The same isssue exists in the Strict version, and in both IntMap versions.

meooow25:
The given functions should be called only on nonempty trees.

Fixes #979.

This could alternately be fixed by changing the docs. I find it slightly better to keep the specification that the user will not receive an empty tree.

### DIFF
--- a/containers/changelog.md
+++ b/containers/changelog.md
@@ -14,6 +14,11 @@
 * `Data.Map.Strict.mergeWithKey` now forces the result of the combining function
   to WHNF. (Soumik Sarkar)
 
+* Fix an issue where `Data.Map.mergeWithKey`, `Data.Map.Strict.mergeWithKey`,
+  `Data.IntMap.mergeWithKey`, `Data.IntMap.Strict.mergeWithKey` could call the
+  provided `only2` function with empty maps, contrary to documentation.
+  (Soumik Sarkar)
+
 ## Unreleased with `@since` annotation for 0.7.1:
 
 ### Additions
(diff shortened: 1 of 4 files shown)


treeowl (approved):
Seems reasonable. Hopefully no one is relying on the current behavior. You're also welcome to make another attempt at function removal, as far as I'm concerned.

meooow25:
> Hopefully no one is relying on the current behavior.

I suppose this does cause a change in behavior. Updated the changelog in case someone is affected.

> You're also welcome to make another attempt at function removal, as far as I'm concerned.

Personally I'm not too keen on seeing it gone. It is unsafe but can be used properly. I wish it were maybe exported from "Data.Map.Unsafe" or something like that.

Issue #1016, opened 2024-08-05
Faster Eq and Ord


meooow25:
The following operations are implemented by converting to lists and comparing.

- [x] Set `==` (improved in #1017)
- [x] Set `compare` (same as above)
- [x] Map `==` (same as above)
- [x] Map `compare` (same as above)
- [x] IntSet `compare` (#1086)
- [x] IntMap `compare` (same as above)
- [x] Seq `==` (#1035)
- [x] Seq `compare` (same as above)

This is not the best we can do for a couple of reasons:

1. No specialization. `Eq` and `Ord` for lists don't specialize (but there a few manually specified exceptions).
2. No list fusion. We materialize both lists.

See list instances [Eq](https://gitlab.haskell.org/ghc/ghc/-/blob/cf47b96f8b8100d9f9ad1b8121cbfb65fd7eaf2a/libraries/ghc-prim/GHC/Classes.hs#L261-267), [Ord](https://gitlab.haskell.org/ghc/ghc/-/blob/cf47b96f8b8100d9f9ad1b8121cbfb65fd7eaf2a/libraries/ghc-prim/GHC/Classes.hs#L482-491).

---

I tried a couple of alternatives and saw decent improvements

```
  Set Word
    compare: OK
      3.03 ms ± 189 μs,  11 MB allocated, 1.8 MB copied,  23 MB peak memory
    setCmp:  OK
      1.64 ms ± 143 μs, 8.4 MB allocated, 583 KB copied,  23 MB peak memory, 0.54x
    setCmp2: OK
      947  μs ±  68 μs, 5.3 MB allocated, 795 B  copied,  23 MB peak memory, 0.31x
```

```hs
-- Current version but fused with the first set
setCmp :: Ord a => Set a -> Set a -> Ordering
setCmp xs ys0 = S.foldr f z xs (S.toList ys0)
  where
    f x k = oneShot $ \ys -> case ys of
      [] -> GT
      y:ys' -> compare x y <> k ys'
    z ys = case ys of
      [] -> EQ
      _ -> LT
{-# INLINABLE setCmp #-}
```

```hs
-- Somewhat like a traverse with StateT [a] (Either Bool)
setCmp2 :: Ord a => Set a -> Set a -> Ordering
setCmp2 xs ys0 = case go xs (S.toList ys0) of
  WithO o ys' ->
    o <> case ys' of
      [] -> EQ
      _ -> LT
  where
    go Tip ys = WithO EQ ys
[...]

PR #1028, opened 2024-08-17
Update Eq IntSet and Eq (IntMap a)

meooow25:
* Add benchmarks
* Mark the function INLINABLE for IntMap. On benchmarks for IntMap Int this reduces the run time by ~22%.
* Remove definitions for (/=). There is no good reason to implement this manually.

Related to #1016.

---

Benchmarks on GHC 9.6.3

```
IntMap before
  eq: OK
    39.0 μs ± 3.4 μs,   0 B  allocated,   0 B  copied,  10 MB peak memory
IntMap after
  eq: OK
    30.2 μs ± 3.0 μs,   0 B  allocated,   0 B  copied,  10 MB peak memory, 22% less than baseline
```

### DIFF
--- a/containers-tests/benchmarks/IntSet.hs
+++ b/containers-tests/benchmarks/IntSet.hs
@@ -56,6 +56,7 @@ main = do
         , bench "split:sparse" $ whnf (IS.split elem_sparse_mid) s_sparse
         , bench "splitMember:dense" $ whnf (IS.splitMember elem_mid) s
         , bench "splitMember:sparse" $ whnf (IS.splitMember elem_sparse_mid) s_sparse
+        , bench "eq" $ whnf (\s' -> s' == s') s
         ]
   where
     bound = 2^12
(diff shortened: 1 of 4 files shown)


containers-tests/benchmarks/IntSet.hs
```diff
@@ -56,6 +56,7 @@ main = do
         , bench "split:sparse" $ whnf (IS.split elem_sparse_mid) s_sparse
         , bench "splitMember:dense" $ whnf (IS.splitMember elem_mid) s
         , bench "splitMember:sparse" $ whnf (IS.splitMember elem_sparse_mid) s_sparse
+        , bench "eq" $ whnf (\s' -> s' == s') s
```
treeowl, line 59: Some comment seems necessary to indicate that this is (only) testing the worst case.

treeowl (changes requested):

containers-tests/benchmarks/IntSet.hs
meooow25: Makes sense, done.

treeowl (approved):

Issue #1018, opened 2024-08-08
Tidy up IntSet/IntMap bitwise stuff


meooow25:
* [ ] Check if `Int`<->`Word` can be avoided (from https://github.com/haskell/containers/pull/995#discussion_r1542147809)
* [x] Improve some stuff such as `bitcount` and delete obsolete comments (from https://github.com/haskell/containers/pull/998#discussion_r1554569460)

PR #1037, opened 2024-09-12
Remove indexOfTheOnlyBit in IntSet.Internal

meooow25:
* This was used for 32-bit GHC, which isn't necessary
* Drop a CPP check for 32/64-bit GHC, GHC doesn't support anything else

Related: #1018

I can't tell why the special case for 32-bit GHC exists. #379 started using the GHC functions but I can't find an explanation for why only on 64-bit. @treeowl do you know?

This PR _should_ improve performance on 32-bit systems, don't know if we want to test that. I don't have a 32-bit system.


treeowl (approved):
I think there may have been some reluctance to change the implementation on 32-bit without benchmarking on 32-bit, bat. I think the change is good anyway.

PR #1039, opened 2024-09-13
Update references

### LINKED ISSUE #1034: Link to "Efficient sets: a balancing act" is rotten
https://github.com/haskell/containers/blob/2ad4ea3123d0d5116d02e8e47f0210117ac256a0/containers/src/Data/Set.hs#L51-L53

I cannot find anything relevant at https://groups.csail.mit.edu/mac/projects/mac/

meooow25:
* Add or update links
* Correct some titles
* Add a reference to "Balancing weight-balanced trees"

Notes:
* The "Fast Mergeable Integer Maps" link I've added is actually dead. However, it is the best link I could find. The contents have been archived on https://web.archive.org.
* I've added a reference to "Balancing weight-balanced trees", which goes into the choice of delta/ratio we use for Set/Map and proves that it preserves balance.

Fixes #1034.

### DIFF
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -36,8 +36,11 @@
 --
 -- The implementation is based on
 --
---   * /Structuring Depth-First Search Algorithms in Haskell/,
---     by David King and John Launchbury, <http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.52.6526>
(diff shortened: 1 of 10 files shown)


Bodigrim:
> The contents have been archived on https://web.archive.org.

Why not link WebArchive then?

meooow25:
I suppose we could.

PR #1040, opened 2024-09-14
Update and export set intersection utilities

### LINKED ISSUE #1038: Export Set intersections and Intersection
#756 added these two definitions

https://github.com/haskell/containers/blob/028e187120c2cd6ed9f74d67a03bec1431f74be3/containers/src/Data/Set/Internal.hs#L897-L911

but they were never exported from `Data.Set`.

We should export them, but there is an opportunity for improvement today. `base-4.18` got [`Foldable1`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Foldable1.html), so we could change `intersections` to be `(Ord a, Foldable1 f) => f (Set a) -> Set a`.

And whatever we do for Set we should also do for IntSet.

meooow25:
* Update Set.intersections to use Foldable1 and be lazier
* Mark definitions INLINABLE for specialization
* Add Intersection and intersections for IntSet
* Add property tests

Fixes #1038

### DIFF
--- a/containers-tests/tests/intset-properties.hs
+++ b/containers-tests/tests/intset-properties.hs
@@ -7,6 +7,11 @@ import Data.List (nub,sort)
 import qualified Data.List as List
 import Data.Maybe (listToMaybe)
 import Data.Monoid (mempty)
+#if MIN_VERSION_base(4,18,0)
+import Data.List.NonEmpty (NonEmpty(..))
+import qualified Data.List.NonEmpty as NE
+import qualified Data.Foldable1 as Foldable1
+#endif
 import qualified Data.Set as Set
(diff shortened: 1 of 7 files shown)


meooow25:
@treeowl what is our stance on conditional exports?  
`intersections` is exported here if `Foldable1` is available in base.

Is this controversial?

If so, I will change it to use `NonEmpty`. But this means a future breaking change of switching to `Foldable1`.

meooow25:
Looks like this would be in violation of the PVP FAQ: https://pvp.haskell.org/faq/#what-implications-does-the-pvp-have-for-conditional-apis

I will make the change to use `NonEmpty`.

PR #1043, opened 2024-09-19
Various changes to make it compile with MicroHs.

augustss:
Mostly #ifdefs and some type signatures.

### DIFF
--- a/.github/workflows/mhs-ci.yml
+++ b/.github/workflows/mhs-ci.yml
@@ -1,0 +1,31 @@
+name: MicroHs CI for containers
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build-mhs-containers:
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout containers repo
+      uses: actions/checkout@v4
+      with:
+        path: cont
+    - name: checkout mhs repo
+      uses: actions/checkout@v4
+      with:
+        repository: augustss/MicroHs
+        path: mhs
+    - name: make mhs
+      run: |
+        cd mhs
+        make
+# It's pretty ugly with the list of modules here, but I don't know a nice way of getting it from the cabal file.
+# I'll make it nicer with mcabal later.
+    - name: compile containers package
+      run: |
+        cd mhs
+        MHSCPPHS=./bin/cpphs ./bin/mhs -Pcontainers-test -ocontainers-test.pkg -i../cont/containers/src -XCPP -I../cont/containers/include Data.Containers.ListUtils Data.IntMap Data.IntMap.Lazy Data.IntMap.Strict Data.IntMap.Strict.Internal Data.IntMap.Internal Data.IntMap.Internal.Debug Data.IntMap.Merge.Lazy Data.IntMap.Merge.Strict Data.IntSet.Internal Data.IntSet.Internal.IntTreeCommons Data.IntSet Data.Map Data.Map.Lazy Data.Map.Merge.Lazy Data.Map.Strict.Internal Data.Map.Strict Data.Map.Merge.Strict Data.Map.Internal Data.Map.Internal.Debug Data.Set.Internal Data.Set Data.Graph Data.Sequence Data.Sequence.Internal Data.Sequence.Internal.Sorting Data.Tree Utils.Containers.Internal.BitUtil Utils.Containers.Internal.BitQueue Utils.Containers.Internal.StrictPair
--- a/.github/workflows/mhs-ci.yml
+++ b/.github/workflows/mhs-ci.yml
@@ -17,7 +17,8 @@ jobs:
     - name: checkout mhs repo
       uses: actions/checkout@v4
       with:
-        repository: augustss/MicroHs@stable-1
+        repository: augustss/MicroHs
+        ref: stable-1
         path: mhs
     - name: make mhs
       run: |
--- a/.github/workflows/mhs-ci.yml
+++ b/.github/workflows/mhs-ci.yml
@@ -17,7 +17,7 @@ jobs:
     - name: checkout mhs repo
       uses: actions/checkout@v4
       with:
-        repository: augustss/MicroHs
+        repository: augustss/MicroHs@stable-1
         path: mhs
     - name: make mhs
       run: |
--- a/containers/include/containers.h
+++ b/containers/include/containers.h
@@ -6,7 +6,7 @@
 #define HASKELL_CONTAINERS_H
 
 /*
- * On GHC, include MachDeps.h to get WORD_SIZE_IN_BITS macro.
+ * On GHC and MicroHs, include MachDeps.h to get WORD_SIZE_IN_BITS macro.
  */
 #if defined(__GLASGOW_HASKELL__) || defined(__MHS__)
 #include "MachDeps.h"
--- a/containers/include/containers.h
+++ b/containers/include/containers.h
@@ -8,7 +8,7 @@
 /*
  * On GHC, include MachDeps.h to get WORD_SIZE_IN_BITS macro.
  */
-#ifdef __GLASGOW_HASKELL__
+#if defined(__GLASGOW_HASKELL__) || defined(__MHS__)
 #include "MachDeps.h"
 #endif
 
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -115,7 +115,7 @@ import Data.Tree (Tree(Node), Forest)
 
 -- std interfaces
 import Data.Foldable as F
-#if MIN_VERSION_base(4,18,0) && defined(__GLASGOW_HASKELL__)
+#if MIN_VERSION_base(4,18,0)
 import qualified Data.Foldable1 as F1
 #endif
 import Control.DeepSeq (NFData(rnf))
@@ -130,9 +130,7 @@ import qualified Data.Array as UA
 import qualified Data.List as L
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NE
-#ifdef __GLASGOW_HASKELL__
 import Data.Functor.Classes
-#endif
 #if !MIN_VERSION_base(4,11,0)
 import Data.Semigroup (Semigroup (..))
 #endif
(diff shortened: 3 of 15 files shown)


treeowl:
Thank you! This is a very welcome development. I don't see any changes to CI. Could we at least compile the library on MicroHS in CI, even if dependencies don't yet let us compile the test suite?

augustss:
Hmm.  That's a good idea.  I will have to look into how easy it is.
Can CI compile other Hackage packages?

containers/include/containers.h
```diff
@@ -8,7 +8,7 @@
 /*
  * On GHC, include MachDeps.h to get WORD_SIZE_IN_BITS macro.
```
treeowl, line 9:
```suggestion
 * On GHC and MicroHs, include MachDeps.h to get WORD_SIZE_IN_BITS macro.
```

meooow25:
What exactly are the differences between getting something to compile on GHC and MicroHs. It seems to be mostly about certain modules not available in MicroHs `base`?

And +1 on having a CI so we don't break this in the future.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -328,8 +330,8 @@ import Text.Read
 import Language.Haskell.TH.Syntax (Lift)
 -- See Note [ Template Haskell Dependencies ]
 import Language.Haskell.TH ()
-#endif
 import qualified Control.Category as Category
+#endif
```
treeowl, line 334: Does MicroHs not support `Control.Category`? It's quite an old API; wasn't it supported by implementations other than GHC in the past?

meooow25:
By the way, I am having to manually approve these CI runs. It seems that this is the default for "first time contributors" and I don't see a way to change it. treeowl could you peek at the settings to see if we can change this (perhaps temporarily) to make things easier?

augustss:
Yes, getting it to run with MHS is mostly a matter of #ifdef-ing out things that base does not support (and cannot support right now, since it needs extensions like QualifiedConstraints).
There's also a semantics difference to the scoping of type variables in signatures.  I might change that for easier porting.

treeowl:
Would it make sense on MicroHs to have a local copyish of `Control.Category` to limit the CPP and offer that interface more universally?

containers/src/Data/IntMap/Internal.hs
augustss: There's no Control.Category, but that's just my laziness.

containers/src/Data/Map/Internal.hs
```diff
@@ -4327,11 +4333,16 @@ sameSizeLiftEq2 keq eq m1 m2 =
       Nothing -> False :*: it
       Just (KeyValue ky y :*: it') -> (keq kx ky && eq x y) :*: it'
 {-# INLINE sameSizeLiftEq2 #-}
+#else
+instance (Eq k,Eq a) => Eq (Map k a) where
+  t1 == t2  = (size t1 == size t2) && (toAscList t1 == toAscList t2)
+#endif
```
meooow25, line 4339: I'd prefer if we did not have entirely separate definitions (without good reason).  
The trouble here is no Eq1/Eq2, so we can instead lift the `liftEq2` definition to the top level and use it for MHS.  
Same for Ord1/Ord2 and other similar instances.

containers/src/Data/Graph.hs
```diff
@@ -115,7 +115,7 @@ import Data.Tree (Tree(Node), Forest)
 
 -- std interfaces
 import Data.Foldable as F
-#if MIN_VERSION_base(4,18,0)
+#if MIN_VERSION_base(4,18,0) && defined(__GLASGOW_HASKELL__)
 import qualified Data.Foldable1 as F1
```
meooow25, line 119: Foldable1 too doesn't require quantified constraints, in case you want to have it in MHS.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2067,7 +2071,7 @@ merge g1 g2 f m1 m2 =
 --
 -- @since 0.5.9
 mergeA
-  :: (Applicative f)
+  :: forall f a b c . (Applicative f)
```
treeowl, line 2074: Could you explain why we need this change, and the type signature below?

meooow25 (commented):
If QuantifiedConstraints is a problem, MicroHs could define the pre-QuantifiedConstraints versions of the classes (Eq1, Ord1, Bifoldable, etc). This would let us get rid of most of the CPP.

augustss:
Just for your amusement: containers has 25kloc of Haskell code and MHS has 10kloc. :)

containers/src/Data/IntMap/Internal.hs
treeowl: Yeah, that seems like a surprising choice. FWIW, GHC seems to be going in the opposite direction of making that even more explicit.

augustss:
I'm making the MHS base library more GHC compatible. The PR will be updated soon.

treeowl:
Don't compromise your principles too much, but the less CPP we have to deal with the better.

augustss:
I've made my changes.  Please have a look again.

.github/workflows/mhs-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout containers repo
+      uses: actions/checkout@v4
+      with:
+        path: cont
+    - name: checkout mhs repo
+      uses: actions/checkout@v4
+      with:
+        repository: augustss/MicroHs
```
meooow25, line 20: Should this check out a particular version?
We wouldn't want containers CI to break in there is an issue with MHS head.
We can update the version periodically, as we do for GHC.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -324,11 +324,11 @@ import Data.Data (Data(..), Constr, mkConstr, constrIndex,
 import qualified Data.Data as Data
 import GHC.Exts (build)
 import qualified GHC.Exts as GHCExts
-import Text.Read
 import Language.Haskell.TH.Syntax (Lift)
 -- See Note [ Template Haskell Dependencies ]
 import Language.Haskell.TH ()
 #endif
+import Text.Read
```
meooow25, line 331: We should use `#if defined(__GLASGOW_HASKELL__) || defined(__MHS__)` for this, to reflect the Read instance.

containers/src/Data/Map/Internal.hs
```diff
@@ -414,12 +414,11 @@ import Language.Haskell.TH ()
 import GHC.Exts (Proxy#, proxy# )
 #  endif
 import qualified GHC.Exts as GHCExts
-import Text.Read hiding (lift)
 import Data.Data
-import qualified Control.Category as Category
 import Data.Coerce
 #endif
-
+import Text.Read hiding (lift)
```
meooow25, line 420: Same here.

containers/src/Data/Sequence/Internal.hs
```diff
@@ -220,28 +220,31 @@ import Data.Functor.Classes
...
 import GHC.Generics (Generic, Generic1)
+
+import qualified GHC.Arr
+import Data.Coerce
+import qualified GHC.Exts
+#else
+import qualified Data.List
 #endif
 
 -- Array stuff, with GHC.Arr on GHC
```
meooow25, line 243: Can `import qualified GHC.Arr` be moved back, to keep this comment valid?

containers/src/Data/Sequence/Internal.hs
```diff
@@ -220,28 +220,31 @@ import Data.Functor.Classes
...
 import qualified Language.Haskell.TH.Syntax as TH
 -- See Note [ Template Haskell Dependencies ]
 import Language.Haskell.TH ()
 import GHC.Generics (Generic, Generic1)
+
+import qualified GHC.Arr
+import Data.Coerce
+import qualified GHC.Exts
+#else
+import qualified Data.List
```
meooow25, line 240: Is this import used somewhere?

containers/src/Data/Tree.hs
```diff
@@ -233,7 +234,7 @@ instance Foldable Tree where
     product = foldlMap1' id (*)
     {-# INLINABLE product #-}
 
-#if MIN_VERSION_base(4,18,0)
+#if MIN_VERSION_base(4,18,0) && (defined(__GLASGOW_HASKELL__) || defined(__MHS__))
```
meooow25, line 237: We can drop the compiler checks here. Let's assume Foldable1 will be around, or we'd be doing the same for Eq1, Ord1, etc.

containers/src/Utils/Containers/Internal/Prelude.hs
```diff
@@ -1,12 +1,25 @@
+{-# LANGUAGE CPP #-}
 -- | This hideous module lets us avoid dealing with the fact that
 -- @liftA2@ and @foldl'@ were not previously exported from the standard prelude.
 module Utils.Containers.Internal.Prelude
   ( module Prelude
   , Applicative (..)
   , Foldable (..)
+#ifdef __MHS__
+  , Traversable(..)
+  , NonEmpty
```
meooow25, line 10: NonEmpty seems a little odd, is it required?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2112,6 +2114,7 @@ mergeA
       EQL -> binA p1 (go l1 l2) (go r1 r2)
       NOM -> linkA (unPrefix p1) (g1t t1) (unPrefix p2) (g2t t2)
 
+    subsingletonBy :: Functor f => (Key -> a -> f (Maybe c)) -> Key -> a -> f (IntMap c)
```
meooow25, line 2117: Is the signature required for MHS?

meooow25 (changes requested):
Thanks for the updates, this is looking much nicer!

containers/src/Data/Sequence/Internal.hs
meooow25: Ah, it's used in `fromList`'s `map_elem`.

containers/src/Data/IntMap/Internal.hs
meooow25: I assume that's a yes then.

meooow25 (approved):

PR #1057, opened 2024-10-26
Implement toList and toNonEmpty for SCC

meooow25:
The default implementations perform an avoidable list copy for NECyclicSCC. Also
* ~~Implement null, which is always False~~
* Fix flattenSCC being made too lazy accidentally in a previous commit

### DIFF
--- a/containers/src/Data/Graph.hs
+++ b/containers/src/Data/Graph.hs
@@ -220,11 +220,16 @@ instance F.Foldable SCC where
   foldr c n (AcyclicSCC v) = c v n
   foldr c n (NECyclicSCC vs) = foldr c n vs
 
+  toList = flattenSCC
+
 #if MIN_VERSION_base(4,18,0)
 -- | @since 0.7.0
 instance F1.Foldable1 SCC where
   foldMap1 f (AcyclicSCC v) = f v
   foldMap1 f (NECyclicSCC vs) = F1.foldMap1 f vs
+
+  toNonEmpty = flattenSCC1
+
   -- TODO define more methods
 #endif
 
@@ -258,7 +263,9 @@ flattenSCCs = concatMap flattenSCC
 -- This function is retained for backward compatibility,
 -- 'flattenSCC1' has the more precise type.
 flattenSCC :: SCC vertex -> [vertex]
-flattenSCC = NE.toList . flattenSCC1
+flattenSCC (AcyclicSCC v) = [v]
+flattenSCC (NECyclicSCC (v :| vs)) = v : vs
+-- Note: Best to avoid NE.toList, it is too lazy.
 
 -- | The vertices of a strongly connected component.
 --
--- a/containers/changelog.md
+++ b/containers/changelog.md
@@ -53,6 +53,11 @@
 
 * Add `foldMap` for `Data.IntSet`. (Soumik Sarkar)
 
+### Performance improvements
+
+* For `Data.Graph.SCC`, `Foldable.toList` and `Foldable1.toNonEmpty` now
+  do not perform a copy. (Soumik Sarkar)
+
 ## Unreleased with `@since` annotation for 0.7.1:
 
 ### Additions


meooow25:
The laziness bug slipped in in #987.
Note to self: Be extra careful whenever any Data.List.NonEmpty function is involved (https://github.com/haskell/core-libraries-committee/issues/107)

containers/src/Data/Graph.hs
```diff
@@ -258,7 +265,8 @@ flattenSCCs = concatMap flattenSCC
 -- This function is retained for backward compatibility,
 -- 'flattenSCC1' has the more precise type.
 flattenSCC :: SCC vertex -> [vertex]
-flattenSCC = NE.toList . flattenSCC1
+flattenSCC (AcyclicSCC v) = [v]
+flattenSCC (NECyclicSCC vs) = NE.toList vs
```
treeowl, line 269: Ah, so this one won't be too lazy because GHC knows that `NECyclicSCC` is a strict constructor, and `NE.toList` will inline. Subtle. Might be worth a comment.

containers/src/Data/Graph.hs
```diff
@@ -220,11 +220,18 @@ instance F.Foldable SCC where
   foldr c n (AcyclicSCC v) = c v n
   foldr c n (NECyclicSCC vs) = foldr c n vs
 
+  null _ = False
```
treeowl, line 223: I imagine GHC can figure this out from the default, but I agree it's nice to make it clear.

containers/src/Data/Graph.hs
```diff
@@ -220,11 +220,18 @@ instance F.Foldable SCC where
...
+
+  toList = flattenSCC
```
treeowl, line 225: Very good catch.

treeowl:
I've said it before and I'll say it again: `NonEmpty` is the wrong default representation for a non-empty list in a lazy-by-default language. Having conversions to and from standard lists be so cheap is a cute party trick, but actually working with the things is one painful judgement call after another.

containers/src/Data/Graph.hs
meooow25: I simply restored the previous definition. But it's probably better to be explicit and add a comment. Done.

containers/src/Data/Graph.hs
meooow25: Apparently it doesn't. However, I realized that we're making `null` lazier here. The default definition using `foldr` would need to inspect the `SCC`. Is this the right thing to do? I'm not sure, though many `Foldable` instances in base do this.

meooow25:
Isn't this just a case of functions being too lazy? I don't see a problem with the definition of the type. What would you prefer?

containers/src/Data/Graph.hs
treeowl: Ugh. I hate those kinds of decisions. I think the change is probably fine, but must be logged. `null` isn't normally used to force computations.

treeowl:
> Isn't this just a case of functions being too lazy? I don't see a problem with the definition of the type. What would you prefer?

Modulo naming, I would personally prefer 

```haskell
data NE a = End a | a :< NE a

instance Foldable1 NE where
  -- foldrMap1 :: (a -> b) -> (a -> b -> b) -> NE a -> b
  foldrMap1 e _c (End a) = e a
  foldrMap1 e c (a :< as) = a `c` foldrMap1 e c as
```

There would still be judgement calls for some things, but I suspect there would be more "natural" defaults. There are two options for `uncons`:

```haskell
uncons1, uncons2 :: NE a -> (a, Maybe (NE a))

uncons1 (End a) = (a, Nothing) 
uncons1 (a :< as) = (a, Just as)

uncons2 q = (a, r)
  where
    (a, r) = uncons1 q
```

Doesn't `uncons1` feel more obvious, with `uncons2` going above and beyond? With `NonEmpty`, there's a *nested* judgement call.

Zips are a bit more subtle, because we have the option of being somewhat lazy in the second list. Let's recons for convenience:

```haskell
recons :: a -> Maybe (NE a) -> NE a
recons a Nothing = End a
recons a (Just as) = a :< as

zipWith1 f (uncons1 -> (a, mas)) (uncons1 -> (b, mbs)) = recons (f a b) $ do
  as <- mas
  bs <- mbs
  Just $ zipWith1 f as bs

zipWith2 f (uncons1 -> (a, mas)) (uncons1 -> ~(b, mbs)) = recons (f a b) $ do
  as <- mas
  bs <- mbs
  Just $ zipWith2 f as bs
```

`zipWith1` feels much more natural to me. `zipWith2` has "short circuiting" like list zips, but if the second argument hits bottom when the first ends, there will usually be a bottom in the last element of the zipped list—weird.

----

Obviously, there's no *objective* way to choose between one and another, but the relationship between `NE` and `foldrMap1` is a particularly pretty thing.

containers/src/Data/Graph.hs
meooow25: After some thought, I feel it's best to let this lie. It is very unlikely someone is using `null` on an `SCC`, so this is unlikely to help someone with performance. But in case they are, this is an unnecessary breaking change to deal with.  
Let me know if you feel otherwise.

meooow25:
I see, certainly this type is better suited for some situations, though I'm not convinced that it should be the default.  
Anyway, we're stuck with the NonEmpty we have today.

Issue #1061, opened 2024-11-02
updateMax Map/IntMap discrepancy


meooow25:
`updateMax`, `updateMaxWithKey`, `updateMin`, `updateMinWithKey` for lazy and strict IntMap do not agree with Map.

IntMap errors on an empty map:

https://github.com/haskell/containers/blob/7e7ce15704dce68f67d5e0813ae442993e55b128/containers/src/Data/IntMap/Internal.hs#L2180-L2189

Map simply returns the empty map:

https://github.com/haskell/containers/blob/7e7ce15704dce68f67d5e0813ae442993e55b128/containers/src/Data/Map/Internal.hs#L1744-L1749

The better option is that IntMap also returns an empty map.

---

Historical note: The Map behavior has been so since it was added in bbbba97cbcf12039810533e3a2daf2eefdefe7f0. `updateMax` and friends for IntMap were added a bit later in 1a2cf0f6d9cfddb272af43653915094ef519caa8.

meooow25:
Discovered when writing IntMap strictness tests for #1019

PR #1063, opened 2024-11-03
Add strictness tests for IntMap construction

meooow25:
For #1019.
Follows #1021.


meooow25:
Failing on #1061 (the test checks for the desired non-erroring behavior).

PR #1065, opened 2024-11-08
returning Nil on empty IntMap as done in corresponding functions for Map

### LINKED ISSUE #1061: updateMax Map/IntMap discrepancy

kushagarr:
Closes #1061

@meooow25

### DIFF
--- a/containers-tests/tests/intmap-properties.hs
+++ b/containers-tests/tests/intmap-properties.hs
@@ -1079,6 +1079,9 @@ test_updateMin = do
     updateMin (\ a -> Just ("X" ++ a)) (fromList [(5,"a"), (-3,"b")]) @?= fromList [(-3, "Xb"), (5, "a")]
     updateMin (\ _ -> Nothing)         (fromList [(5,"a"), (-3,"b")]) @?= singleton 5 "a"
 
+    updateMin (\ a -> Just ("X" ++ a)) (empty :: SMap) @?= empty
+    updateMin (\ _ -> Nothing)         (empty :: SMap) @?= empty
+
 test_updateMax :: Assertion
 test_updateMax = do
     updateMax (\ a -> Just ("X" ++ a)) (fromList [(5,"a"), (3,"b")]) @?= fromList [(3, "b"), (5, "Xa")]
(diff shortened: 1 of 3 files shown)


treeowl (changes requested):
Please adjust the property tests accordingly.

meooow25:
Thanks for the PR. Your change seems fine, but as mentioned above please update/add property tests for affected functions. These will be in `containers-tests/intmap-properties.hs`.

treeowl:
@meooow25 , it's obviously not an issue for this PR, but it looks like we only have unit tests for these and many other `IntMap` functions, and no property tests. That's no good!

kushagarr:
So, what is it that I should be doing here, 
If I take `updateMinWithKey` as an example, I see only one `Assertion` called `test_updateMinWithKey` in `containers-tests/tests/intmap-properties.hs`.

Should I just update that assertion to have this change also included in that or you want me to add property tests for this as well.
FYI, Although I have read about tests and property tests in Haskell, I haven't really written them in Haskell before, So please bear with me and guide me if I make mistakes.

meooow25:
I took at look at the tests we have. You could add a new unit test for the empty map case in `test_updateMinWithKey`, `test_updateMaxWithKey`, etc. That's good enough for this PR.
Otherwise, if you want to add new property tests for these functions, it would be very helpful. Let us know if you need any help.

> ...but it looks like we only have unit tests for these and many other `IntMap` functions, and no property tests. That's no good!

Yes that's not ideal, we should look into having property tests for everything.

treeowl:
> Not able to come up with property tests yet, apologies. I need more time may be.

Don't worry about that. @meooow25 opened a new issue to deal with that.

treeowl (approved):

meooow25:
Thanks!

PR #1076, opened 2024-11-30
Implement Seq folds as coerced FingerTree folds

meooow25:
* foldr1 and foldl1 are now implemented as coercions.
* Other folds were implemented using coercions in a complicated manner, which is now simplified. GHC is smart enough to compile these to the same Core, so there is no change in runtime behavior.


containers/src/Data/Sequence/Internal.hs
```diff
@@ -395,33 +394,45 @@ fmapSeq f (Seq xs) = Seq (fmap (fmap f) xs)
...
 #endif
 
-getSeq :: Seq a -> FingerTree (Elem a)
-getSeq (Seq xs) = xs
-
 instance Foldable Seq where
-    foldMap f = foldMap (f .# getElem) .# getSeq
-    foldr f z = foldr (f .# getElem) z .# getSeq
-    foldl f z = foldl (f .^# getElem) z .# getSeq
+#ifdef __GLASGOW_HASKELL__
```
treeowl, line 398: We need `foldMap` for other implementations too.

meooow25: What do you mean? We have `foldMap` implementations.

treeowl: Am I missing something? It looks like you just hid it behind an `ifdef`.

meooow25: Oh, you mean _Haskell_ implementations, I misunderstood. Those definitions are in the `#else` part of this `#ifdef`.

meooow25: Does it seem fine, @treeowl?

treeowl:
Looks good. Thanks.

PR #1077, opened 2024-11-30
Make Seq foldl', foldr' strict in the initial value

### LINKED ISSUE #1072: foldl' and foldr' for Seq are not strict in the starting value
This is unlike every other structure we have.

```
λ> import qualified Data.Foldable as F
λ> import qualified Data.Set as S -- or Map or IntSet or IntMap
λ> F.foldl' (\_ _ -> ()) undefined (S.singleton ())
*** Exception: Prelude.undefined
CallStack (from HasCallStack):
  ...
λ> F.foldr' (\_ _ -> ()) undefined (S.singleton ())
*** Exception: Prelude.undefined
  ...
λ> import qualified Data.Sequence as Seq
λ> F.foldl' (\_ _ -> ()) undefined (Seq.singleton ())
()
λ> F.foldr' (\_ _ -> ()) undefined (Seq.singleton ())
()
```

I think we should make these strict.

Perhaps it's worth mentioning that for lists, `foldl'` is strict in the starting value but `foldr'` is not ([GHC #25508](https://gitlab.haskell.org/ghc/ghc/-/issues/25508)).

meooow25:
This matches the behavior of strict folds on sets and maps.

Closes #1072.

### DIFF
--- a/containers-tests/containers-tests.cabal
+++ b/containers-tests/containers-tests.cabal
@@ -395,11 +395,16 @@ test-suite seq-properties
   hs-source-dirs:   tests
   main-is:          seq-properties.hs
   type:             exitcode-stdio-1.0
+  build-depends:
+      ChasingBottoms
 
   ghc-options:      -O2
   other-extensions:
     BangPatterns
     CPP
+
+  other-modules:
+      Utils.Strictness
 
 test-suite tree-properties
   import: test-deps, warnings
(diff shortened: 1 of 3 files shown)


treeowl:
I just merged this myself to clear it out of the way so I can play with GADTy folds. However, we need a changelog entry!

meooow25:
Yep!

Looking at it again, some of those `$!` are unnecessary now, but they shouldn't hurt. I'll check once with the benchmarks.

PR #1079, opened 2024-12-19
Improve folds over bits for IntSet

### LINKED ISSUE #666: IntSet foldl and foldr smell funny

meooow25:
* Make foldr and foldl short-circuit instead of lazily accumulating thunks.
* Switch to a non-empty style to avoid unnecessary comparisons. This also helps GHC with arity-analysis (somehow), which greatly improves performance of CPS-style foldr and foldl.
* Change the bitwise operations used from ``bitmask = bm .&. -bm; bi = ctz bitmask; bm' = bm `xor` bitmask`` to `bi = ctz bm; bm' = bm .&. (bm-1)` which is slightly faster.

---

Benchmarks with GHC 9.10.1

```
Name                          Time - - - - - - - -    Allocated - - - - -
dense.foldMap_elem            5.6 μs  5.3 μs   -4%      0 B     0 B
dense.foldMap_traverseSum     6.5 μs  4.5 μs  -31%    1.0 KB  1.0 KB   +0%
dense.foldl'_sum              5.6 μs  4.3 μs  -22%      0 B     0 B
dense.foldl_cpsOneShotSum      19 μs  5.7 μs  -70%    194 KB  2.5 KB  -98%
dense.foldl_cpsSum             20 μs  6.2 μs  -69%    162 KB  5.0 KB  -96%
dense.foldl_elem               12 μs  5.7 μs  -53%    162 KB  2.0 KB  -98%
dense.foldl_traverseSum        20 μs  5.9 μs  -70%    162 KB  5.0 KB  -96%
dense.foldr'_sum              5.8 μs  5.5 μs   -6%      0 B     0 B
dense.foldr_cpsOneShotSum      20 μs  5.6 μs  -72%    194 KB  2.5 KB  -98%
dense.foldr_cpsSum             21 μs  5.9 μs  -71%    162 KB  5.0 KB  -96%
dense.foldr_elem               12 μs  4.8 μs  -60%    162 KB  2.0 KB  -98%
dense.foldr_traverseSum        21 μs  4.9 μs  -76%    162 KB  5.0 KB  -96%
sparse.foldMap_elem            19 μs   18 μs   -5%      0 B     0 B
sparse.foldMap_traverseSum     21 μs   20 μs   -5%     64 KB   64 KB   +0%
sparse.foldl'_sum              16 μs   14 μs  -14%      0 B     0 B
sparse.foldl_cpsOneShotSum     58 μs   38 μs  -34%    319 KB  160 KB  -49%
sparse.foldl_cpsSum            58 μs   60 μs   +4%    287 KB  320 KB  +11%
sparse.foldl_elem              27 μs   29 μs   +6%    287 KB  128 KB  -55%
sparse.foldl_traverseSum       58 μs   55 μs   -5%    287 KB  320 KB  +11%
sparse.foldr'_sum              27 μs   24 μs  -10%      0 B     0 B
sparse.foldr_cpsOneShotSum     66 μs   29 μs  -55%    319 KB  160 KB  -49%
sparse.foldr_cpsSum            65 μs   50 μs  -22%    287 KB  320 KB  +11%
sparse.foldr_elem              37 μs   20 μs  -47%    287 KB  128 KB  -55%
sparse.foldr_traverseSum       65 μs   44 μs  -32%    287 KB  320 KB  +11%
```

---

Fixes #666

### DIFF
--- a/containers/src/Data/IntSet/Internal.hs
+++ b/containers/src/Data/IntSet/Internal.hs
@@ -1724,9 +1724,10 @@ foldlBits prefix f z0 bitmap = go z0 $! revWord bitmap
   where
     -- Note: We pass the z as a static argument because it helps GHC with demand
     -- analysis. See GHC #25578 for details.
-    go z !bm = f (if bm' == 0 then z else go z bm') (prefix .|. bi)
+    go z !bm = f (if bm' == 0 then z else go z bm') x
       where
         bi = WORD_SIZE_IN_BITS - 1 - countTrailingZeros bm
+        !x = prefix .|. bi
         bm' = bm .&. (bm-1)
 
 foldl'Bits prefix f z0 bitmap = go z0 bitmap
@@ -1734,16 +1735,18 @@ foldl'Bits prefix f z0 bitmap = go z0 bitmap
     go !z !bm = if bm' == 0 then z' else go z' bm'
       where
         bi = countTrailingZeros bm
-        !z' = f z (prefix .|. bi)
+        !x = prefix .|. bi
+        !z' = f z x
         bm' = bm .&. (bm-1)
 
 foldrBits prefix f z0 bitmap = go bitmap z0
   where
     -- Note: We pass the z as a static argument because it helps GHC with demand
     -- analysis. See GHC #25578 for details.
-    go !bm z = f (prefix .|. bi) (if bm' == 0 then z else go bm' z)
+    go !bm z = f x (if bm' == 0 then z else go bm' z)
       where
         bi = countTrailingZeros bm
+        !x = prefix .|. bi
         bm' = bm .&. (bm-1)
 
 foldr'Bits prefix f z0 bitmap = (go $! revWord bitmap) z0
@@ -1751,7 +1754,8 @@ foldr'Bits prefix f z0 bitmap = (go $! revWord bitmap) z0
     go !bm !z = if bm' == 0 then z' else go bm' z'
       where
         bi = WORD_SIZE_IN_BITS - 1 - countTrailingZeros bm
-        !z' = f (prefix .|. bi) z
+        !x = prefix .|. bi
+        !z' = f x z
         bm' = bm .&. (bm-1)
 
 foldMapBits prefix f bitmap = go bitmap
@@ -1765,7 +1769,7 @@ foldMapBits prefix f bitmap = go bitmap
 #endif
       where
         bi = countTrailingZeros bm
-        x = prefix .|. bi
+        !x = prefix .|. bi
         bm' = bm .&. (bm-1)
 
 takeWhileAntitoneBits prefix predicate bitmap =
--- a/containers/src/Data/IntSet/Internal.hs
+++ b/containers/src/Data/IntSet/Internal.hs
@@ -1699,10 +1699,6 @@ takeWhileAntitoneBits :: Int -> (Int -> Bool) -> Word -> Word
 
 #if defined(__GLASGOW_HASKELL__)
 
-lowestBitMask :: Word -> Word
-lowestBitMask x = x .&. negate x
-{-# INLINE lowestBitMask #-}
-
 lowestBitSet x = countTrailingZeros x
 
 highestBitSet x = WORD_SIZE_IN_BITS - 1 - countLeadingZeros x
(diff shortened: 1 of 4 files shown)


meooow25:
Related GHC issue for a curious behavior I observed: [GHC#25578](https://gitlab.haskell.org/ghc/ghc/-/issues/25578) (also mentioned in the code).

meooow25:
Okay, I'm pretty happy with this. Can't think of anything else that might be improved here.

containers/src/Data/IntSet/Internal.hs
```diff
@@ -1724,45 +1720,53 @@ revWord x1 = case ((x1 `shiftRL` 1) .&. 0x5555555555555555) .|. ((x1 .&. 0x55555
...
-        go bm !acc = go (bm `xor` bitmask) ((f $! (prefix+(WORD_SIZE_IN_BITS-1)-bi)) acc)
-          where !bitmask = lowestBitMask bm
-                !bi = countTrailingZeros bitmask
-
-foldMapBits prefix f bitmap = go (prefix + bi0) (bitmap `xor` bitmask0)
+foldlBits prefix f z0 bitmap = go z0 $! revWord bitmap
+  where
+    -- Note: We pass the z as a static argument because it helps GHC with demand
+    -- analysis. See GHC #25578 for details.
+    go z !bm = f (if bm' == 0 then z else go z bm') (prefix .|. bi)
```
treeowl, line 1727: Does GHC calculate the second argument eagerly when `f` is unknown? Should it?

meooow25: > Does GHC calculate the second argument eagerly when `f` is unknown?

GHC decides to do that here, yes. The STG looks like

```
let {
  Rec {
  $wgo8_s4BNB [InlPrag=[2], Occ=LoopBreaker, Dmd=SC(S,C(1,L))]
    :: a_a4wJ1 -> Word# -> a_a4wJ1
  [LclId, Arity=2, Str=<ML><L>, Unf=OtherCon []] =
      {$wgo8_s4BNB, f_s4BNo, bx_s4BNz} \r [z1_s4BNC ww_s4BND]
          case ctz# [ww_s4BND] of sat_s4BNH {
          __DEFAULT ->
          case word2Int# [sat_s4BNH] of sat_s4BNI {
          __DEFAULT ->
          case -# [63# sat_s4BNI] of sat_s4BNJ {
          __DEFAULT ->
          case orI# [bx_s4BNz sat_s4BNJ] of sat_s4BNK {
          __DEFAULT ->
          let {
            sat_s4BNL [Occ=Once1] :: Int
            [LclId] =
                I#! [sat_s4BNK]; } in
          let {
            sat_s4BNG [Occ=Once1] :: a_a4wJ1
            [LclId] =
                {z1_s4BNC, $wgo8_s4BNB, ww_s4BND} \u []
                    case minusWord# [ww_s4BND 1##] of sat_s4BNE {
                    __DEFAULT ->
                    case and# [ww_s4BND sat_s4BNE] of wild1_s4BNF {
                      __DEFAULT -> $wgo8_s4BNB z1_s4BNC wild1_s4BNF;
                      0## -> z1_s4BNC;
                    };
                    };
          } in  f_s4BNo sat_s4BNG sat_s4BNL;
          };
          };
          };
          };
  end Rec }
} in
...
```

> Should it?

I don't see any issue with leaving this up to GHC. Do you?  
Compared the forcing it always, the (perhaps small) benefit is that GHC can optimize for some known `f` to drop this calculation if it isn't used.

treeowl: The only time that could happen would be if the result only depends on the size of the set. I believe the general case can be handled by a function something like this:

```haskell
sc :: IntSet -> (Int -> StateT s (Either r) ()) -> s -> r
```

The function argument should have the property that `f m *> f n = f (m + n)` when `m` and `n` are positive.

meooow25: > The only time that could happen would be if the result only depends on the size of the set.

No, why would that be? Let's say we want to sum every 10th element in the set.

```hs
sum10 :: IntSet -> Int
sum10 s = foldr f z s (0 :: Int) 0
  where
    f x k !i !acc = k (if i == 9 then 0 else i+1) (if i == 0 then acc+x else acc)
    z !_ acc = acc
```

This doesn't care about 90% of the elements, and I expect GHC to optimize this to skip all the unnecessary `prefix .|. ctz bm`.

meooow25: I checked the Core for `sum10` and it turned to be identical when forcing and not forcing the element. GHC did the reasonable thing and moved the `prefix .|. ctz bm` to the branch where `i == 0`.

> Should it?

So the answer to this seems to me to be "it doesn't matter". But if you would prefer to force the element I can make the change.

treeowl: I think it's probably better to force it, so no one else needs to wonder. But I don't feel super strongly about it.

PR #1081, opened 2024-12-28
MicroHs has pattern synonyms now

augustss:
What the title says.

Also, update CI to use mcabal for building the package.


meooow25 (approved):

meooow25:
A question: We have dependencies on `array` and `deepseq` for some imports, which MicroHs instead provides from `base`. How does that workout with mcabal?

augustss:
Mcabal has a hack for a few packages that are in base instead of separate.

meooow25:
Thanks!

PR #1086, opened 2025-01-04
Improve compare for IntSet and IntMap

### LINKED ISSUE #470: Improve Ord instances for IntSet and IntMap
The current `Ord` instances for `IntSet` and `IntMap` don't take advantage of the way these types are structured at all. We may therefore be leaving some performance on the table. In particular:

1. I don't think we should need to convert the maps to lists for comparison.
2. We can recognize certain special arrangements of keys, prefixes, and perhaps masks to avoid following a bunch of pointers for nothing. For example, if one map has a negative prefix and the other has a (strictly) positive prefix, we are *done*.

### LINKED ISSUE #787: Improve Ord IntSet instance
An attempt was made in #670, but it proved to have some subtle bug in #783 and had to be reverted. It would be nice to fix and reinstate.

meooow25:
Compare the trees directly instead of converting to lists.
The implementation follows broadly the same approach as the previous attempt in commit 7aff529.

Closes #470, closes #787.

Benchmarks with GHC 9.10:

Map
```
Name       Time - - - - - - - -    Allocated - - - - -
                A       B     %         A      B     %
compare    167 μs   32 μs  -80%    767 KB    0 B   -100%
```

Set
```
Name              Time - - - - - - - -    Allocated - - - - -
                       A       B     %         A      B     %
compare:dense      55 μs  412 ns  -99%    640 KB   12 B   -99%
compare:sparse     87 μs   26 μs  -70%    893 KB    0 B   -100%
```

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -308,6 +308,7 @@ import Data.IntSet.Internal.IntTreeCommons
   , TreeTreeBranch(..)
   , treeTreeBranch
   , i2w
+  , Order(..)
   )
 import Utils.Containers.Internal.BitUtil (shiftLL, shiftRL, iShiftRL)
 import Utils.Containers.Internal.StrictPair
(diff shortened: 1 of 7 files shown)


containers/src/Data/IntMap/Internal.hs
```diff
@@ -3473,12 +3474,74 @@ instance Eq1 IntMap where
...
-  liftCompare cmp m n =
-    liftCompare (liftCompare cmp) (toList m) (toList n)
+  liftCompare = liftCmp
+
+liftCmp :: (a -> b -> Ordering) -> IntMap a -> IntMap b -> Ordering
+liftCmp cmp m1 m2 = case (splitSign m1, splitSign m2) of
+  ((l1, r1), (l2, r2)) -> case go l1 l2 of
+    Less -> LT
+    Prefix' -> if null r1 then LT else GT
+    Equals -> case go r1 r2 of
```
treeowl, line 3489: Consider moving this case to the bottom for readability.

treeowl (commented):
I'm having trouble understanding why we need a fancy result type for the `go` function. What's going on here?

meooow25:
For a lexicographical comparison of the elements it's not enough to use `Ordering`. Suppose we have
```
t1 = [0]   [2]
t2 = [0,1] [2]
```
then comparing the left branches (`compare [0] [0,1]`) would return `LT`, but the overall result should be `GT`.  
Detecting whether one set is a prefix of the other lets us handle this correctly.

containers/src/Data/IntMap/Internal.hs
meooow25: Do you think that would be better? I prefer the consistent order of branches.

treeowl:
> For a lexicographical comparison of the elements it's not enough to use `Ordering`. Suppose we have
> [...]
Oh right ... That makes sense.

containers/src/Data/IntMap/Internal.hs
treeowl: I don't feel super strongly about it.

meooow25:
The type is introduced in this PR. The constructor names are from the previous attempt (7aff529), I don't mind changing them if you have suggestions.

meooow25:
What names would you prefer @treeowl?

meooow25:
Okay, this is good to go.  
@treeowl would you still like me to change the names?

containers/src/Data/IntSet/Internal.hs
```diff
@@ -1477,8 +1478,90 @@ equal _   _   = False
...
+  EQ | bm1 == bm2 -> Equals
+     | otherwise ->
+         let diff = bm1 `xor` bm2
+             lowestDiff = diff .&. negate diff
+             highMask = negate lowestDiff
+         in if bm1 .&. lowestDiff == 0
+            then (if bm1 .&. highMask == 0 then Prefix' else Greater)
+            else (if bm2 .&. highMask == 0 then FlipPrefix else Less)
+  GT -> Greater
+{-# INLINE orderTips #-}
```
treeowl, line 1552: Please document `leftmostTipSure` and `orderTips`. `orderTips could also use some internal comments to explain what's going on with the bit fiddling.

containers/src/Data/IntSet/Internal/IntTreeCommons.hs
```diff
@@ -161,6 +162,14 @@ i2w :: Int -> Word
...
 {-# INLINE i2w #-}
 
+-- Used to compare IntSets and IntMaps
+data Order
+  = Less       -- holds for [0,3,4] [0,3,5,1]
+  | Prefix'    -- holds for [0,3,4] [0,3,4,5]
+  | Equals     -- holds for [0,3,4] [0,3,4]
+  | FlipPrefix -- holds for [0,3,4] [0,3]
+  | Greater    -- holds for [0,3,4] [0,2,5]
+
```
treeowl, line 172: The naming is ... not trivial. What do you think about something like this?

```haskell
data Order
  = A_LT_B
  | A_Prefix_B
  | A_EQ_B
  | B_Prefix_A
  | A_GT_B
```

meooow25: Seems fine, changed.

containers/src/Data/IntSet/Internal.hs
meooow25: Left a comment in `orderTips`.  
`leftmostTipSure` returns ... the leftmost Tip. I don't know what else to say about that 😄

containers/src/Data/IntSet/Internal.hs
```diff
@@ -1486,8 +1487,94 @@ equal _   _   = False
...
+    go Nil _ = A_Prefix_B
+    go _ Nil = B_Prefix_A
+
+-- This type allows GHC to return unboxed ints from leftmostTipSure, as
+-- $wleftmostTipSure :: IntSet -> (# Int#, Word# #)
+-- On a modern enough GHC (>=9.4) this is unnecessary, we could use StrictPair
+-- instead and get the same Core.
+data Tip' = Tip' {-# UNPACK #-} !Int {-# UNPACK #-} !BitMap
+
+leftmostTipSure :: IntSet -> Tip'
```
treeowl, line 1544: Please add a Haddock string.

containers/src/Data/IntSet/Internal.hs
```diff
@@ -1486,8 +1487,94 @@ equal _   _   = False
...
+leftmostTipSure (Bin _ l _) = leftmostTipSure l
+leftmostTipSure (Tip k bm) = Tip' k bm
+leftmostTipSure Nil = error "leftmostTipSure: Nil"
+
+orderTips :: Int -> BitMap -> Int -> BitMap -> Order
```
treeowl, line 1549: Please add a haddock string.

treeowl (commented):
I think that's much better, but a bit more documentation is in order.

containers/src/Data/IntSet/Internal.hs
meooow25: Why? These functions are not exposed.

meooow25: @treeowl do you still want to add doc strings?

treeowl: I'll try to write one for `compareTips` by tomorrow night. But otherwise I guess you can merge and I'll open an issue to remember.

meooow25: I'll go ahead and merge it then.

PR #1091, opened 2025-01-12
Fix the ReadTheDocs config

### LINKED ISSUE #1090: ReadTheDocs builds are failing
Failing since Jan 14, 2024. See https://readthedocs.org/projects/haskell-containers/builds/

Looks like https://github.com/readthedocs/readthedocs.org/issues/11173

(I noticed the problem when I checked if #1088 was live)

meooow25:
Fixes #1090


meooow25:
This _should_ work, but I can't find a quick way to confirm that, so I guess we'll just find out.

PR #1094, opened 2025-01-17
Skip rebalancing in glue

meooow25:
If two non-empty trees are balanced with respect to each other, removing one element from the larger does not take away that property.


treeowl (approved):
I have no idea how this wasn't caught before, though I doubt it affects performance much. Do the benchmarks show anything?

meooow25:
Yes it's too small a change to make a noticeable difference in our benchmarks.

If I construct a (very artificial) benchmark like

```hs
s = Set.fromList [1,2,3]
...
bench "delete123" $ whnf (S.delete 2) s123
```

the change is ~15.8ns to ~9.7ns.

treeowl:
Good catch regardless.

PR #1095, opened 2025-01-18
Remove unnecessary return definitions

meooow25:
The default return=pure works and we are long past the Functor-Applicative-Monad Proposal.


meooow25:
I was about to remove Seq's `(>>) = (*>)` but realized that the default is unfortunately not that but `m >> k = m >>= \_ -> k`. \*sigh\*

treeowl:
> I was about to remove Seq's `(>>) = (*>)` but realized that the default is unfortunately not that but `m >> k = m >>= \_ -> k`. *sigh*

It's not at all unusual to have to write `(*>) = (>>)` either. Thinking is unfortunately required for these.

treeowl (approved):

PR #1097, opened 2025-01-24
Modernize the cabal file

meooow25:
* Set cabal-version to 2.2
* Set license to the right SPDX identifier
* Drop old -fwarn flags which are included in -Wall (since GHC 9.2)
* Move changelog.md to extra-doc-files
* Add homepage
* Tweak source repo location

### DIFF
--- a/containers-tests/containers-tests.cabal
+++ b/containers-tests/containers-tests.cabal
@@ -31,7 +31,7 @@ tested-with:
 
 source-repository head
   type:     git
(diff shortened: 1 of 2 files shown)


treeowl:
The only thing I question is moving the changelog. Why are we doing that? Don't people expect to find it at the top?

treeowl:
Never mind! I understand now.

treeowl (approved):

PR #1098, opened 2025-01-25
Drop "amortized" in docs where it doesn't apply

meooow25:


treeowl (approved):
Dunno how those ever snuck in.

PR #1099, opened 2025-01-26
Fix the ReadTheDocs setup

### LINKED ISSUE #1090: ReadTheDocs builds are failing

meooow25:
* Fix Sphinx and sphinx-rtd-theme versions and update the code to make it work.
* Use sphinx-rtd-theme instead of alabaster. This makes the appearance consistent with GHC and Cabal RTD.

Fixes #1090


meooow25:
Works! https://readthedocs.org/projects/haskell-containers/builds/26971807/

PR #1106, opened 2025-02-15
some fixes to the short example

ninioArtillero:


meooow25 (approved):
A bit unfortunate that `index` has an unusual order, but it's just always been that way.

ninioArtillero:
We could say the same about society, but in that case we don't really know how long it's been like that. Anyway, backwards compatibility is very much like honoring the ancestors.

meooow25:
Thanks for the fix!

ninioArtillero:
My pleasure and honor!

PR #1108, opened 2025-02-16
unified approach for different cases

qxrein:
- Better buildQ implementation to handle all cases consistently
- Added proper handling of the case where all trailing zeros are in the lower word
- Improved the calculation of trailing zeros by considering both words together
- Added proper bounds checking for the shift operations

### DIFF
--- a/containers/src/Utils/Containers/Internal/BitQueue.hs
+++ b/containers/src/Utils/Containers/Internal/BitQueue.hs
@@ -86,15 +86,17 @@ snocQB bq b = case shiftQBR1 bq of
 -- off.
 {-# INLINE buildQ #-}
 buildQ :: BitQueueB -> BitQueue
-buildQ (BQB hi 0) = BQ (BQB 0 lo') where
-  zeros = countTrailingZeros hi
-  lo' = ((hi `shiftRL` 1) .|. (1 `shiftLL` (wordSize - 1))) `shiftRL` zeros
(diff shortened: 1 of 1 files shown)


treeowl:
This type exists for speed. Things like proper bounds checking aren't entirely necessary. How does this change affect performance?

qxrein:
I am really sorry for recommending changes that would have compromised performance in a module explicitly designed for speed. You're absolutely right - this implementation relies on pattern matching and direct code paths that the compiler can optimize effectively.
I am closing this PR and again sorry for this, thanks for consideration.

meooow25:
No need to apologize, treeowl only asked you about the effect on performance. It wasn't a rejection.
The motivation for this change is also unclear to me. You're making some changes but you haven't explained why.

Feel free to reopen if you wish to continue.

qxrein:
>No need to apologize, treeowl only asked you about the effect on performance. It wasn't a rejection.

I am new to contributing to the Haskell organization, and really look forward to expertise within this community. I value your input as I’m still learning, and I appreciate the guidance as you guys are more experienced in this community. About the PR I opened, @treeowl was correct, after rechecking I found that this change could have negatively impacted performance. I’ll make sure to take that into account moving forward.

>The motivation for this change is also unclear to me. You're making some changes but you haven't explained why.

I was trying to make it look more readable that's it. Again, sorry for that - I'll research more on this thing and reopen this PR if I find better solution, thanks for this motivation.

PR #1109, opened 2025-02-18
Some additions to Data.Tree

### LINKED ISSUE #1107: Additions to Data.Tree
I propose adding some definitions to the rather minimal `Data.Tree` module.

### `para`

```hs
-- | A paramorphism for @Tree@.
para :: (Tree a -> [b] -> b) -> Tree a -> b
```

Generally useful function, like `foldTree`.  
I would love to have a better name for it though. `subtreeFold`? `foldSubtrees`?

### `leaves`

```hs
-- | The leaves of a tree in left-to-right order.
leaves :: Tree a -> [a]
```

Elsewhere:
* [`Debian.Apt.Dependencies.leaves`](https://hackage.haskell.org/package/debian-4.0.5/docs/Debian-Apt-Dependencies.html#v:leaves) (accidentally $O(n^2)$ )
* [`Math.TreeFun.Tree.leaves`](https://hackage.haskell.org/package/tree-fun/docs/Math-TreeFun-Tree.html#v:leaves) (accidentally $O(n^2)$ )

### `edges`

```hs
-- | The edges of a tree as parent-child pairs in pre-order.
edges :: Tree a -> [(a, a)]
```

Elsewhere:
* [`GHC.CmmToAsm.CFG.Dominators.parents`](https://hackage.haskell.org/package/ghc/docs/GHC-CmmToAsm-CFG-Dominators.html#v:parents) (swapped version, accidentally $O(n^2)$ )
* [`Topograph.pairs`](https://hackage.haskell.org/package/topograph/docs/Topograph.html#v:treePairs) (accidentally $O(n^2)$ )

### `pathsToRoot`

```hs
-- | Labels on the paths from each node to the root.
pathsToRoot :: Tree a -> Tree (NonEmpty a)
```

Elsewhere:

* [`GHC.CmmToAsm.CFG.Dominators.ancestors`](https://hackage.haskell.org/package/ghc/docs/GHC-CmmToAsm-CFG-Dominators.html#v:ancestors) (a variation, accidentally $O(n^2)$ )

### `pathsFromRoot`

```hs
-- | Labels on the paths from the root to each node.
pathsFromRoot :: Tree a -> Tree (NonEmpty a)
-- This should use a queue to make the lists generate elements efficiently. A banker's queue like in `Data.List.inits` should do it.
```

Elsewhere:

* A variation proposed in #140 as `steps`, with some unresolved issues.

### `PostOrder`

```hs
-- | A newtype that folds and traverses the tree in post-order.
newtype PostOrder a = PostOrder { unPostOrder :: Tree a }

instance Foldable PostOrder where
   ...
instance Foldable1 PostOrder where
   ...
instance Traversable PostOrder where
   ...
```

Elsewhere:

* [Data.Graph.Inductive.Basic.postorder](https://hackage.haskell.org/package/fgl/docs/Data-Graph-Inductive-Basic.html#v:postorder) (accidentally $O(n^2)$ )

I think these will be useful, saving people from redefining common functions repeatedly and often inefficiently.

---

Previously

* #39 (our oldest open issue!) proposed `treefilter`, `subtreeat`, `subtreeinforest`. I'm not convinced on the utility of `treefilter`. The name is also a bit misleading because it does not seem like a well-behaved "filter"—it removes entire subtrees based on a root's result. Also not convinced on `subtreeat`, but I could change my mind on this. It is also easily defined using `para`: `subtreeat x = para $ \t@(Node y ts) ms -> if x == y then Just t else asum ms`.
* #140 proposed `paths` and `steps`. `paths = toList . leaves . pathsFromRoot`, `steps = toList . pathsFromRoot`.

meooow25:
Add some useful definitions to Data.Tree: leaves, edges, pathsToRoot, pathsFromRoot, PostOrder.

Closes #1107.

### DIFF
--- a/containers-tests/tests/tree-properties.hs
+++ b/containers-tests/tests/tree-properties.hs
@@ -28,7 +28,6 @@ main :: IO ()
 main = defaultMain $ testGroup "tree-properties"
          [
            testCase     "foldr"                    test_foldr
-         , testCase "PostOrder_foldr"              test_PostOrder_foldr
          , testProperty "monad_id1"                prop_monad_id1
          , testProperty "monad_id2"                prop_monad_id2
          , testProperty "monad_assoc"              prop_monad_assoc
@@ -61,7 +60,8 @@ main = defaultMain $ testGroup "tree-properties"
          , testProperty "pathsToRoot"              prop_pathsToRoot
          , testProperty "pathsFromRoot"            prop_pathsFromRoot
          , testGroup "PostOrder"
-           [ testProperty "toList"                 prop_PostOrder_toList
+           [ testCase "foldr"                      test_PostOrder_foldr
+           , testProperty "toList"                 prop_PostOrder_toList
            , testProperty "foldMap"                prop_PostOrder_foldMap
            , testProperty "foldMap_structure"      prop_PostOrder_foldMap_structure
            , testProperty "foldl'"                 prop_PostOrder_foldl'


containers/src/Data/Tree.hs
```diff
@@ -532,6 +543,243 @@ unfoldForestQ f aQ = case viewl aQ of
...
+-- A tree with \(n\) nodes has \(n-1\) edges.
+--
+-- ==== __Examples__
+--
+-- >>> edges (Node 1 [Node 2 [Node 4 [], Node 5 []], Node 3 [Node 6 []]])
+-- [(1,2),(2,4),(2,5),(1,3),(3,6)]
+-- >>> edges (Node "root" [])
+-- []
+--
+-- @since FIXME
```
treeowl, line 583: I think it would be useful to indent the documentation to show tree structure.

meooow25: That would be nice, but it looks like Haddock doesn't support multi-line REPL examples. So we would have to give that up and use an ordinary code block. Would you still prefer that?

meooow25: I realized we can just pretend that Haddock supports it 😄 
Doesn't look too bad.
screenshot-leaves

meooow25:
@treeowl, any other comments?

meooow25:
I'll merge it then.

PR #1110, opened 2025-02-20
Tighter bounds for IntMap and IntSet merge

### LINKED ISSUE #1011: Complexity of IntMap-IntMap operations
Recently, something made me notice the documented complexity of `IntMap`-`IntMap` operations, and I realized that I have never really thought about it.

Let's take `IntMap` union, for instance.

https://github.com/haskell/containers/blob/5d4bc2ed62475062a09d76a037ca7a72ad3cbd45/containers/src/Data/IntMap/Internal.hs#L1090

The claim is $O(n + m)$, which is not hard to verify. We match on some (in a bad case _all_) of the constructors of the two maps, of which there are a linear number, and whenever we do so we perform a constant amount of work. Hence $O(n + m)$.

But we can try to find a tighter bound.

Consider that $m \le n$, and we think of the smaller map. In the worst case, we might visit all constructors of the map, which makes $m$ a lower bound. The work we do in addition to this, is whenever we reach forks in the bigger map which are not in the smaller map. How many such forks can there be? The worst case is that for every leaf in the smaller map, the path to the root from a leaf hits every possible fork (i.e. forks at every power of 2). So, we need the number of nodes of the subset of the complete binary tree of all ints which forms paths from the root to the leaf keys of the map. For a map of size $m$ and the complete binary tree having height $W$, this is $O(m(W - \log m))$. Since we still have the $n + m$ upper bound as before, the overall complexity for $m \le n$ is

$$O(\min(n, m(W - \log m)))$$

I can note two interesting things about it:

1. If we consider $m$ as a small constant, the complexity is $O(\min(n,W))$, which checks out to be the complexity we have for `insert`.
2. The second term, $m (W - \log m) = m \log(\frac{2^W}{m})$, is not unlike the complexity for `Map` union, $O(m \log ( \frac{n}{m} ))$.

If this looks correct, I would like to update the documentation to reflect it.

Why does it matter? After all, the current bound is not incorrect.  
For accuracy, and also it would help better estimate complexity in derived scenarios. For instance, consider `foldl' union empty intmaps` on $r$ `IntMap`s of size $m$ each. With the current bound, we would calculate the complexity to be $O(r^2 m)$. With the new bound, we get $O(rm(W - \log m))$. We went from quadratic in $r$ to linear, which is a very useful distinction.

meooow25:
Closes #1011

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -740,7 +740,8 @@ unsafeFindMax (Bin _ _ r) = unsafeFindMax r
 {--------------------------------------------------------------------
   Disjoint
 --------------------------------------------------------------------}
--- | \(O(n+m)\). Check whether the key sets of two maps are disjoint
+-- | \(O(\min(n, m \log \frac{2^W}{m})), m \leq n\).
+-- Check whether the key sets of two maps are disjoint
 -- (i.e. their 'intersection' is empty).
 --
 -- > disjoint (fromList [(2,'a')]) (fromList [(1,()), (3,())])   == True
@@ -1091,7 +1092,8 @@ unionsWith :: Foldable f => (a->a->a) -> f (IntMap a) -> IntMap a
 unionsWith f ts
   = Foldable.foldl' (unionWith f) empty ts
 
--- | \(O(n+m)\). The (left-biased) union of two maps.
+-- | \(O(\min(n, m \log \frac{2^W}{m})), m \leq n\).
+-- The (left-biased) union of two maps.
 -- It prefers the first map when duplicate keys are encountered,
 -- i.e. (@'union' == 'unionWith' 'const'@).
 --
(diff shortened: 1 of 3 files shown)


meooow25:
It would be good to get some more eyes on this.

Lysxia (approved):
Your analysis looks good to me.

It may be worth explicitly defining $n$ as the min and $m$ as the max of the two sizes, otherwise at first glance it looks like there is something asymmetric going on.

meooow25:
> $O(\min(n, m \log \frac{2^W}{m})), m \leq n$

It does say $m \leq n$, in the same manner as Set and Map.

Though maybe you mean one may mistakenly think $m$ and $n$ are the first and second maps?

Lysxia:
Yes that's what I mean.

meooow25:
Hmm that goes for Set and Map too. And I'm not sure how to make it better.

$O(\min(\max(m,n), \min(m,n) \log \frac{2^W}{\min(m,n)}))$ looks terrible.

Lysxia:
How about a top level comment under the "Combine" header, with a slight change of naming:

> Most of these combining functions have complexity $O(\min(N, n \log(\frac{2^W}{n})))$, where $n$ is the minimum of the sizes of the arguments and $N$ is their maximum.

meooow25:
The "Combine" section might not be enough, since a few functions like `isSubmapOf` are elsewhere. But I like the idea, we can put this comment at the top of the module instead.

I'll make a separate PR for it (with Set and Map too).

PR #1115, opened 2025-02-24
Improve module Haddocks

meooow25:
* Make module Haddocks largely consistent across Set, Map, IntSet, IntMap, Seq.
* Make public modules for each structure (e.g. Data.Map, Data.Map.Lazy, Data.Map.Strict) have similar structure and content. This means some duplication of text, but there is no way to avoid it.
* Trim down internal module Haddocks (e.g. Data.Map.Internal) to a structure description and references. There is no need to repeat everything in the public module.
* Add a few lines about the complexity of common operations on each set and map structure. Notably, union and intersection are mentioned with a clear description of 'm' and 'n', to reduce the chances that a reader assumes them to be tied to positions and not sizes.

(The last bit is motivated by https://github.com/haskell/containers/pull/1110#issuecomment-2676750220)


Lysxia (approved):
This looks nice!

meooow25:
Thanks!

PR #1117, opened 2025-02-27
Remove unused TypeError util

meooow25:
Remove Utils.Containers.Internal.TypeError, which is currently unused.

We can bring it back in the future if we need to.


treeowl:
It was an interesting experiment, but it didn't work out so well in the end with compat policies.

PR #1139, opened 2025-04-30
Update compareSize implementations

meooow25:
The previous implementations went straight down the left spine, which does not match the documented complexity. We must decrement and potentially return at each step to meet O(min(n,c)).

Also add some benchmarks for size and compareSize.

For #826.

---

Benchmarks on GHC 9.10.1:

IntMap

```
Name             Time - - - - - - - -    Allocated - - - - -
compareSize:2     25 ns  6.0 ns  -76%     24 B    24 B    +0%
compareSize:n    7.4 μs  7.3 μs   -1%     36 B    35 B    -2%
```

IntSet

```
Name                    Time - - - - - - - -    Allocated - - - - -
compareSize:2            16 ns  6.4 ns  -60%     24 B    24 B    +0%
compareSize:n           255 ns  270 ns   +5%     32 B    32 B    +0%
compareSize:sparse:2     28 ns  6.4 ns  -76%     24 B    24 B    +0%
compareSize:sparse:n     17 μs   18 μs   +4%     35 B    40 B   +14%
```

### DIFF
--- a/containers-tests/benchmarks/IntMap.hs
+++ b/containers-tests/benchmarks/IntMap.hs
@@ -79,6 +79,9 @@ main = do
             whnf (uncurry M.withoutKeys) (m_random, s_random2)
         , bench "restrictKeys:random" $ -- large keys, no overlap
             whnf (uncurry M.restrictKeys) (m_random, s_random2)
+        , bench "size" $ whnf M.size m
+        , bench "compareSize:2" $ whnf (flip M.compareSize 2) m
+        , bench "compareSize:n" $ whnf (flip M.compareSize bound) m
         , bgroup "folds" $ foldBenchmarks M.foldr M.foldl M.foldr' M.foldl' foldMap m
         , bgroup "folds with key" $
             foldWithKeyBenchmarks M.foldrWithKey M.foldlWithKey M.foldrWithKey' M.foldlWithKey' M.foldMapWithKey m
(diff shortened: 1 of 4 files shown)


treeowl:
I didn't look at this earlier, but ... it seems like a bit of a mess. I can't make head or tail of what's going on. That outer comparison to 0 feels like a hack, and potentially an inefficient one at that. Going down a left branch is not the same as going down a right branch, and it kind of feels like maybe that gets lost? Unclear. I also can't tell if this code takes full advantage of the fact that `Nil` can't be a child of `Bin`.

meooow25:
As the author I might be biased, but I don't find it complicated 😅 
We just decrement `c` for each `Bin` we find. If it drops below 0, we are done. Happy to add some explanatory comments, but I'm not sure what I can write that is not already conveyed by the code.

> I also can't tell if this code takes full advantage of the fact that `Nil` can't be a child of `Bin`.

It does, in the branch `go _ c = c` it can only be a `Tip` and so it doesn't inspect the map.

treeowl:
Do you think you could add a few comments? And explain why we compare calculated values?

treeowl:
I think I see where I erred. I missed the fact that `compareSize` can't take quite as many shortcuts as a hypothetical `sizeAtLeast`. For example, `sizeAtLeast 2 (Bin p l r)` is always true, without needing to look at `p`, `l`, or `r`. However, `compareSize (Bin p l r) 2` can be either `EQ` or `GT`, depending on whether at least one of the children is `Bin`. I dunno if it's worth adding another function to get that shortcut.

treeowl:
However, I believe there may be a little room for improvement. In particular, I would try breaking the `go` function into `goL` and `goR`. `goR` can be tail recursive, which saves on popping up the stack comparing.

meooow25:
> However, I believe there may be a little room for improvement. In particular, I would try breaking the `go` function into `goL` and `goR`. `goR` can be tail recursive, which saves on popping up the stack comparing.

`go` is already tail recursive in the call to `go r`. I don't think I see the improvement you're referring to, would you mind writing it out?

PR #1143, opened 2025-05-12
Fix haddocks of Data.IntMap.compose

tomsmeding:
The words "former" and "latter" in the documentation of `compose` on `IntMap` were swapped. This PR puts them the right way round.

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -811,9 +811,8 @@ disjoint t1@(Bin p1 l1 r1) t2@(Bin p2 l2 r2) = case treeTreeBranch p1 p2 of
 {--------------------------------------------------------------------
   Compose
 --------------------------------------------------------------------}
--- | Relate the keys of one map to the values of
--- the other, by using the values of the former as keys for lookups
--- in the latter.
+-- | Given maps @bc@ and @ab@, relate the keys of @ab@ to the values of @bc@,
+-- by using the values of @ab@ as keys for lookups in @bc@.
 --
 -- Complexity: \( O(n * \min(m,W)) \), where \(m\) is the size of the first argument
 --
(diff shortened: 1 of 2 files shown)


Mikolaj (approved):
LGTM

alexfmpe:
Mind fixing this in `Data.Map.compose` as well?

meooow25:
The docs say

```hs
-- | Relate the keys of one map to the values of
-- the other, by using the values of the former as keys for lookups
-- in the latter.
...
compose :: IntMap c -> IntMap Int -> IntMap c
```

which looks correct to me:

```hs
-- | Relate the keys of one map (IntMap Int) to the values of
-- the other (IntMap c), by using the values of the former (IntMap Int) as keys for lookups
-- in the latter (IntMap c).
```

tomsmeding:
Ah, my bad. You are both correct:
1. The hadocks for `Data.Map` could to be improved too
2. The issue is not actually with former/latter, but with the fact that the "former" is the _second_ argument to `compose` and the "latter" is the _first_ argument to compose, which confused me, @Mikolaj and @alexfmpe -- @meooow25 was the first to notice.

I've force-pushed a new suggestion. Thanks for reading critically!

Mikolaj:
Aha! "Relate" as in "map", as opposed to as in "identify". I think this proves the text needs to be rephrased totally.

tomsmeding:
I think "relate" can be justified, because a Map (and an IntMap) is, after all, a [relation](https://en.wikipedia.org/wiki/Finitary_relation).

alexfmpe:
Maybe "link" ?

tomsmeding:
Perhaps:
> Given maps `bc` and `ab`, build a new map that relates keys of `ab` to values of `bc`, by using the values of `ab` as keys for lookups in `bc`.

This still uses "relates", but uses it in a way that makes it fairly unambiguous what it means, I think. @Mikolaj?

meooow25 (approved):
Thanks, this does seem like an improvement.

PR #1148, opened 2025-06-15
Improve mapping over keys for IntMap and IntSet 

### LINKED ISSUE #1027: mapKeys can be more efficient
https://github.com/haskell/containers/blob/734785d2d3ee862726c6c5bb30e6484fff370e50/containers/src/Data/Map/Internal.hs#L3208-L3209

https://github.com/haskell/containers/blob/734785d2d3ee862726c6c5bb30e6484fff370e50/containers/src/Data/Map/Internal.hs#L3227-L3228

https://github.com/haskell/containers/blob/734785d2d3ee862726c6c5bb30e6484fff370e50/containers/src/Data/Map/Strict/Internal.hs#L1454-L1455

https://github.com/haskell/containers/blob/734785d2d3ee862726c6c5bb30e6484fff370e50/containers/src/Data/IntMap/Internal.hs#L2578-L2579

https://github.com/haskell/containers/blob/734785d2d3ee862726c6c5bb30e6484fff370e50/containers/src/Data/IntMap/Internal.hs#L2593-L2595

https://github.com/haskell/containers/blob/734785d2d3ee862726c6c5bb30e6484fff370e50/containers/src/Data/IntMap/Strict/Internal.hs#L983-L984

It would be more efficient to skip the intermediate list and `foldl'` over the map accumulating into another map.

An exception here is `Map.mapKeys`, since `Map.fromList` is O(n) on an already sorted list.

This raises a related question of whether `fromListWith` should also have such an optimization.

meooow25:
* Implement the functions in terms of the builders instead of converting to and from a list. This reduces time and allocations.
* Document that they take linear time for monotonic functions.

Closes #1027

---

Benchmarks with GHC 9.10.1:

IntSet:
```
Name            Time - - - - - - - -    Allocated - - - - -
map:asc          36 μs   22 μs  -37%    456 KB  136 KB  -70%
map:random      547 μs  513 μs   -6%    2.9 MB  2.6 MB  -10%
mapMonotonic     35 μs   23 μs  -34%    456 KB  136 KB  -70%
```

IntMap:
```
Name                    Time - - - - - - - -    Allocated - - - - -
mapKeys:asc             101 μs   68 μs  -32%    960 KB  608 KB  -36%
mapKeys:random          594 μs  600 μs   +1%    3.0 MB  2.7 MB  -11%
mapKeysMonotonic         94 μs   63 μs  -32%    960 KB  608 KB  -36%
mapKeysWith:asc:dups    129 μs   61 μs  -52%    1.2 MB  512 KB  -58%
```

### DIFF
--- a/containers-tests/benchmarks/IntMap.hs
+++ b/containers-tests/benchmarks/IntMap.hs
@@ -10,7 +10,7 @@ import qualified Data.IntMap.Strict as MS
 import qualified Data.IntSet as S
 import Data.Maybe (fromMaybe)
 import Data.Word (Word8)
-import System.Random (StdGen, mkStdGen, randoms)
+import System.Random (StdGen, mkStdGen, random, randoms)
 import Prelude hiding (lookup)
 
 import Utils.Fold (foldBenchmarks, foldWithKeyBenchmarks)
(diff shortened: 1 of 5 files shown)


treeowl (approved):
Good idea.

meooow25:
Looks like we can get faster (~15%) if we work with the non-empty part of the builder separately, since we have, e.g.

```hs
data MonoState
  = MSNada
  | MSPush {-# UNPACK #-} !Int {-# UNPACK #-} !BitMap !Stack
```

If we get to first entry separately to avoid the `MSNada`, the `MSPush` can be unpacked in the rest of the fold. This isn't something GHC can figure out, though it is a bit like SpecConstr.

The downside is that we'll need a few more functions operating on non-empty `MonoState`, but it shouldn't be too messy.

PR #1157, opened 2025-08-13
Add light weight union ptr equality check 

HuwCampbell:
It's not unusual when building compilers to do transformations which require collecting sets over an AST – but you'll often need to do branching and unions across very similar sets, for example, one side added 2 items into a set of 1000. This is currently quite slow.

Adding a light weight pointer equality check can actually help quite a lot in these cases.

If I add some small additional benchmarks to union as such:
```haskell
defaultMain
      [ bench "union_same" $ whnf (S.union s_even) s_even
      , bench "union_minor_diff" $ whnf (S.union s_even) (S.insert 1 s_even)
      , bench "union" $ whnf (S.union s_even) s_odd
      ]
```

I'll see these results:

```
Before:
  union_same:       OK
    11.4 μs ± 863 ns
  union_minor_diff: OK
    11.3 μs ± 850 ns
  union:            OK
    77.9 μs ± 7.0 μs

After:
  union_same:       OK
    2.02 ns ± 138 ps
  union_minor_diff: OK
    137  ns ± 7.2 ns
  union:            OK
    77.4 μs ± 6.9 μs
```

### DIFF
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -831,7 +831,8 @@ unions = Foldable.foldl' union empty
 -- | \(O\bigl(m \log\bigl(\frac{n}{m}+1\bigr)\bigr), \; 0 < m \leq n\). The union of two sets, preferring the first set when
 -- equal elements are encountered.
 union :: Ord a => Set a -> Set a -> Set a
-union t1 Tip  = t1
+union t1 t2 | t1 `ptrEq` t2 = t1
+union t1 Tip = t1
 union t1 (Bin 1 x _ _) = insertR x t1
 union (Bin 1 x _ _) t2 = insert x t2
 union Tip t2  = t2
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -836,10 +836,8 @@ union t1 Tip = t1
 union t1 (Bin 1 x _ _) = insertR x t1
 union (Bin 1 x _ _) t2 = insert x t2
 union Tip t2  = t2
-union t1@(Bin _ x l1 r1) t2 = case splitS x t2 of
-  (l2 :*: r2)
-    | l1l2 `ptrEq` l1 && r1r2 `ptrEq` r1 -> t1
-    | otherwise -> link x l1l2 r1r2
+union (Bin _ x l1 r1) t2 = case splitS x t2 of
+  (l2 :*: r2) -> link x l1l2 r1r2
     where !l1l2 = union l1 l2
           !r1r2 = union r1 r2
 #if __GLASGOW_HASKELL__


HuwCampbell:
Forgive me, there is one case where the other check (which the second commit removes) does provide an improvement.

If we imagine creating the same set in completely separate manners (i.e., they don't share any pointers) then the insertR rule will kick in and you'll get a 25% speed up saving on rebalance costs.

```
Before
  union_same_but_different:   OK
    29.0 μs ± 2.0 μs

After
  union_same_but_different:   OK
    40.0 μs ± 3.4 μs
```

I think this best case 25% improvement is massively dwarfed by the 5000% we see elsewhere.

HuwCampbell:
There's a pretty reasonable question to be asked about whether this optimisation should also be applied to maps. The `insert` operation there does quite a bit dance on pointer equalities effectively just for maps structurally including the same items (and noting this is often only true for union).

treeowl:
I'm having a bit of trouble understanding the exact difference here. At first glance, it looks like a simplification of some sort, but it's not immediately clear if it catches more or fewer cases. Could you walk me through it?

meooow25:
I'm not convinced that this is a good idea. You're proposing a *time optimization* that applies when the two maps being union-ed happen to have identical (by pointer) subtrees. This might be happening commonly for your case, but there is no reason to expect this in general and the price to pay is a linear overhead for everybody else.

This is also completely orthogonal to the pointer checks you have removed. Those are a *memory optimization*, to avoid allocating a result `Bin` when it would be equal to the input `Bin`.

HuwCampbell:
I don't particularly mind putting back the second pointer equality checks. I just thought that with the proposed change they're less likely to be hit.

But let's talk it through. The existing code and checks are actually very good at dealing with trees of different balance, so two sets which are `Eq` equal should turn out with the balance (tree structure) of the first argument, and indeed, in the end, share a pointer (I don't have a proof of this).

So what we had before will lower allocations and be faster for union `Eq` equal sets, or sets with significant chunks which are `(==)`.

```haskell
-- These two have the same elements, but are structured differently
s1 = Set.fromList [1,4,3,2,5]
-- 3
-- +--1
-- |  +--|
-- |  +--2
-- +--4
--    +--|
--    +--5
s2 = Set.fromList [1,5,2,3,4]
-- 2
-- +--1
-- +--4
--    +--3
--    +--5
s3 = Set.union s1 s2
-- ptrEq s1 s3 should be True
```

But where do these come from? In what scenarios do we create two sets with the same elements and then union them? In my case, it was because I was working with two sets which share a common history and have been rejoined, and I hypothesised that for large sets it's probably the main way this will happen.

So my thought was that we should be able to see if this is actually test if this is the case and just short-circuit where we see it.

For performance in cases with no sharing, the cost is very modest; 77.9 to 78.1 microseconds – lets go 1%. The cost of the 2 pointer checks also checks in at about 1% each above (so 2%).

But, there are occasional _significant_ wins with 100x seen above when we can effectively short circuit out all but one path to a new element.

meooow25:
Sorry, I've been a little too busy to look at this properly until now.

My main concern was with the overhead of the added check, but you're right that that does seem to be minimal.

I'm still a little uneasy about relying on pointer-equality checks, especially when it drastically affects the running time. For example, it wouldn't be great if a user makes a small change and their program becomes much slower because the check doesn't trigger now. But I haven't been able to think of such a situation so far.

This change would also open the door to similar checks in other operations (intersection, difference, isSubsetOf, etc.) and other structures, so we have to consider that too.

I'm also not aware of any other Haskell library using pointer-equality for short-circuits. That could have made the decision easier.

@treeowl do you have any opinions on this?

PR #1159, opened 2025-09-09
Add `catMaybes` and `mapMaybe` to sets, maps and `Seq`

ChickenProp:
As discussed in #346. We now have `mapMaybe` in all of `Seq`, `Set`, `IntSet`, strict and lazy `Map`, and strict and lazy `IntMap`. And `catMaybes` in all of those except `IntSet` (where it doesn't make sense). The issue doesn't mention all of them, but I figured it would be good to have a consistent API.

### DIFF
--- a/containers-tests/tests/intset-properties.hs
+++ b/containers-tests/tests/intset-properties.hs
@@ -6,6 +6,7 @@ import Data.IntSet
 import Data.List (nub,sort)
 import qualified Data.List as List
 import Data.Maybe (listToMaybe)
+import qualified Data.Maybe as Maybe
 import Data.Monoid (mempty)
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NE
@@ -79,6 +80,7 @@ main = defaultMain $ testGroup "intset-properties"
                    , testProperty "prop_splitRoot" prop_splitRoot
                    , testProperty "prop_partition" prop_partition
                    , testProperty "prop_filter" prop_filter
+                   , testProperty "prop_mapMaybe" prop_mapMaybe
                    , testProperty "takeWhileAntitone" prop_takeWhileAntitone
                    , testProperty "dropWhileAntitone" prop_dropWhileAntitone
                    , testProperty "spanAntitone" prop_spanAntitone
@@ -456,6 +458,12 @@ prop_filter s i =
   in valid odds .&&.
      valid evens .&&.
      parts === (odds, evens)
+
+prop_mapMaybe :: IntSet -> Property
+prop_mapMaybe s =
+  let f n = if odd n then Just n else Nothing
+      odds = mapMaybe f s
+  in valid odds .&&. toList odds === Maybe.mapMaybe f (toList s)
 
 prop_takeWhileAntitone :: Int -> [Int] -> Property
 prop_takeWhileAntitone x ys =
--- a/containers-tests/tests/set-properties.hs
+++ b/containers-tests/tests/set-properties.hs
@@ -2,7 +2,8 @@
 import qualified Data.IntSet as IntSet
 import Data.List (nub, sort, sortBy)
 import qualified Data.List as List
-import Data.Maybe
+import Data.Maybe (isJust, fromJust)
+import qualified Data.Maybe as Maybe
 import Data.Set
 import Data.Set.Internal (link, merge)
 import Prelude hiding (lookup, null, map, filter, foldr, foldl, foldl', all, take, drop, splitAt)
@@ -98,6 +99,8 @@ main = defaultMain $ testGroup "set-properties"
                    , testProperty "prop_splitRoot" prop_splitRoot
                    , testProperty "prop_partition" prop_partition
                    , testProperty "prop_filter" prop_filter
+                   , testProperty "prop_catMaybes" prop_catMaybes
+                   , testProperty "prop_mapMaybe" prop_mapMaybe
                    , testProperty "takeWhileAntitone"    prop_takeWhileAntitone
                    , testProperty "dropWhileAntitone"    prop_dropWhileAntitone
                    , testProperty "spanAntitone"         prop_spanAntitone
@@ -240,6 +243,32 @@ instance IsInt a => Arbitrary (Set a) where
         let i' = i + diff
         put i'
         pure (fromInt i')
+
+newtype MaybeIntSet = MaybeIntSet (Set (Maybe Int)) deriving (Show)
+instance Arbitrary MaybeIntSet where
+  arbitrary = sized $ \sz0 -> do
+    sz <- choose (0, sz0)
+    middle <- choose (-positionFactor * (sz + 1), positionFactor * (sz + 1))
+    let shift = (sz * (gapRange) + 1) `quot` 2
+        start = middle - shift
+    hasNothing <- arbitrary
+    t <- evalStateT (mkArbSet (step start) sz)
+                    (if hasNothing then Nothing else Just start)
+    if valid t
+      then pure $ MaybeIntSet t
+      else error "Test generated invalid tree!"
+    where
+      step start = do
+        mi <- get
+        case mi of
+          Nothing -> do
+            put $ Just start
+            pure Nothing
+          Just i -> do
+            diff <- lift $ choose (1, gapRange)
+            let i' = i + diff
+            put $ Just i'
+            pure $ Just i'
 
 data TwoSets = TwoSets (Set Int) (Set Int) deriving (Show)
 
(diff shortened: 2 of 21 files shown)


containers/src/Data/Sequence/Internal.hs
```diff
@@ -4199,6 +4201,22 @@ partition p = toPair . foldl' part (empty :*: empty)
...
+
+-- | \( O(n) \). Map values and collect the 'Just' results.
+--
+-- @since FIXME
+mapMaybe :: (a -> Maybe b) -> Seq a -> Seq b
+mapMaybe f = foldl' go empty
+  where go xs x = case f x of
+          Nothing -> xs
+          Just x' -> xs `snoc'` x'
+
```
treeowl, line 4211: I'm somewhat skeptical about this being the most efficient implementation. How does it compare to one that uses an intermediate list?

ChickenProp: No idea :sweat_smile: I don't currently know how to benchmark these things but I can look into it. I just copied `filter` so I assumed it would be fine (though I used a `where` instead of an inline function, dunno if that makes a difference).

treeowl: Oh, if filter does it, it's probably right. Never mind.

treeowl:
I also don't think we really need `catMaybes`, but maybe it would serve some didactic purpose? Do we want an effectful version?

```haskell
traverseMaybe :: Applicative f => (a -> f (Maybe b)) -> Whatever a -> f (Whatever b)
```

containers/src/Data/IntSet/Internal.hs
```diff
@@ -880,6 +882,18 @@ filter predicate t
                          | otherwise           = bm
         {-# INLINE bitPred #-}
 
+-- | \( O(n \log n) \). Map values and collect the 'Just' results.
```
meooow25: 1. The complexity should be O(n min(n,W)) (same as `map`).
2. I would prefer "elements" to "values", since it's a set. Same for `Set` and `Seq`.

containers-tests/tests/intset-properties.hs
```diff
@@ -457,6 +459,12 @@ prop_filter s i =
      valid evens .&&.
      parts === (odds, evens)
 
+prop_mapMaybe :: IntSet -> Property
+prop_mapMaybe s =
+  let f n = if odd n then Just n else Nothing
```
meooow25: Consider using an arbitrary `Fun` instead, makes the test simpler.
Same for the other `mapMaybe` property tests.

containers/src/Data/Sequence/Internal.hs
meooow25: It is reasonable for `filter` and `mapMaybe` to have similar behavior (as I wrote in https://github.com/haskell/containers/issues/346#issuecomment-2726199830), so I think this is fine.

We can check later if both can be improved.

containers/src/Data/Set/Internal.hs
```diff
@@ -993,6 +996,25 @@ partition p0 t0 = toPair $ go p0 t0
...
+catMaybes = mapMonotonic Maybe.fromJust . dropWhileAntitone Maybe.isNothing
+
+-- | \(O(n \log n)\). Map values and collect the 'Just' results.
+--
+-- If the function is monotonically non-decreasing, this function takes \(O(n)\)
+-- time.
+--
+-- @since FIXME
+mapMaybe :: Ord b => (a -> Maybe b) -> Set a -> Set b
+mapMaybe f = fromList . Maybe.mapMaybe f . toList
```
meooow25: We can do better here with a `foldl'` in the style of `map`:

https://github.com/haskell/containers/blob/98ea59e234d8a67817e01580b436b1ff2560c7b1/containers/src/Data/Set/Internal.hs#L1033-L1034

This is can be confirmed with benchmarks, e.g. if I make this change

```diff
diff --git a/containers-tests/benchmarks/Set.hs b/containers-tests/benchmarks/Set.hs
index 9e367391..1fff8fce 100644
--- a/containers-tests/benchmarks/Set.hs
+++ b/containers-tests/benchmarks/Set.hs
@@ -9,2 +9,3 @@ import Data.List (foldl')
 import qualified Data.Set as S
+import qualified Data.Set.Internal as S

@@ -26,2 +27,4 @@ main = do
         , bench "map:desc" $ whnf (S.map (negate . (+ 1))) s
+        , bench "mapMaybe1" $ whnf (S.mapMaybe1 Just) s
+        , bench "mapMaybe2" $ whnf (S.mapMaybe2 Just) s
         , bench "filter" $ whnf (S.filter ((== 0) . (`mod` 2))) s
diff --git a/containers/src/Data/Set/Internal.hs b/containers/src/Data/Set/Internal.hs
index 8b6be2e8..d9a057fa 100644
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -228,4 +228,8 @@ module Data.Set.Internal (
             , merge
+
+            , mapMaybe1
+            , mapMaybe2
             ) where

+import qualified Data.Maybe as Maybe
 import Utils.Containers.Internal.Prelude hiding
@@ -1016,2 +1020,12 @@ map f t = finishB (foldl' (\b x -> insertB (f x) b) emptyB t)

+mapMaybe1, mapMaybe2 :: Ord b => (a -> Maybe b) -> Set a -> Set b
+mapMaybe1 f = fromList . Maybe.mapMaybe f . toList
+mapMaybe2 f t = finishB (foldl' g emptyB t)
+  where
+    g b x = case f x of
+      Nothing -> b
+      Just y -> insertB y b
+{-# INLINABLE mapMaybe1 #-}
+{-# INLINABLE mapMaybe2 #-}
+
 -- | \(O(n)\).
```

and run the benchmarks, I get

```
$ cabal run set-benchmarks -- -p mapMaybe +RTS -T
All
  mapMaybe1: OK
    147  μs ± 6.5 μs, 704 KB allocated,  14 KB copied, 8.0 MB peak memory
  mapMaybe2: OK
    113  μs ± 8.0 μs, 416 KB allocated, 8.3 KB copied, 8.0 MB peak memory
```

meooow25 (changes requested):

containers/src/Data/Set/Internal.hs
meooow25: Oh and please mark it `INLINABLE`. This will allow specialization on `Ord b` where it's used.

meooow25:
> Do we want an effectful version?
> ```hs
> traverseMaybe :: Applicative f => (a -> f (Maybe b)) -> Whatever a -> f (Whatever b)
> ```

That's a good question, but there's a problem since we are doing a tail-recursive construction with an accumulator. To keep the same style, we need `Monad`. If we want `Applicative` I think we would need to sacrifice some efficiency.

I'm fine with considering this later, doesn't have to be in this PR.

ChickenProp:
Thanks for the comments, will get to them at some point.

* `catMaybes`: I don't have strong feelings about this. I'm not sure I've ever wanted them. For `Set` (and nothing else), maybe `catMaybes` would be more efficient than `mapMaybe id` (though both are O(n))? I can benchmark later if it's a crux for anything. I think two reasons to keep are

    * Consistency with `Data.Maybe`;
    * `traverseMaybe` and `mapMaybeWithIndex` can both be pretty easily implemented with it, so maybe it would have at least some use.

  But they're not very strong, I'm also fine with removing, and it's easier to add them later if we want them than to remove if we don't.

* `traverseMaybe`: agree this can be postponed.

* `WithKey` variants: `Map` and `IntMap` already have `mapMaybeWithKey`. `Seq` has `mapWithIndex` so maybe it wants `mapMaybeWithIndex` too. I don't think `Set` or `IntSet` have a comparable variant of `map`, so I wouldn't want to add it for `mapMaybe`.

treeowl:
For `mapMaybe` for `Map`, shouldn't we be able to follow the map structure instead, and build up the resulting structure using `join` and such? How would that compare in benchmarks?

Also, I still want `WithKey` versions.

ChickenProp:
`mapMaybe` already exists in `Map`. If it can be improved I'd prefer that go in a separate PR to avoid balooning this one.

> Also, I still want `WithKey` versions.

To clarify, `WithKey` versions of what? The only place I'd expect functions named `mapMaybeWithKey` are in `Map` and `IntMap`, and those already have them.

treeowl:
Sorry, I seem to have half-remembered some bits. I need to look through again.

containers-tests/tests/set-properties.hs
```diff
@@ -241,6 +243,32 @@ instance IsInt a => Arbitrary (Set a) where
...
+        case mi of
+          Nothing -> do
+            put $ Just start
+            pure Nothing
+          Just i -> do
+            diff <- lift $ choose (1, gapRange)
+            let i' = i + diff
+            put $ Just i'
+            pure $ Just i'
+
```
meooow25: This seems to be no longer necessary.

containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -989,6 +989,7 @@ mapKeysWith c f t =
 {--------------------------------------------------------------------
   Filter
 --------------------------------------------------------------------}
+
```
meooow25: This file doesn't need to be changed.

PR #1160, opened 2025-09-10
Adding HasCallStack to Partial functions

L0neGamer:
continuing the work of https://github.com/haskell/containers/pull/493/

### DIFF
--- a/containers/changelog.md
+++ b/containers/changelog.md
@@ -13,6 +13,9 @@
 * Add `fromSetA` for `Map` and `IntMap`. (L0neGamer)
   ([#1163](https://github.com/haskell/containers/pull/1163))
 
+* Added `HasCallStack` to many partial functions when compiling for GHC. (L0neGamer & dwincort)
+  ([#1160](https://github.com/haskell/containers/pull/1160))
+
 ### Performance improvements
 
 * Improved performance for `Data.IntMap.restrictKeys` and


containers/src/Data/IntMap/Internal.hs
```diff
@@ -2378,7 +2389,11 @@ kvToTuple :: KeyValue a -> (Key, a)
 kvToTuple (KeyValue k x) = (k, x)
 {-# INLINE kvToTuple #-}
 
+#if __GLASGOW_HASKELL__ >= 800
+lookupMinSure :: HasCallStack => IntMap a -> KeyValue a
```
treeowl: This is an internal function. We call it only when we know the map isn't empty.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -2394,12 +2409,20 @@ lookupMin (Bin p l r) =
...
+findMin :: HasCallStack => IntMap a -> (Key, a)
+#else
 findMin :: IntMap a -> (Key, a)
+#endif
 findMin t
   | Just r <- lookupMin t = r
   | otherwise = error "findMin: empty map has no minimal element"
 
+#if __GLASGOW_HASKELL__ >= 800
+lookupMaxSure :: HasCallStack => IntMap a -> KeyValue a
```
treeowl: This one is internal too. Don't add a constraint here.

containers/src/Data/Sequence/Internal.hs
```diff
@@ -2401,7 +2429,11 @@ scanl f z0 xs = z0 <| snd (mapAccumL (\ x z -> let x' = f x z in (x', x')) z0 xs
 -- __Note__: This function is partial.
 --
 -- > scanl1 f (fromList [x1, x2, ...]) = fromList [x1, x1 `f` x2, ...]
+#if __GLASGOW_HASKELL__ >= 800
+scanl1 :: HasCallStack => (a -> a -> a) -> Seq a -> Seq a
```
treeowl, line 2427: I'm concerned this could build a call stack recursively. Check the Core, please.

containers/src/Data/Sequence/Internal.hs
```diff
@@ -2415,7 +2447,11 @@ scanr f z0 xs = snd (mapAccumR (\ z x -> let z' = f x z in (z', z')) z0 xs) |> z
 -- Calls 'error' if the sequence is empty.
 --
 -- __Note__: This function is partial.
+#if __GLASGOW_HASKELL__ >= 800
+scanr1 :: HasCallStack => (a -> a -> a) -> Seq a -> Seq a
```
treeowl, line 2445: Also concerned about recursion here.

containers/src/Data/Sequence/Internal.hs
L0neGamer: scanl1_hascallstack
Here's a rough approximation of the core. I think my attempts to clean up got rid of a variable on the call to scanl2, but as can be seen there is very little difference; the only real difference being what callstack the error is called with.

containers/src/Data/Sequence/Internal.hs
L0neGamer: scanr1_hascallstack
Similarly here's scanr1 with the same comments.

containers/src/Data/Sequence/Internal.hs
treeowl: Oh good. I guess GHC cuts out the detected recursion when automatically pushing. It's unfortunate that we push at all (on error), but that's the system.

meooow25:
Hi, I need some time to review this, since I need to first understand how `HasCallstack` works and the current consensus on using them in boot libraries.

About the benchmarks, what have you tried? I would run the benchmarks for the changed functions using tasty patterns, e.g.

```
$ cabal run map-benchmarks -- -p lookup
```

(See also [CONTRIBUTING.md](https://github.com/haskell/containers/blob/master/CONTRIBUTING.md#procedure))

L0neGamer:
I ran `cabal bench all` before the changes, after, and then checked out master again and reran, and got very different results. 

I can certainly try running individual benchmarks; in that case I'll probably try doing "run each benchmark component individually".

treeowl:
@meooow25 , the constraints are used in `base`, so there shouldn't be any issue using them in other boot libraries.

treeowl:
FWIW, I always run `killall -STOP firefox` before running benchmarks and `killall -CONT firefox` after, but I don't claim to be good at achieving a "quiet" system.

jappeace:
It needs to calculate the standard deviations (also known as expected error, or volatility) to see if they're within margin per bench suite.

But I think these are kind of excessive. Nobody complained about adding callstack to unordered containers, so I doubt they will if it's added here. And if they do just revert the ones they complain about 😅

Bodigrim:
> Here are the percentage differences of the end results of each:

What do you mean by "end results"? Are you comparing total time as reported in "All XX tests passed"? This would make little sense.

Could you use `--csv` and `--baseline`?

meooow25:
> Would the maintainers be interested if I took this opportunity to add the `x-partial` warning to all of these functions as well?

A no from me on that, sorry. I find the warnings on `head` and `tail` annoying enough.
I'm all for documenting clearly which functions are partial and the conditions under which they error. This was done in #1140, if I missed any do let me know. The rest is on the user, if they use the function they know what they are doing and there is no reason to annoy them when they compile their code.

L0neGamer:
As much as I would love to add `x-partial` warnings onto these functions, I don't think adding HasCallStack should be blocked on that. What do I need to do to keep this moving?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -366,6 +366,9 @@ import Text.Read
 #endif
 import qualified Control.Category as Category
 
+#if __GLASGOW_HASKELL__ >= 800
```
meooow25: We support only GHC >=8.2, so all of these can simply be `#ifdef __GLASGOW_HASKELL__`

https://github.com/haskell/containers/blob/51902237d52dbb10002c5ac897330a584db8e671/containers/containers.cabal#L41-L42

containers/src/Data/Map/Strict/Internal.hs
```diff
@@ -820,6 +824,21 @@ atKeyIdentity k f t = Identity $ atKeyPlain Strict k (coerce f) t
 -- > updateAt (\_ _  -> Nothing)  2    (fromList [(5,"a"), (3,"b")])    Error: index out of range
 -- > updateAt (\_ _  -> Nothing)  (-1) (fromList [(5,"a"), (3,"b")])    Error: index out of range
 
+#if __GLASGOW_HASKELL__ >= 800
+updateAt :: HasCallStack => (k -> a -> Maybe a) -> Int -> Map k a -> Map k a
+updateAt = go where
```
meooow25, line 829: It's not worth having two copies of this, let's keep only this version.

containers/src/Data/Map/Internal.hs
```diff
@@ -438,8 +441,18 @@ infixl 9 !,!?,\\ --
...
-(!) m k = find k m
+#endif
+(!) m !k = go m
+  where
+    go Tip = error "Map.!: given key is not an element in the map"
+    go (Bin _ kx x l r) = case compare k kx of
+      LT -> go l
+      GT -> go r
+      EQ -> x
 {-# INLINE (!) #-}
```
meooow25: This should be `INLINABLE`, as `find` was.

Also note that you have changed the implementation so the `go` closes over `k` instead of taking it as argument. But I don't think that should have an effect since it has to close over the CallStack now anyway.

containers/src/Data/Map/Internal.hs
```diff
@@ -604,17 +617,11 @@ member = go
 
 notMember :: Ord k => k -> Map k a -> Bool
 notMember k m = not $ member k m
+#if __GLASGOW_HASKELL__
```
meooow25: Please remove, we got rid of these in #1167.

containers/src/Data/Map/Internal.hs
```diff
@@ -1469,6 +1480,18 @@ lookupIndex = go 0
 -- > elemAt 1 (fromList [(5,"a"), (3,"b")]) == (5, "a")
 -- > elemAt 2 (fromList [(5,"a"), (3,"b")])    Error: index out of range
 
+#if __GLASGOW_HASKELL__ >= 800
+elemAt :: HasCallStack => Int -> Map k a -> (k,a)
+elemAt = go where
```
meooow25, line 1486: Let's keep just one version for the functions here too.

containers/src/Data/Sequence/Internal.hs
```diff
@@ -516,7 +520,11 @@ instance MonadFix Seq where
 -- This is just like the instance for lists, but we can take advantage of
 -- constant-time length and logarithmic-time indexing to speed things up.
 -- Using fromFunction, we make this about as lazy as we can.
+#if __GLASGOW_HASKELL__ >= 800
+mfixSeq :: HasCallStack => (a -> Seq a) -> Seq a
```
meooow25: This is an internal function, HasCallStack isn't useful here.

containers/src/Data/Set/Internal.hs
```diff
@@ -1514,6 +1530,18 @@ lookupIndex = go 0
 --
 -- @since 0.5.4
 
+#if __GLASGOW_HASKELL__ >= 800
+elemAt :: HasCallStack => Int -> Set a -> a
```
meooow25, line 1534: Let's keep just one version for the functions here too.

containers/src/Data/Sequence/Internal.hs
treeowl: The reason it's not useful, I believe, is not that it's internal. The reason is that `mfix` is a class method that does not take a call stack. So there's no way to get the desired call stack info to the right place.

L0neGamer: It would've been good to HasCallStack the mfix, but you are right that we can't do that.

containers/src/Data/Map/Internal.hs
```diff
@@ -1469,15 +1474,20 @@ lookupIndex = go 0
...
-      EQ -> (kx,x)
-  where
-    sizeL = size l
+#endif
+elemAt = go where
+  go !_ Tip = error "Map.elemAt: index out of range"
+  go i (Bin _ kx x l r)
+    = case compare i sizeL of
+        LT -> elemAt i l
+        GT -> elemAt (i-sizeL-1) r
```
meooow25: These should be `go` instead of `elemAt`.

meooow25:
This seems good overall. Benchmarks don't show anything of concern. I'm adding some more in #1182 and they seem okay too.

This will be a breaking change, so I can approve it but it will be sitting for a while until we have a major release.

L0neGamer:
I've fixed the `elemAt` recursion and added a changelog entry, which can be moved or adjusted as needed.

I'm not sure why this is a breaking change considering that none of the exports have changed and none of the functions should've changed in functionality.

The only potential change I can see is where the error is reported from; see below for the example change.

definitions:
image

usages:
image

You can see that `accessHCS` has a more useful callstack, but otherwise everything is identical.

meooow25 (approved):

PR #1162, opened 2025-10-10
Switch from template-haskell to template-haskell-lift

TeofilC:
This new boot library should be more stable than template-haskell and
should eventually allow us to remove much of the CPP around TH.

It will also make it easier for end-users to reinstall template-haskell
as it will no longer be used by any boot libraries

This GHC MR tests this PR against GHC-HEAD: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/14978

### DIFF
--- a/containers-tests/tests/seq-properties.hs
+++ b/containers-tests/tests/seq-properties.hs
@@ -46,14 +46,18 @@ import Control.DeepSeq (deepseq)
 import Control.Monad.Fix (MonadFix (..))
 import Test.Tasty.HUnit
 import Test.ChasingBottoms.IsBottom (isBottom)
+#if __GLASGOW_HASKELL__ >= 914
+import qualified Language.Haskell.TH.Lift as TH
+#else
 import qualified Language.Haskell.TH.Syntax as TH
+#endif
 
 import Utils.Strictness (Bot(..), Func2, applyFunc2)
 
 main :: IO ()
 main = defaultMain $ testGroup "seq-properties"
        [ test_lift
-#if MIN_VERSION_template_haskell(2,16,0)
+#if __GLASGOW_HASKELL__ >= 810
        , test_liftTyped
 #endif
        , testProperty "fmap" prop_fmap
@@ -987,7 +991,7 @@ test_lift = testCase "lift" $ do
   (mempty :: Seq Int) @=? $([| $(TH.lift (fromList [] :: Seq Integer)) |])
   fromList [1..3 :: Int] @=? $([| $(TH.lift (fromList [1..3 :: Integer])) |])
 
-#if MIN_VERSION_template_haskell(2,16,0)
+#if __GLASGOW_HASKELL__ >= 810
 test_liftTyped :: TestTree
 test_liftTyped = testCase "liftTyped" $ do
   (mempty :: Seq Int) @=? $$([|| $$(TH.liftTyped (fromList [])) ||])
--- a/containers/containers.cabal
+++ b/containers/containers.cabal
@@ -42,7 +42,12 @@ Library
         base >= 4.10 && < 5
       , array >= 0.4.0.0
       , deepseq >= 1.4.3.0 && < 1.6
-    if impl(ghc)
+    -- template-haskell-lift was added as a boot library in GHC-9.14
+    -- once we no longer wish to backport releases to older major releases,
+    -- this conditional can be dropped
+    if impl(ghc >= 9.14)
+       build-depends: template-haskell-lift >= 0.1 && <0.2
+    elif impl(ghc)
        build-depends: template-haskell
     hs-source-dirs: src
     ghc-options: -O2 -Wall
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -358,9 +358,13 @@ import Data.Data (Data(..), Constr, mkConstr, constrIndex,
 import qualified Data.Data as Data
 import GHC.Exts (build)
 import qualified GHC.Exts as GHCExts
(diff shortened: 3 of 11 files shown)


TeofilC:
template-haskell-lift is compatible with GHC>=8.10. But it is only a boot library with GHC>=9.14. So you wouldn't be able to backport new releases of containers to older GHCs as a boot lib (they would build fine from Hackage though).

Let me know what you'd prefer.

Eventually though, `template-haskell-lift` should let us remove all the TH CPP since that should all in that package.

treeowl:
That's only a potential issue for new releases on old GHC branches, right? I suppose it would be okay to have some CPP temporarily to support that as long as such branches are "supported" by GHC HQ. But wouldn't it be cleaner to just add `template-haskell-lift` as a boot package in those releases?

containers/containers.cabal
```diff
@@ -42,7 +42,12 @@ Library
         base >= 4.10 && < 5
       , array >= 0.4.0.0
       , deepseq >= 1.4.3.0 && < 1.6
-    if impl(ghc)
+    -- template-haskell-lift was added as a boot library in GHC-9.14
+    -- once we no longer wish to backport releases to older major releases,
```
treeowl, line 46:
```suggestion
    -- once we no longer wish to backport releases to older major releases of GHC,
```

TeofilC: Done

containers-tests/tests/seq-properties.hs
```diff
@@ -46,14 +46,18 @@ import Control.DeepSeq (deepseq)
 import Control.Monad.Fix (MonadFix (..))
 import Test.Tasty.HUnit
 import Test.ChasingBottoms.IsBottom (isBottom)
+#  if __GLASGOW_HASKELL__ >= 914
```
meooow25: This is minor, but let's not have spaces after `#`. I think it would be less surprising (I'm surprised to learn that this is acceptable syntax).

containers/src/Data/IntMap/Internal.hs
```diff
@@ -358,9 +358,13 @@ import Data.Data (Data(..), Constr, mkConstr, constrIndex,
 import qualified Data.Data as Data
 import GHC.Exts (build)
 import qualified GHC.Exts as GHCExts
-import Language.Haskell.TH.Syntax (Lift)
+#  if __GLASGOW_HASKELL__ >= 914
+import Language.Haskell.TH.Lift (Lift(..))
+#  else
+import Language.Haskell.TH.Syntax (Lift(..))
```
meooow25: Why the change from `Lift` to `Lift(..)`? Same for other files.

containers/src/Data/Set/Internal.hs
```diff
@@ -255,10 +255,14 @@ import Text.Read ( readPrec, Read (..), Lexeme (..), parens, prec
...
-import Language.Haskell.TH.Syntax (Lift)
+#  if __GLASGOW_HASKELL__ >= 914
+import Language.Haskell.TH.Lift (Lift(..))
+#  else
+import Language.Haskell.TH.Syntax (Lift(..))
 -- See Note [ Template Haskell Dependencies ]
 import Language.Haskell.TH ()
-import Data.Coerce (coerce)
+#  endif
+import Data.Coerce
```
meooow25: Why change this line?

containers/src/Data/IntMap/Internal.hs
```diff
@@ -358,9 +358,13 @@ import Data.Data (Data(..), Constr, mkConstr, constrIndex,
...
 -- See Note [ Template Haskell Dependencies ]
 import Language.Haskell.TH ()
```
meooow25, line 366: So we don't need this any more if we depend on `template-haskell-lift`? The current reason is documented in [note-template-haskell-deps.md](https://github.com/haskell/containers/blob/5669b01afd9da9b8db242c8f9e0a5aa6b1a7c3f8/containers/note-template-haskell-deps.md)

containers-tests/tests/seq-properties.hs
treeowl: I agree it doesn't belong here, but in nested conditions it's very helpful to see what's going on visually.

TeofilC: Good spot. That was a copy-paste error

containers/src/Data/IntMap/Internal.hs
TeofilC: I don't think it's needed yeah.

containers/src/Data/Set/Internal.hs
TeofilC: Typo. Fixed now

containers/src/Data/IntMap/Internal.hs
TeofilC: Yep in fact we haven't needed it since GHC 9.12, I think, when these things moved to ghc-internal from template-haskell

meooow25: Nice!

containers-tests/tests/seq-properties.hs
meooow25: There are more of these in other files.

TeofilC: I've taken a look through them and all of them are inside `#ifdef __GLASGOW_HASKELL__` blocks so they are nested conditionals. I can drop the indention on these too of you'd like but that's what we seem to do elsewhere as well.

meooow25: Ah, I didn't realized they were all nested. Looks good then!

meooow25 (approved):

Issue #875, opened 2022-11-20
More set/map interactions


treeowl:
I was thinking today that it might be nice to have things like

```haskell
fromSetA :: Applicative f => (k -> f a) -> Set k -> f (Map k a)
fromSetWithIndexA :: Applicative f => (Int -> k -> f a) -> Set k -> f (Map k a)
fromSetWithIndex :: (Int -> k -> a) -> Set k -> Map k a
fromSetWithIndex f s = runIdentity (fromSetWithIndexA (coerce f) s)
```

These would let the user produce the values in an arbitrary `Applicative`, and the `WithIndex` version would let the passed function know where in the set it was working. For example,

```haskell
counting :: Set k -> Set k Int
counting = fromSetWithIndex (\i _ -> i)

-- counting (fromList "abc") = fromList [('a', 0), ('b', 1), ('c', 2)]
```

I'd also really love to get something like `mergeA` for `Set`, and ideally something similar that would *combine* sets with maps. Having to convert a set to a map first seems pretty bad.

meooow25:
I'm on board with adding `fromSetA`. See also: https://github.com/haskell/containers/issues/815#issuecomment-2764579287

I'm not too sure about the `WithIndex` versions.

PR #1163, opened 2025-10-28
Implement fromSetA for IntMap and Map

L0neGamer:
Initial stab at https://github.com/haskell/containers/issues/875

Happy to adjust, test, or benchmark as wanted. I'll look into benchmark comparisons for the changed functions later.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,17 @@ To do so on Windows 10 or higher, follow these steps:
 2. Enable git symlinks: `git config --global core.symlinks true`.
 3. Clone the repository again once git is properly configured.
 
+### Benchmarking script
+
+To run the all benchmarks of your branch against master, you can run the script
+`./bench.sh` while on your feature branch.
+
+This first builds and benchmarks against the master branch, and then builds and
+benchmarks on your branch, and compares between them.
+
+You can also fiddle with the script as long as the changes are reflected in your
+local master branch and run less than the full suite of benchmarks.
+
 ## Sending Pull Requests
 
 When you send a pull request, please:


containers/src/Data/IntMap/Internal.hs
```diff
@@ -3309,9 +3310,18 @@ keysSet (Bin p l r)
...
+
+-- | \(O(n)\). Build a map from a set of keys and a function which for each key
+-- computes its value, while within an 'Applicative' context.
+--
+-- > fromSetA (\k -> pure $ replicate k 'a') (Data.IntSet.fromList [3, 5]) == pure (fromList [(5,"aaaaa"), (3,"aaa")])
+-- > fromSetA undefined Data.IntSet.empty == pure empty
+
+fromSetA :: Applicative f => (Key -> f a) -> IntSet -> f (IntMap a)
+fromSetA _ IntSet.Nil = pure Nil
+fromSetA f (IntSet.Bin p l r) = liftA2 (Bin p) (fromSetA f l) (fromSetA f r)
```
treeowl: You have to watch out for sign at the root to get the right effect order. Compare `traverseWithKey`.

containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -1056,25 +1058,43 @@ mapEitherWithKey f0 t0 = toPair $ go f0 t0
...
-          _ -> case bits `iShiftRL` 1 of
-                 bits2 | bmask .&. ((1 `shiftLL` bits2) - 1) == 0 ->
-                           buildTree g (prefix + bits2) (bmask `shiftRL` bits2) bits2
-                       | (bmask `shiftRL` bits2) .&. ((1 `shiftLL` bits2) - 1) == 0 ->
-                           buildTree g prefix bmask bits2
-                       | otherwise ->
-                           Bin (Prefix (prefix .|. bits2)) (buildTree g prefix bmask bits2) (buildTree g (prefix + bits2) (bmask `shiftRL` bits2) bits2)
+fromSet f = runIdentity . fromSetA f'
+  where
+  f' k = let fk = f k in fk `seq` pure fk
```
treeowl:
```suggestion
  f' = coerce f
```
(Actually, just inline that definition.)
You can't do the forcing you're trying to do within the `Identity` context; you have to do it in the strict version of `fromSetA`. To explain a bit more, this code claims to force `f k` and then "wrap" it, but the "wrapper" is a newtype constructor, so it doesn't actually exist operationally, so you're really saying "force `f k` and return the result", which in Haskell is exactly the same as saying "return `f k`".

containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -1056,25 +1058,43 @@ mapEitherWithKey f0 t0 = toPair $ go f0 t0
...
+--
+-- This can only be as strict as the 'Applicative' allows it to be.
+--
+-- > fromSetA (\k -> pure $ replicate k 'a') (Data.IntSet.fromList [3, 5]) == pure (fromList [(5,"aaaaa"), (3,"aaa")])
+-- > fromSetA undefined Data.IntSet.empty == pure empty
+
+fromSetA :: Applicative f => (Key -> f a) -> IntSet.IntSet -> f (IntMap a)
+fromSetA _ IntSet.Nil = pure Nil
+fromSetA f (IntSet.Bin p l r) = liftA2 (Bin p) (fromSetA f l) (fromSetA f r)
+fromSetA f (IntSet.Tip kx bm) = buildTree f kx bm (IntSet.suffixBitMask + 1)
```
treeowl, line 1095: Same wibble as above about effect order.

containers/src/Data/IntMap/Strict/Internal.hs
treeowl: (And you actually do seem to force it where you're supposed to, though I'm still reading through that, so just remove this redundancy.)

containers/src/Data/Map/Internal.hs
```diff
@@ -3535,8 +3536,21 @@ argSet (Bin sz kx x l r) = Set.Bin sz (Arg kx x) (argSet l) (argSet r)
 -- > fromSet undefined Data.Set.empty == empty
 
 fromSet :: (k -> a) -> Set.Set k -> Map k a
-fromSet _ Set.Tip = Tip
-fromSet f (Set.Bin sz x l r) = Bin sz x (f x) (fromSet f l) (fromSet f r)
+fromSet f = runIdentity . fromSetA (pure . f)
```
treeowl, line 3542:
```suggestion
fromSet f = runIdentity . fromSetA (coerce f)
```

We don't want to build a new closure that indirectly calls `f`, which is what `pure . f` will be. (It has to do with detailed semantics of function arity; `undefined` is bottom but `Identity . undefined` is not.)

You may need to use some type annotation on the coercion; I haven't checked.

containers/src/Data/Map/Strict/Internal.hs
```diff
@@ -1455,8 +1454,25 @@ mapKeysWith c f m =
 -- > fromSet undefined Data.Set.empty == empty
 
 fromSet :: (k -> a) -> Set.Set k -> Map k a
-fromSet _ Set.Tip = Tip
-fromSet f (Set.Bin sz x l r) = case f x of v -> v `seq` Bin sz x v (fromSet f l) (fromSet f r)
+fromSet f = runIdentity . fromSetA f'
+  where
+  f' k = let fk = f k in fk `seq` pure fk
```
treeowl: Again, use `runIdentity . fromSetA (coerce f)`.

treeowl (commented):
Whoops ... I forgot that you'll need CPP around the coercions.

containers/src/Data/IntMap/Strict/Internal.hs
L0neGamer: forgot about the newtype transparency issue, thanks!

This would've worked if I was using `Solo` for whatever reason, but there is not a reason to do that.

treeowl (commented):
I forgot to mention the need for QuickCheck properties. At the least, we need one that will verify that the effect order is correct. The simplest thing will probably be to work in `Writer`, recording each key as it's reached.

L0neGamer:
No worries, not requiring a review just yet since I've not done benchmarks. I appreciate the comments, there's oh so much to make sure is done.

containers-tests/tests/intmap-properties.hs
```diff
@@ -1698,6 +1701,17 @@ prop_fromSet ys =
...
+prop_fromSetA_action_order :: [(Int, Int)] -> Bool
+prop_fromSetA_action_order ys =
+  let iSet = IntSet.fromList (fst <$> ys)
+      lookupYs = List.nubBy ((==) `on` fst) ys
+      doLookup k = fromJust $ List.lookup k lookupYs
+      action = \k ->
+        let v = doLookup k
+        in tell [v] $> v
+      xs = IntSet.toList iSet
+  in execWriter (fromSetA action iSet) == List.map doLookup xs
```
treeowl: I think what you've written is mostly fine. That said, the part you inherited from the `fromSet` test is blecherous. It looks like it predates our use of QuickCheck's function generation machinery, or perhaps even that machinery itself. If I'm not mistaken, what we really want is

```haskell
prop_fromSetA_action_order :: [Int] -> Fun Int A -> Property
```

The reason for the `Property` change is so we can use `.&&.` to combine the result test with the effect test, each tagged with an explanation using `counterexample`.

If you could, I'd appreciate if you'd backport the `Fun` bit to the `fromSet` test as well.

L0neGamer: will do! originally just had `[Int]` as input, but wanted to demonstrate I wasn't just putting the keys in order.

L0neGamer: Let me know if https://github.com/haskell/containers/pull/1163/commits/ca4234260b907d2c9f955739a9a41421480310ab needs any changes.

containers-tests/tests/intmap-properties.hs
```diff
@@ -1689,14 +1694,23 @@ prop_FoldableTraversableCompat :: Fun A [B] -> IntMap A -> Property
...
-prop_keysSet xs =
-  keysSet (fromList xs) == IntSet.fromList (List.map fst xs)
-
-prop_fromSet :: [(Int, Int)] -> Bool
-prop_fromSet ys =
-  let xs = List.nubBy ((==) `on` fst) ys
-  in fromSet (\k -> fromJust $ List.lookup k xs) (IntSet.fromList $ List.map fst xs) == fromList xs
+prop_keysSet :: [Int] -> Property
+prop_keysSet keys =
+  keysSet (fromList (fmap (, ()) keys)) === IntSet.fromList keys
```
treeowl, line 1699: I assume this is just to avoid having to generate values that parametricity ensures don't matter anyway? Seems fine.

containers-tests/tests/intmap-properties.hs
```diff
@@ -1689,14 +1694,23 @@ prop_FoldableTraversableCompat :: Fun A [B] -> IntMap A -> Property
...
+
+prop_fromSet :: [Int] -> Fun Int A -> Property
+prop_fromSet keys funF =
+  let f = applyFun funF
```
treeowl:
```suggestion
  let f = apply funF
```

Same but shorter, and I'm pretty sure what we use elsewhere.

containers-tests/tests/intmap-strictness.hs
```diff
@@ -70,11 +71,22 @@ prop_strictFromSet fun set =
   where
     f = coerce (applyFunc fun) :: Key -> A
 
+prop_strictFromSetA :: Func Key (Bot A) -> IntSet -> Property
+prop_strictFromSetA fun set =
+  isBottom (getSolo (M.fromSetA f set)) === any (isBottom . getSolo . f) (IntSet.toList set)
+  where
+    f = MkSolo . coerce (applyFunc fun) :: Key -> Solo A
```
treeowl: What's this coercing from/to?

containers-tests/tests/intmap-properties.hs
L0neGamer: Yeah, no point generating values that aren't relevant

containers-tests/tests/intmap-strictness.hs
L0neGamer: looks like this is a confusing pattern in this and other testing files, where we coerce to and from `Bot A` instead of just using it. I'll remove it in this case, but there are 142 usages of coerce in just intmap-strictness alone so I won't change all of them for now.

treeowl:
> Turns out pre-9.0.2 Solo does not exist, so I'll whip up an equivalent.

Ugh. And then there was a nontrivial amount of churn after its introduction with the (unfortunate, IMO) constructor name change. Feel free to roll your own unconditionally if you don't want to futz with all that CPP.

treeowl:
Wait a sec ... That wasn't very smart of me.... In the test suite we can use whatever packages we like. So we can just depend on `OneTuple` and get `Solo` everywhere. I'm sorry for wasting your time.

L0neGamer:
as you say, CPS writer is overkill. Unless you really want it, I'll switch to lazy writer and call it a day.

treeowl (approved):

containers/src/Data/IntMap/Strict/Internal.hs
```diff
@@ -1056,25 +1062,47 @@ mapEitherWithKey f0 t0 = toPair $ go f0 t0
...
+fromSet f = runIdentity . fromSetA (pure . f)
+#endif
+
+-- | \(O(n)\). Build a map from a set of keys and a function which for each key
+-- computes its value, while within an 'Applicative' context.
+--
+-- This can only be as strict as the 'Applicative' allows it to be.
+--
+-- > fromSetA (\k -> pure $ replicate k 'a') (Data.IntSet.fromList [3, 5]) == pure (fromList [(5,"aaaaa"), (3,"aaa")])
+-- > fromSetA undefined Data.IntSet.empty == pure empty
```
treeowl, line 1077: Please document that 

```haskell
fromSetA f = fmap forceValues . L.fromSetA f
  where
    forceValues xs = foldr (\ !_ r -> r) () xs `seq` xs

fromSetA f = getSolo . L.fromSetA (fmap (MkSolo $!) . f)
```

L0neGamer: I don't know what the second example means

L0neGamer: the MkSolo inside the brackets is going to be the value that is mapped into. did you mean to Compose or similar?

treeowl: That part is right, but the outside should be `fmap getSolo .`. Sorry about that.

L0neGamer: I think it needs to be fmapped twice? Once into the solo, and then into the map

treeowl: No, we don't need to `fmap` into the map. Try this out:

```haskell
myFromSetA f = fmap getSolo . L.fromSetA (fmap (MkSolo $!) . f)
```

If it doesn't work, I'll fire up my laptop and investigate.

treeowl: Wait a second .... I did get mixed up. Yeah, you need a `Compose` or something. Sorry!

L0neGamer: I think it needs to be fmapped twice? Once into the solo, and then into the map

L0neGamer: I'll work out an equivalence and encode it in a property test.

treeowl: The equivalence is

```haskell
fromSetA f = fmap getSolo . getCompose . L.fromSetA (Compose . fmap (MkSolo $!) . f)
```

Apologies again for the errors. I believe I've worked that out before, but I misremembered the details.

L0neGamer: added additional property tests, and added comment on strictness with regards to these examples

containers-tests/tests/intmap-strictness.hs
```diff
@@ -68,12 +70,35 @@ prop_strictFromSet :: Func Key (Bot A) -> IntSet -> Property
...
+prop_fromSetA_equiv_strictness :: Func Int (Bot A) -> IntSet -> Property
+prop_fromSetA_equiv_strictness fun set =
+  -- strict fromSetA is the same as lazy and then forcing
+  bottomOn (M.fromSetA f set) (fmap forceValues (L.fromSetA f set)) .&&.
+  -- strict fromSetA is the same as lazy fromSetA composed with strictly applied
+  -- wrapper
+  bottomOn (M.fromSetA f set) (fmap getSolo . getCompose $ L.fromSetA (Compose . fmap (MkSolo $!) . f) set)
+  where
+  forceValues xs = foldr (\ !_ r -> r) () xs `seq` xs
+  bottomOn = on (===) (isBottom . getSolo)
```
treeowl:
```suggestion
  bottomOn = (===) `on` (isBottom . getSolo)
```

<strike>The absolute objective reality</strike> My very strong opinion is that `on` should only ever be used as an infix operator.

L0neGamer: I am running benchmarks, I'll endeavour to remember this.

treeowl: Thank you for your attention to this most critical matter.

L0neGamer:
image
image


benchmarks are good as far as I can see.

I will bump version number, add to changelog, and adjust `on` tomorrow.

containers-tests/benchmarks/IntMap.hs
```diff
@@ -80,6 +81,9 @@ main = do
             whnf (\n -> M.fromAscList (unitValues [1..n])) bound
         , bench "minView" $ whnf (maybe 0 (\((k,v), m) -> k+v+M.size m) . M.minViewWithKey)
                     (M.fromList $ zip [1..10] [1..10])
+        , bench "fromSet" $ whnf (M.fromSet pred) elems_set
+        , bench "Lazy.fromSetA" $ whnf (M.fromSetA (MkSolo . pred)) elems_set
```
treeowl: This doesn't actually build the result map. You need to force the contents of the resulting `Solo`.

L0neGamer: Ah, assumed whnf did what nf does. A task for tomorrow then.

L0neGamer:
Do you have a suggestion for "larger test cases", or is it just that the structure is not being forced?

L0neGamer:
image
image
here I increase the upper bound to 2^14 from 2^12. I do like your suggestion of additional benchmarks, which will be very easy to add.

treeowl:
I would be more comfortable if I saw a few hundred milliseconds rather than microseconds. Is that feasible? Also, you're only using an `IntSet` consisting of a contiguous block of numbers. There's also a pseudo-random list in the test suite; could you possibly use that as well (or instead)?

L0neGamer:
I can easily just bump the bound size up arbitrarily. I'm worried about bumping the benchmarks above what is reasonable for people to run normally. I'll change the sets used to be the random ones, as well.

L0neGamer:
When I was running all of them it was taking a while, it takes less time when I'm just running these. Increaasing the test cases for `fromSet` and co to hundreds of milliseconds would be increasing the lengtth of the tests by three orders of magnitude, however. My computer is also relatively new so I'd worry for other's setups, or if we were to set these benchmarks up in future.

L0neGamer:
image
Here are some larger benchmarks with randomised sets.

containers-tests/benchmarks/Map.hs
```diff
@@ -136,12 +147,13 @@ main = do
...
         ]
   where
-    bound = 2^12
+    bound = 2^14
     elems = shuffle elems_distinct_asc
     elems_even = zip evens evens
     elems_odd = zip odds odds
     elems_distinct_asc = zip keys_distinct_asc values
     elems_distinct_desc = reverse elems_distinct_asc
+    elems_set = Set.fromDistinctAscList keys_distinct_asc
```
treeowl: This doesn't seem to be used anymore.

containers-tests/benchmarks/IntMap.hs
```diff
@@ -121,9 +127,10 @@ main = do
         , take 63 (map negate (iterate (*2) 1))
         , interleave [1..] (map negate [1..])
         ]
+    elems_set = S.fromDistinctAscList keys
```
treeowl: This doesn't seem to be used anymore.

bench.sh
treeowl, line 1: You're welcome to add this script, but please add a comment to the top explaining what it does and add a note in the `CONTRIBUTING` file to explain when to use it.

L0neGamer: I've commented in the script and in contributing.

treeowl:
I'll merge once the final CI run completes, unless @meooow25 has any objections.

containers-tests/benchmarks/IntMap.hs
```diff
@@ -123,7 +129,7 @@ main = do
         ]
 
     --------------------------------------------------------
-    !bound = 2^12
+    !bound = 2^14
```
meooow25, line 132: Did increasing this demonstrate anything new?  
If not, this just makes *everything* slower for no reason.

containers-tests/benchmarks/Map.hs
```diff
@@ -124,6 +130,11 @@ main = do
         , bench "fromDistinctAscList:fusion" $ whnf (\n -> M.fromDistinctAscList [(i,i) | i <- [1..n]]) bound
         , bench "fromDistinctDescList" $ whnf M.fromDistinctDescList elems_distinct_desc
         , bench "fromDistinctDescList:fusion" $ whnf (\n -> M.fromDistinctDescList [(i,i) | i <- [n,n-1..1]]) bound
+        , bench "fromSet" $ whnf (M.fromSet pred) s_random
+        , bench "Lazy.fromSetA outer" $ whnf (M.fromSetA (MkSolo . pred)) s_random
+        , bench "Strict.fromSetA outer" $ whnf (MS.fromSetA (MkSolo . pred)) s_random
+        , bench "Lazy.fromSetA inner" $ whnf (getSolo . M.fromSetA (MkSolo . pred)) s_random
+        , bench "Strict.fromSetA inner" $ whnf (getSolo . MS.fromSetA (MkSolo . pred)) s_random
```
meooow25, line 137: Why do we have both "outer" and "inner" benchmarks? I wouldn't expect any difference for `Solo`.

containers-tests/tests/Utils/Strictness.hs
```diff
@@ -27,6 +27,12 @@ instance Arbitrary a => Arbitrary (Bot a) where
     , (4, Bot <$> arbitrary)
     ]
 
+instance CoArbitrary a => CoArbitrary (Bot a) where
+  coarbitrary (Bot x) = coarbitrary x
+
+instance Function a => Function (Bot a) where
+  function = functionMap (\(Bot x) -> x) Bot
+
```
meooow25, line 35: These shouldn't be necessary.
In fact if we need them, we're doing something odd. There is no reason to generate functions that take in `Bot a`.

containers-tests/tests/intmap-properties.hs
```diff
@@ -1689,14 +1694,23 @@ prop_FoldableTraversableCompat :: Fun A [B] -> IntMap A -> Property
...
+  in fromSet f (IntSet.fromList keys) === fromList (fmap (id &&& f) keys)
+
+prop_fromSetA_action_order :: [Int] -> Fun Int A -> Property
+prop_fromSetA_action_order keys funF =
+  let iSet = IntSet.fromList keys
+      f = apply funF
+      action = \k ->
+        let v = f k
+        in tell [v] $> v
+  in execWriter (fromSetA action iSet) === List.map f (IntSet.toList iSet)
```
meooow25, line 1713: We should also check that the result map is correct.

containers-tests/tests/intmap-strictness.hs
```diff
@@ -68,12 +70,35 @@ prop_strictFromSet :: Func Key (Bot A) -> IntSet -> Property
...
   isBottom (M.fromSet f set) === any (isBottom . f) (IntSet.toList set)
   where
-    f = coerce (applyFunc fun) :: Key -> A
+    f = applyFunc fun
+
+prop_strictFromSetA :: Func Key (Bot A) -> IntSet -> Property
+prop_strictFromSetA fun set =
+  isBottom (getSolo (M.fromSetA (MkSolo . f) set)) === any (isBottom . f) (IntSet.toList set)
+  where
+    f = applyFunc fun
```
meooow25, line 79: This is a weird way to do this.

What we want in the test here is a `Key -> Solo A` where `A` can be bottom. That translates to generating an arbitrary `Func Key (Solo (Bot A))`, which can be coerced to `Key -> Solo A`.

containers-tests/tests/map-properties.hs
```diff
@@ -1708,23 +1712,31 @@ prop_bifoldl' ck cv n m = Bifoldable.bifoldl' ck' cv' n m === Foldable.foldl' c'
...
+      action = \k ->
+        let v = f k
+        in tell [v] $> v
+  in execWriter (fromSetA action iSet) === List.map f (Set.toList iSet)
+
+prop_fromArgSet :: [(OrdA, B)] -> Property
 prop_fromArgSet ys =
-  let xs = List.nubBy ((==) `on` fst) ys
-  in fromArgSet (Set.fromList $ List.map (uncurry Arg) xs) == fromList xs
+  fromArgSet (Set.fromList $ List.map (uncurry Arg) ys) === fromList ys
```
meooow25, line 1739: Why touch this test? In general, please don't include changes completely unrelated to the PR's purpose.

containers/src/Data/IntMap/Internal.hs
```diff
@@ -3322,17 +3338,23 @@ fromSet f (IntSet.Tip kx bm) = buildTree f kx bm (IntSet.suffixBitMask + 1)
...
-                (buildTree g (prefix + bits2) (bmask `shiftRL` bits2) bits2)
+              liftA2
+                (Bin (Prefix (prefix .|. bits2)))
+                  (buildTree g prefix bmask bits2)
+                  (buildTree g (prefix + bits2) (bmask `shiftRL` bits2) bits2)
+#if __GLASGOW_HASKELL__
+{-# INLINABLE fromSetA #-}
+#else
+{-# INLINE fromSetA #-}
+#endif
```
meooow25, line 3357: Why `INLINE` on non-GHC?

containers/src/Data/Map/Internal.hs
```diff
@@ -3535,8 +3536,30 @@ argSet (Bin sz kx x l r) = Set.Bin sz (Arg kx x) (argSet l) (argSet r)
...
+--
+-- > fromSetA (\k -> pure $ replicate k 'a') (Data.Set.fromList [3, 5]) == pure (fromList [(5,"aaaaa"), (3,"aaa")])
+-- > fromSetA undefined Data.Set.empty == pure empty
+
+fromSetA :: Applicative f => (k -> f a) -> Set.Set k -> f (Map k a)
+fromSetA _ Set.Tip = pure Tip
+fromSetA f (Set.Bin sz x l r) =
+  flip (Bin sz x)
+    <$> fromSetA f l
+    <*> f x
```
meooow25, line 3556: I would prefer to use `liftA2`, since it may be defined to be more efficient for the Applicative used.

containers-tests/benchmarks/IntMap.hs
treeowl: I don't so much trust stability of super-short tests (microseconds), which we were getting for the functions in question. If necessary, we can limit the size increase to only some of the benchmarks.

containers-tests/benchmarks/Map.hs
treeowl: That depends on how well GHC analyses things, right? It has to know what's already been evaluated to avoid suspending constructor application. But maybe it would be better to do this test in a different functor.

containers-tests/tests/Utils/Strictness.hs
treeowl: I don't know why you say that. It avoids having to unwrap the `Bot`s manually.

containers-tests/benchmarks/Map.hs
L0neGamer: As can be seen in the benchmarks, the `inner` benchmarks take about 3 times as long as the outer ones, so there is a difference here.

containers/src/Data/IntMap/Internal.hs
treeowl: Why would it be different on a different implementation?

containers-tests/tests/intmap-strictness.hs
L0neGamer: `Func Key (Bot A)` and `Func Key (Solo (Bot A))` are equivalent, and the only difference is whether you unwrap the Solo in the RHS or you wrap with Solo on the LHS. I preferred the former because it's clearer what the functions is meant to do - produce a bottom value from a key.

containers-tests/tests/map-properties.hs
L0neGamer: I was asked to fix the weirdness of related property functions by treeowl, so I did so.

containers/src/Data/IntMap/Internal.hs
L0neGamer: I presume that there is a reason this pattern is used 45 other times for other functions; I didn't want to break anything by using INLINABLE when it isn't able to be used.

containers-tests/tests/Utils/Strictness.hs
L0neGamer: I'm happy to remove this theoretically. Note that `Fun` from quickcheck cannot generate lazy functions, so it is a little bad to have this instance theoretically.

containers-tests/tests/intmap-properties.hs
L0neGamer: Agreed.

containers/src/Data/Map/Internal.hs
L0neGamer: Good idea.

L0neGamer (commented):
I'll look into improvements in a follow up PR, and request your approval for that in time @ meooow25

containers-tests/tests/Utils/Strictness.hs
meooow25: > It avoids having to unwrap the `Bot`s manually.

Do you have any example of where we would want this instance?

containers-tests/tests/intmap-strictness.hs
meooow25: It's not about what might seem clearer in this particular case, it's a simple mechanical translation from the type of the function being tested which can be applied to *all* tests here.

containers/src/Data/IntMap/Internal.hs
meooow25: They seem to have originated in https://github.com/haskell/containers/commit/06e91accde415ccdf1a680d9436bf2d83731183b, where INLINABLE was added for specialization on GHC. There is really no reason to do this today. Perhaps we should clean these up to avoid confusion. Anyway, it should be harmless to let it be for now.

containers-tests/benchmarks/IntMap.hs
meooow25: > I don't so much trust stability of super-short tests (microseconds)

I think we've discussed this before, but I don't really see why. In case there's noise, I trust the benchmarking library to take care of it and provide a stable result. Anyway, we can adjust the size later if we need to.

containers-tests/benchmarks/Map.hs
meooow25: Oh you're right. Yes it does depend on GHC, and in this case it doesn't know that the value inside the `Solo` will be evaluated, so it has to build up thunks. I think it's okay to have separate tests then.

PR #1164, opened 2025-11-14
Change example to not be contrived due to a duplicate value

lehins:
I am not sure about others, but having the same value in the list made it really hard to follow the example.

I think it will be slightly clearer if there are no duplicate values, especially considering that they serve as an indicator what happens upon duplicate keys.


lehins:
> Would you mind making the same change in Data.Map.Lazy?

Not at all. I already forgot about existence of the lazy interface :smile:

meooow25:
Thanks!

PR #1165, opened 2025-11-16
fromSetA follow up changes

L0neGamer:
Finally getting to some of the changes requested by @meooow25 on https://github.com/haskell/containers/pull/1163

In this PR we remove some unused instances, and try and share the implementations for some other instances. We also do some cleanup for some of the testing code. We use `liftA3` instead of manually combining three actions (which uses liftA2 in the default implementation), and we also verify that the result of `fromSetA` is correct in the case of the action ordering property test.

### DIFF
--- a/containers-tests/tests/intmap-properties.hs
+++ b/containers-tests/tests/intmap-properties.hs
@@ -1701,7 +1701,10 @@ prop_keysSet keys =
 prop_fromSet :: [Int] -> Fun Int A -> Property
 prop_fromSet keys funF =
   let f = apply funF
-  in fromSet f (IntSet.fromList keys) === fromList (fmap (id &&& f) keys)
+      m = fromSet f (IntSet.fromList keys)
+  in
+    valid m .&&.
+    m === fromList (fmap (id &&& f) keys)
 
 prop_fromSetA_action_order :: [Int] -> Fun Int A -> Property
 prop_fromSetA_action_order keys funF =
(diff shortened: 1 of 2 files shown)


treeowl:
Both `fromSet` and `fromSetA` seem to be missing validity (i.e., data structure invariant) tests. I don't understand your (partial?) collection of `Arbitrary` instances, or why that's in this PR.

L0neGamer:
I would appreciate guidance on what to look for for data structure invariant tests.

I moved those arbitrary instances to a separate module because, at least for the Set one, the instance was needlessly duplicated across modules and had increased complexity for no reason. If there was some shared containers-testing lib, then this module could be compiled once for all testing executables, but that is not the case currently.

I would've moved the Map ones as well but of course the strict and lazy variants would have to have the odd cpp-based substitution, and that was complexity I wanted to avoid.

treeowl:
Understood. The function you're looking for is called `valid`. For largely hysterical raisins, it's actually exposed from `Data.Set`. (In modern Haskell library design, there'd probably be a separate `Unsafe` module for functions that can break the invariants, and we'd probably expose the functions for diagnosing said breakage there.)

meooow25 (changes requested):
Looks good to remove the Bot instances, improve the tests, and use liftA3.

But I also don't think the other Arbitrary changes belong here.  
IMO there is no benefit in touching those instances at all. That little bit of duplication is harmless.

treeowl:
You seem to have done some substantial reworking of `Arbitrary` instances here, none of which are related to the purpose of the PR. Please don't do that. You can propose those changes (with explanations of why, and preferably some more source comments too) in a separate PR.

L0neGamer:
Happy to discuss IsInt on the other PR; one option is that we could use mhs in the testing pipeline as well if we do want to commit to non-ghc testing. My impression was that while the testing is ghc only, we may as well make it more readable.

treeowl:
I definitely don't think that change belongs in this pull request; it's a much better fit for the other one.

meooow25 (approved):
Thanks!

PR #1166, opened 2025-11-26
Simplify arbitrary instances for map and set

L0neGamer:
These are over-complicated for what they need to do.

### DIFF
--- a/containers-tests/tests/map-properties.hs
+++ b/containers-tests/tests/map-properties.hs
@@ -324,6 +324,7 @@ instance (Int ~ k, Arbitrary v) => Arbitrary (Map k v) where
         t <- evalStateT (mkArbMap step sz) start
         if valid t then pure t else error "Test generated invalid tree!")
     where
+      step :: StateT Int Gen Int
       step = do
         i <- get
         diff <- lift $ choose (1, gapRange)
--- a/containers-tests/tests/set-properties.hs
+++ b/containers-tests/tests/set-properties.hs
@@ -221,18 +221,23 @@ instance (Int ~ a) => Arbitrary (Set a) where
 
 data TwoSets = TwoSets (Set Int) (Set Int) deriving (Show)
 
--- We produce two lists from a simple "universe". This instance
+-- | We produce two lists from a simple "universe". This instance
 -- is intended to give good results when the two lists are then
 -- combined with each other; if other elements are used with them,
 -- they may or may not behave particularly well.
+--
+-- The universe is made of ints in multiples of three. Each value is equally
+-- likely to be in the left set, the right set, or both sets.
 instance Arbitrary TwoSets where
   arbitrary = do
     (l, r) <- sized $ \sz0 -> do
       sz <- choose (0, sz0)
       let universe = [0,3..3*(sz - 1)]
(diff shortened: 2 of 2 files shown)


containers-tests/tests/set-properties.hs
```diff
@@ -241,45 +208,45 @@ instance IsInt a => Arbitrary (Set a) where
...
+    where
+    divide2Gen :: [a] -> Gen ([a], [a])
+    divide2Gen [] = pure ([], [])
+    divide2Gen (x : xs) = do
+      mIsFirst <- arbitrary
+      (ls, rs) <- divide2Gen xs
+      pure $ case mIsFirst of
+        Just True -> ((x : ls), rs)
+        Just False -> (ls, (x : rs))
+        Nothing -> ((x : ls), (x : rs))
```
treeowl, line 244: This seems strictly less clear. It also produces more "both" results than the current code, which might be fine.

containers-tests/tests/set-properties.hs
```diff
@@ -241,45 +208,45 @@ instance IsInt a => Arbitrary (Set a) where
...
-    sz <- choose (0, sz0)
-    let universe = [0,3..3*(fromInt sz - 1)]
-    divide2Gen universe
-
 instance Arbitrary TwoSets where
   arbitrary = do
-    TwoLists l r <- arbitrary
+    (l, r) <- sized $ \sz0 -> do
+      sz <- choose (0, sz0)
+      let universe = [0,3..3*(sz - 1)]
```
treeowl, line 232: I know I wrote the original code, but I no longer remember the purpose of using products of 3. Has your work here given you any hints?

containers-tests/tests/set-properties.hs
treeowl: If you don't want another type, you could do something basic with `rem` inline.

containers-tests/tests/set-properties.hs
L0neGamer: Not immediately; I would presume because they're not incredibly densely packed that lets other manipulations happen, but I can't say for certain.

treeowl: Yeah, I think it's probably to get some gaps for testing with values in neither? I can't imagine that's the best way to do it; Please feel free to do something else about that.

containers-tests/tests/set-properties.hs
L0neGamer: I would prefer a more numeric approach, I think that makes more sense than the ternary data types here.

containers-tests/tests/map-properties.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeFamilies #-}
```
meooow25, line 4: As brought up in #1165, we would like to get tests working with MHS, which does not support type families. So this is potentially setting up a future problem.

L0neGamer:
I don't think I'm able to commit to bringing this further usefully, and I don't feel strongly enough about it.

PR #1167, opened 2025-12-25
Drop "ifdef GHC"s on INLINABLE pragmas

meooow25:
These were added long ago in 06e91acc as an improvement on the already existing INLINE pragmas. We can get rid of these unnecessary (and confusing to contributors) ifdefs today since we do not have other compilers where we want the non-INLINABLE behavior to apply.


meooow25:
Motivated by https://github.com/haskell/containers/pull/1163#discussion_r2536067700

treeowl:
Aren't they there because `INLINABLE` isn't a Report standard feature?

treeowl:
They'll very likely throw up a slew of warnings 🫤. That hypothetical problem is probably worth incurring to remove the actual clutter though.

meooow25:
Yes, I also think cleaning up the clutter would be nice. For what it's worth, MHS is the only such compiler at the moment and it's not complaining.

PR #1169, opened 2025-12-27
Add a merge API for Set

meooow25:
For #732

### DIFF
--- a/containers/src/Data/Set/Internal.hs
+++ b/containers/src/Data/Set/Internal.hs
@@ -220,11 +220,27 @@ module Data.Set.Internal (
             , showTreeWith
             , valid
 
+            -- * General merge functions
+            , WhenMissing(..)
+            , SimpleWhenMissing
+            , dropMissing
+            , preserveMissing
+            , filterMissing
+            , filterAMissing
+            , runWhenMissing
+            , WhenMatched(..)
+            , SimpleWhenMatched
+            , filterMatched
+            , filterAMatched
+            , runWhenMatched
+            , merge
+            , mergeA
+
             -- Internals (for testing)
             , bin
             , balanced
             , link
-            , merge
+            , link2
             , balanceL
             , balanceR
             , glue
@@ -235,11 +251,11 @@ module Data.Set.Internal (
 import Utils.Containers.Internal.Prelude hiding
   (filter,foldl,foldl',foldr,null,map,take,drop,splitAt)
 import Prelude ()
-import Control.Applicative (Const(..))
+import Control.Applicative (Const(..), liftA3)
 import qualified Data.List as List
 import Data.Semigroup (Semigroup(..), stimesIdempotentMonoid, stimesIdempotent)
 import Data.Functor.Classes
-import Data.Functor.Identity (Identity)
+import Data.Functor.Identity (Identity(..))
 import qualified Data.Foldable as Foldable
 import Control.DeepSeq (NFData(rnf),NFData1(liftRnf))
 import Data.List.NonEmpty (NonEmpty(..))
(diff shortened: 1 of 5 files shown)


containers/src/Data/Set/Internal.hs
```diff
@@ -2192,6 +2208,271 @@ instance Monoid (MergeSet a) where
...
+filterAMissing :: Applicative f => (a -> f Bool) -> WhenMissing f a
+filterAMissing f = WhenMissing
+  { missingSubtree = filterA f
+  , missingElem = f
+  }
+{-# INLINE filterAMissing #-}
+
+filterA :: Applicative f => (a -> f Bool) -> Set a -> f (Set a)
+filterA f = go
+  where
```
treeowl: Do we want to start with this? I think it should help a little in general and a lot for some functors.

```haskell
go t@(Bin 1 x _ _) = finish <$> f x
  where
    finish False = Tip
    finish True = t
```

The extra sharing here makes me think we probably want to share any unchanged subtrees. We don't even need to use pointer equality for that, though doing so might turn out to be fastest.

meooow25: Thanks, updated to use `ptrEq` as we do in `filter`. I think it makes sense to keep them consistent.

treeowl: Agreed. I still think you should consider special-casing a size of 1. We inherently know its subtrees don't change, plus we avoid the `liftA3`.

meooow25:
I'm not sure I see how the `alterF` idea could be applied for `filterA`. `alterF` tries to eagerly calculate the set we would get if the function happens to change it. This is feasible because there is only one such set. For `filterA`, each element can be kept or dropped, which leads to $2^n-1$ other sets if we try the same thing.

Personally I don't think the `alterF` implementation is a good idea because it always allocates the other set, even if it turns out to be not the result. But this is a little off the current topic.

containers/src/Data/Set/Internal.hs
meooow25: Good point about the `liftA3`, done.

I expect that this is also a little faster because of #1069, but whether we should do this everywhere is still an open question.

containers/src/Data/Set/Internal.hs
```diff
@@ -2150,7 +2186,262 @@ instance Monoid (MergeSet a) where
...
+--
+-- @since FIXME
+merge
+  :: Ord a
+  => SimpleWhenMissing a -- ^ What to do with elements in @s1@ but not @s2@
+  -> SimpleWhenMissing a -- ^ What to do with elements in @s2@ but not @s1@
+  -> SimpleWhenMatched a -- ^ What to do with elements in both @s1@ and @s2@
+  -> Set a -- ^ Set @s1@
+  -> Set a -- ^ Set @s2@
+  -> Set a
```
Vlix, line 2375: Just from a Developer Experience perspective, would the following be a good idea?

```hs
merge
  :: Ord a
  => (Set a, SimpleWhenMissing a)
     -- ^ @s1@ and what to do with elements that are not in @s2@
  -> (Set a, SimpleWhenMissing a)
     -- ^ @s2@ and what to do with elements that are not in @s1@
  -> SimpleWhenMatched a -- ^ What to do with elements in both @s1@ and @s2@
  -> Set a
```

meooow25: Worth considering, but I would say the current form is preferable.
(For the record, this interface is the same as in the existing Map.Merge.)

The idea is that this lets users define their merge function as

```hs
myMerge :: Ord a => Set a -> Set a -> Set a
myMerge = merge (... whenMissing1 ...) (... whenMissing2 ...) (... whenMatched ...)
```

which they can use where they need it. The doc also refers to this:

```
-- When 'mergeA' is given three arguments, it is inlined at the call
-- site. To prevent excessive inlining, you should generally only use
-- 'mergeA' to define custom combining functions.
```

PR #1170, opened 2025-12-28
Define a common StrictTriple for internal use

meooow25:
* Move StrictPair to Utils.Containers.Internal.Strict next to StrictTriple
* Remove the currently unused strict Maybe


containers/src/Data/IntMap/Internal.hs
```diff
@@ -2974,14 +2975,14 @@ split k t =
     go _ Nil = (Nil :*: Nil)
 
 
-data SplitLookup a = SplitLookup !(IntMap a) !(Maybe a) !(IntMap a)
+type SplitLookup a = StrictTriple (IntMap a) (Maybe a) (IntMap a)
```
treeowl, line 2978: Have you tried stitching to an unlifted maybe here? I don't remember if I have.

meooow25: Would that make a difference compared to simply being strict?

But switching to an unboxed/unpacked Maybe would avoid allocating the Justs, we could try that later.

treeowl: It's just one allocation, so probably not a big deal, but yeah, I think it'd be worth trying.

PR #1172, opened 2026-01-01
Use StrictTriple for splitMember

meooow25:
This tracks the strictness in the implementation more cleanly and makes it easier to optimize. For instance, with GHC 9.12.2, the Set intersection benchmark time improves by 13%.

---

```
Name                       Time - - - - - - - -    Allocated - - - - -
intersection                58 μs   50 μs  -13%    321 KB  322 KB   +0%
```


treeowl (approved):
Nice catch.

treeowl:
That might be worth a GHC ticket, especially if the problem survives code generation.

PR #1173, opened 2026-01-03
Update CI to test with GHC 9.14

meooow25:


meooow25:
Failures are due to https://github.com/haskell-CI/haskell-ci/pull/789

IMO it's not worth trying to fix the warning for containers-tests. It defines some dependency groups which seem harmless and it would be a pain to inline them into each test suite.

https://github.com/haskell/containers/blob/32ed12aadf34efa14d1e1b055dea87980bf15f7d/containers-tests/containers-tests.cabal#L36-L63

treeowl:
Looking more closely, I see that most of your pattern match fixes are substantial clarity improvements as well. I just happened to see the least valuable one first.

PR #1176, opened 2026-01-11
Refactor containers h includes

### LINKED ISSUE #968: Unnecessary CPP and C header in `Data.Map.Internal.Debug.html`?
I see `containers.h` being `#include`d in almost every module, which is probably the point, but is this just a specific choice: "include it everywhere, if we ever need it it'll be in scope and if we don't the compiler will just ignore it"? Or is this an oversight and is it better to limit the amount of CPP pragmas and inclusion of C headers?

RikuMinamiyama:
**Overview**
Removed unnecessary CPP and C header includes.

**Correction details**

- C header removal

   The following modules use values defined in `containers.h`:
   ```
   /haskell-containers/containers/src/Data/Graph.hs
   → DEFINE_PATTERN_SYNONYMS, USE_ST_MONAD, USE_UNBOXED_ARRAYS

   /haskell-containers/containers/src/Data/Sequence.hs
   → DEFINE_PATTERN_SYNONYMS

   /haskell-containers/containers/src/Data/IntSet/Internal.hs
   → WORD_SIZE_IN_BITS

   /haskell-containers/containers/src/Data/Map/Internal.hs
   → WORD_SIZE_IN_BITS

   /haskell-containers/containers/src/Data/Sequence/Internal.hs
   → DEFINE_PATTERN_SYNONYMS

   /haskell-containers/containers-tests/tests/IntSetValidity.hs
   → WORD_SIZE_IN_BITS
   ```
   Removed `#include "containers.h"` from all other modules.

- CPP removal

   Removed CPP language pragmas from files that do not use CPP directives.

**Testing**

- `cabal build containers` succeeds
- `cabal test containers-tests` passes all tests
- `cabal bench containers-tests` passes all benchmarks

While the changes are straightforward, they affect a wide range of files, so careful review is requested.

Fixes #968


RikuMinamiyama:
Thank you for your confirmation.

Regarding `containers.h` , I searched for values defined in `containers.h` , such as `DEFINE_PATTERN_SYNONYMS` , and removed the `containers.h` include from files where these values were absent. (LLMs were not used)

Regarding C++ pragmas, I first checked each file for the presence of `#ifdef` and similar directives. (No LLM was used.) Afterward, out of concern, I asked the LLM if there were any places where C++ pragma declarations might have been accidentally left in.

Additionally, when drafting the PR description, I used an LLM to translate from Japanese to English.

If there are any errors or omissions, please provide guidance.

meooow25 (approved):
I think `map-strictness.hs` could also do without the `{-# LANGUAGE CPP #-}`, would you please add that to the PR?
Also, it needs a rebase to avoid conflicts.

meooow25:
Closing in favor of #1213.

PR #1180, opened 2026-02-14
Run CI on a big-endian system

meooow25:


treeowl:
How did you write/generate this? How can someone else update it? It would be good to record this info in the repo.

meooow25:
The caching of the install part seems to not be working. The error message is
```
error from registry: installation not allowed to Create organization package
```

I can see that it works fine for `text` and `bytestring`, e.g. https://github.com/haskell/bytestring/packages

It might be something to do with workflow permissions for this repository which I don't have access to. @treeowl maybe you can do something about this?

.github/workflows/s390x.yml
treeowl, line 1: Where does all this come from, and what maintenance will it need? I still don't see documentation on that.

meooow25: I'm not sure what you wish to be documented. It's configuration for a GitHub Action.

meooow25: @treeowl? As mentioned above, I wrote this and it uses https://github.com/uraimo/run-on-arch-action which emulates s390x. Is there anything that you wish to document?

meooow25: I'll merge it for now.

PR #1185, opened 2026-03-15
Add whenMissing, which takes user-supplied fns

### LINKED ISSUE #1054: Can we make the Map.Merge API more expressive?
I wanted to demonstrate `partitionKeys` recently (https://github.com/haskell/containers/pull/975#issuecomment-2417976839) and realized that the public Map.Merge API is not expressive enough for it.

What I need:

```hs
wm1 :: WhenMissing Pair k a a
wm1 = WhenMissing (\t -> Pair empty t) (\_ x -> Pair Nothing (Just x))
```

Best I can do with the public API:

```hs
wm1 :: WhenMissing Pair k a a
wm1 = traverseMaybeMissing (\_ x -> Pair Nothing (Just x))
```

which is terribly inefficient! (O(1) vs O(n))

Is there a safe way to allow such use cases?

meooow25:
This allows for some useful applications which are currently not possible with the non-internal API. The functions carry a warning that the two input functions are related by a law and the user is responsible for satisfying it.

Export filterA for Set, now that it is referred to from whenMissing.

Closes #1054

### DIFF
--- a/containers-tests/tests/set-properties.hs
+++ b/containers-tests/tests/set-properties.hs
@@ -104,6 +104,7 @@ main = defaultMain $ testGroup "set-properties"
                    , testProperty "prop_splitRoot" prop_splitRoot
                    , testProperty "prop_partition" prop_partition
                    , testProperty "prop_filter" prop_filter
+                   , testProperty "prop_filterA" prop_filterA
                    , testProperty "prop_mapMaybe" prop_mapMaybe
                    , testProperty "takeWhileAntitone"    prop_takeWhileAntitone
                    , testProperty "dropWhileAntitone"    prop_dropWhileAntitone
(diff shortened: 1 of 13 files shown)


meooow25:
I decided to go with the simple approach from #1054.

~~Just a note that I haven't added this function to the `Merge.Strict` modules, since I cannot think of a situation where we would also need to force the values.~~

meooow25:
@treeowl did you get a chance to look at the examples?

meooow25:
Rebased and updated.

@treeowl would you please review this? I would really like to include it in the next release.

meooow25:
@Lysxia may I ask you for a review? I would like to get another pair of eyes on this.

meooow25:
Alright, I'm merging it.

Issue #758, opened 2020-12-16
from*DescList* missing from IntMap


hseg:
`fromDescList`, `fromDescListWith`, `fromDescListWithKey`, `fromDistinctDescList` are all missing from `IntMap`.

treeowl:
Good point. Do you want to try writing them?

hseg:
A cursory glance reveals some structural differences between `Map.fromAscList` and `IntMap.fromAscList`, with the former basically doing `fromList . groupWith f` and marking everything in sight as `INLINABLE` to get the relevant fusion, whereas the latter has a `fromMonoListWithKey`  which seems to be a fused version of the above with a flag controlling whether the user asserts the input is distinct, and for some reason is marked everywhere as `NOINLINE`.
Probably the sane thing to do is clone the `IntMap` code, but the inconsistency is jarring enough to make me second-guess myself. That, and `fromMonoListWithKey` is complex enough to make me doubt I can easily flip it.

treeowl:
`Map` needs inlining to specialize to an `Ord` instance. `IntMap` gets much less, if any, benefit from large-scale inlining, so code size is prioritized. I think a lot of the other differences are about differences in the data structure shapes. Anyway, I can give it a shot.

meooow25:
PRs welcome.

The implementation of `fromAscList*` is simpler now after #1123, and can be a template for `fromDescList*`.

PR #1194, opened 2026-04-06
Add fromDescList for IntSet and IntMap

meooow25:
For #758.


meooow25:
I hesitate to add `fromDescListWith` since we know now that the `a -> a -> a` isn't the clearest interface and mistakes can lead to performance issues (#961).

It might be better to add `fromAscListUpsert` and `fromDescListUpsert` once #1190 is in.

PR #1196, opened 2026-04-08
Define the CI config manually

### LINKED ISSUE #1195: Drop haskell-ci?
haskell-ci has been convenient to use, but I am wondering if it would be better to just configure the CI ourselves. It isn't too complicated (see #1183 as a simple example).

Some reasons why this would be good:

* We won't need a [config file](https://github.com/haskell/containers/blob/17908ce984e1a227b691183c98cc8594da21a14f/cabal.haskell-ci) to generate another config file.
* We can merge #886. It doesn't work today since haskell-ci packs an sdist, unpacks it, then runs the tests, and that's a problem if there are relative paths going out of the current directory.
* No surprising CI failures, e.g. because haskell-ci decided to add a Werror (https://github.com/haskell-CI/haskell-ci/pull/789, https://github.com/haskell-CI/haskell-ci/pull/790)
* We can set `--quickcheck-tests` (https://github.com/haskell-CI/haskell-ci/issues/743)
* Maybe faster CI times, since e.g. we don't need to build Haddocks for `containers-tests`.

meooow25:
Configure the CI directly without haskell-ci. This greatly simplifies the config and skips unnecessary steps.

Also,
* Set --quickcheck-tests to 1000 for more thorough testing (default is 100).
* Test on windows-latest and macos-latest

Closes #1195


meooow25:
@treeowl any comments on the config?

PR #1198, opened 2026-04-11
Depend on OneTuple conditionally

meooow25:
It's a compat package, so we need it only for older GHCs.


meooow25:
Oh, they put it in `Data.Tuple.Solo` instead of `Data.Tuple`...

PR #1212, opened 2026-05-16
Use `SPECIALIZE` consistently

sol:
- consistent terminology is important (e.g. if you write a paper you don't want variation in your terminology)
- most people are not native English speakers, having everybody deal with two dialects of English just does not check out in terms of fairness
- it's also an issue for assistive technologies / search / `grep`


meooow25 (approved):

meooow25:
Thanks!

Issue #1209, opened 2026-05-15
Take a look at the alterF implementations


meooow25:
Check if we can improve/simplify some things.

For instance, I don't really like that `Set`'s `alterF` always allocates the changed set (previously mentioned in https://github.com/haskell/containers/pull/1169#issuecomment-3694319872). I would like to try changing it to the bitqueue approach from Map's alterF.

meooow25:
I decided to benchmark all the `Set` `alterF` implementations I can think of: https://gist.github.com/meooow25/e919ae830fb429bab817e57d5f953e7d

To briefly describe them, they are

1. The current implementation ("current")
2. `member`, then `insert` or `delete` ("twopass")
3. Just like `alter` but with `fmap` at each step ("simple")
4. Trace the path, and use it to modify the tree if needed ("bitqueue", "index")
5. Go down to the target while keeping track of how to reconstruct the tree ("cont", "stackstrict", "stacklazy")

Unsurprisingly, the current implementation performs poorly when the set does not need to be modified.

I think the `Map` `alterF` strategy (4) is a good selection of tradeoffs. It may traverse the tree twice, but it minimizes everything else: the number of comparisons, the number of allocations, and the number of fmaps. Interestingly, in this strategy "index" seems to be a little faster than "bitqueue" while also being simpler to implement. I plan to replace the `Set` implementation with this, if I don't find anything better.

meooow25:
While looking around I spotted that `Data.Map.alterF` is too strict.

```
> import Data.IntMap as IntMap
> import Data.IntSet as IntSet
> import Data.Map as Map
> import Data.Set as Set
> import Data.Proxy
> IntMap.alterF (\_ -> Proxy) undefined undefined `seq` ()
()
> IntSet.alterF (\_ -> Proxy) undefined undefined `seq` ()
()
> Set.alterF (\_ -> Proxy) undefined undefined `seq` ()
()
> Map.alterF (\_ -> Proxy) undefined undefined `seq` ()
*** Exception: Prelude.undefined
```

meooow25:
I think this is good enough for now.

I am wondering whether we should also change Map to use an index instead of bitqueue. Judging by the Set benchmarks the performance will be comparable, so the only benefit will be simplifying the implementation and deleting a bunch of code. I will open a new issue if I try it later.

PR #1215, opened 2026-05-31
Use a different strategy for Set.alterF

meooow25:
This is an improvement over the previous implementation in many (but not
all) scenarios. Most importantly it avoids allocating a Set when the Set
doesn't need to be changed. See Note [alterF implementation] for more
about the choice of implementation.

alterF also gets an INLINE since it's a small definition and inlining
permits further GHC optimizations based on the altering function.

For #1209.

### DIFF
--- a/containers-tests/benchmarks/Set.hs
+++ b/containers-tests/benchmarks/Set.hs
@@ -15,10 +15,11 @@ main = do
     let s = S.fromList elems :: S.Set Int
         s_even = S.fromList elems_even :: S.Set Int
         s_odd = S.fromList elems_odd :: S.Set Int
-        strings_s = S.fromList strings
+        strings_s = S.fromList strings :: S.Set String
     evaluate $ rnf [s, s_even, s_odd]
     evaluate $ rnf
       [elems_distinct_asc, elems_distinct_desc, elems_asc, elems_desc]
+    evaluate $ rnf strings_s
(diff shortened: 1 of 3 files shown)


meooow25:
Benchmarks with GHC 9.14:

```
Name                     Time - - - - - - - -    Allocated - - - - -
alterF:delete:absent     182 μs   26 μs  -85%    1.5 MB    3 B   -99%
alterF:delete:present    133 μs  185 μs  +39%    1.0 MB  740 KB  -27%
alterF:four              513 μs  629 μs  +22%    3.1 MB  1.8 MB  -40%
alterF:four:strings      819 μs  886 μs   +8%    3.1 MB  1.8 MB  -40%
alterF:insert:absent     195 μs  259 μs  +33%    1.4 MB  1.0 MB  -29%
alterF:insert:present    174 μs   19 μs  -88%    1.2 MB    7 B   -99%
alterF:member            479 μs  176 μs  -63%    2.5 MB   46 B   -99%
```

sjakobi (approved):
I like it! :)

Some haddocks on the new functions and `MemberIndex` wouldn't hurt, and you could consider exposing them. Or wait until there's explicit demand.

meooow25:
Good point about adding some docs, done.

About exposing these, I would prefer to wait until someone asks for it.

PR #1219, opened 2026-06-13
Avoid some allocations with Map.alterF

meooow25:
Inline the function so that GHC can optimize it based on the altering function. Also unpack the Maybe in TraceResult. Both of these result in less allocations.

For #1209.

### DIFF
--- a/containers/src/Data/Map/Internal.hs
+++ b/containers/src/Data/Map/Internal.hs
@@ -1201,7 +1201,7 @@ alterF :: (Functor f, Ord k)
 alterF f k m = atKeyImpl Lazy k f m
 
 #ifdef __GLASGOW_HASKELL__
-{-# INLINABLE [2] alterF #-}
+{-# INLINE [2] alterF #-}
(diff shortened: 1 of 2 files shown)


meooow25:
Benchmarks with GHC 9.14:

```
Name                              Time - - - - - - - -    Allocated - - - - -
alterF no rules delete absent     517 μs  502 μs   -2%    280 B   123 B   -56%
alterF no rules delete present    2.0 ms  1.9 ms   +0%    5.1 MB  5.0 MB   -2%
alterF no rules insert absent     1.8 ms  1.8 ms   +0%    5.8 MB  5.5 MB   -4%
alterF no rules insert present    1.5 ms  1.4 ms   -5%    5.1 MB  4.7 MB   -7%
alterF no rules lookup absent     490 μs  514 μs   +4%    155 B   141 B    -9%
alterF no rules lookup present    521 μs  520 μs   +0%    128 KB  140 B   -99%
```

PR #1226, opened 2026-07-06
Add filterTree to Data.Tree

tchoutri:
@meooow25 Hi, I'd like to upstream a filter function we have at work for Trees. Happy to include a property test if you have something that you'd like tested specifically.

### DIFF
--- a/containers/src/Data/Tree.hs
+++ b/containers/src/Data/Tree.hs
@@ -43,6 +43,7 @@ module Data.Tree(
     -- * Elimination
     , foldTree
     , flatten
+    , filterTree
     , levels
     , leaves
     , edges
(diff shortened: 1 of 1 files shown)


meooow25:
Hi, thanks for the PR. However there are multiple possible implementations for a "filterTree". As an example, a different "treefilter" was proposed in #39. Hoogle shows yet another possible implementation used in hedgehog [here](https://hackage-content.haskell.org/package/hedgehog-1.7/docs/Hedgehog-Internal-Tree.html#v:filter). So I am not convinced that this one is particular deserves to be included in Data.Tree.

tchoutri:
@meooow25 yeah of course, you should benchmark the possible implementations, I just had to write one for work and I thought I'd upstream it, since there wasn't one available in Data.Tree. :)

meooow25:
I don't mean that these implementation differ in efficiency (which is also something to consider, of course), but that they have different behaviors.

PR #1227, opened 2026-07-13
Add an API to merge a Set and Map into a Map

### LINKED ISSUE #1208: Generalized API to merge Map and Set into a Map
For previous discussion see #975.

I'm working on the implementation.

meooow25:
This is intended to serve some use cases more efficiently than existing methods, such as partitioning a map according to whether the keys exist in a set.

Similar to Map merging, there are two modules, "Data.Map.Merge.Set.Lazy" and "Data.Map.Merge.Set.Strict" exposing the merge functions and tactics. The WhenMissing tactic for missing maps is simply reused from the Map merging interface. A new WhenMatched tactic is added. A new WhenMissingSet tactic is also added, and I chose to added the "Set" suffix to distinguish it from the existing map tactic. Similarly, the functions that create WhenMissingSet tactics are suffixed with "MissingSet".

Closes #1208

### DIFF
--- a/containers-tests/containers-tests.cabal
+++ b/containers-tests/containers-tests.cabal
@@ -117,6 +117,9 @@ library
     Data.Map.Lazy
     Data.Map.Merge.Lazy
     Data.Map.Merge.Strict
+    Data.Map.Merge.Set.Internal
+    Data.Map.Merge.Set.Lazy
+    Data.Map.Merge.Set.Strict
     Data.Map.Strict
     Data.Map.Strict.Internal
     Data.Sequence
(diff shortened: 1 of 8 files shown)


sjakobi:
I think this looks pretty neat, but I think there should be an illustrating example somewhere. E.g. something like your `partitionKeys` from https://github.com/haskell/containers/pull/975#issuecomment-2417976839.

meooow25:
> I think this looks pretty neat, but I think there should be an illustrating example somewhere. E.g. something like your `partitionKeys` from [#975 (comment)](https://github.com/haskell/containers/pull/975#issuecomment-2417976839).

Thanks for taking a look!

I've expanded the docs to be consistent with `Map`'s `merge` and `mergeA`. While the `partitionKeys` example is possible with this change, I'm reluctant to add it since it requires using the internal module. I'm hoping to address that with #1185.

Issue #1205, opened 2026-04-25
Add tryInsert for sets and maps


meooow25:
```hs
tryInsert :: Ord k => k -> a -> Map k a -> Maybe (Map k a)
tryInsert :: Ord a => a -> Set a -> Maybe (Set a)
tryInsert :: Int -> a -> IntMap a -> Maybe (IntMap a)
tryInsert :: Int -> IntSet -> Maybe IntSet
```

Inserts the value if the key is not in the map/set, otherwise returns `Nothing`.

This will be useful in `nubOrdOn` to avoid `member`+`insert`:

https://github.com/haskell/containers/blob/d82a404141537c97df421d589a946a09272cc6d6/containers/src/Data/Containers/ListUtils.hs#L86-L88

Alternate name suggestions are welcome.

If we go with `tryInsert` I think we should rename `pop` from #1134 to `tryDelete` for consistency. I didn't think of this name at that time. Luckily we haven't released it yet, so we can still change the name.

meooow25:
Thinking more about this, it seems more appropriate for the map functions to be

```hs
tryInsert :: Ord k => k -> Map k a -> Maybe (a -> Map k a)
tryInsert :: Int -> IntMap a -> Maybe (a -> IntMap a)
```

This avoids creating an unnecessary thunk for the value, it needs to be supplied only if the key is absent.

alexfmpe:
> Alternate name suggestions are welcome.

I'd say `tryInsert` is ambiguous. I was going to suggest `replace` until I noticed it's actually the other way around, so how about `insertIfMissing` ?

> This avoids creating an unnecessary thunk for the value, it needs to be supplied only if the key is absent.

Also, it's more primitive as the original signature can be defined in terms of it 
`tryInsert' k a m = fmap ($ a) $ tryInsert k m`
but AFAICT not the other way around - where would we get the `a` from?

meooow25:
`insertIfMissing` doesn't seem bad, maybe a bit longer than ideal.

Do you have a name suggestion for `pop` too?

```hs
pop :: Ord k => k -> Map k a -> Maybe (a, Map k a)
pop :: Ord a => a -> Set a -> Maybe (Set a)
```

It would be nice if the names can reflect that they are inverses somewhat.  
`deleteIfPresent` seems like an obvious candidate, though it doesn't indicate that in case of the map the result contains the value.

treeowl:
[...]

PR #1228, opened 2026-07-20
Use Set.alterF in nubOrd

meooow25:
Compared to the previous implementation of `member` followed by `insert`, `alterF` uses `memberIndex` and `insertAt`. This avoids all the element comparisons in `insert`, at the cost of `memberIndex` being a little more expensive than `member`. I expect this change to be an improvement in typical use cases.

Related: #1205

---

Benchmarks with GHC 9.14:

```
Name                                Time - - - - - - - -    Allocated - - - - -
nubOrd.issue1202_distinct            48 μs   46 μs   -4%    672 KB  672 KB   +0%
nubOrd.issue1202_repeat             3.8 μs  4.0 μs   +3%     16 KB   16 KB   +0%
nubOrd.noFusion_distinct             44 μs   43 μs   -1%    633 KB  633 KB   +0%
nubOrd.noFusion_repeat              1.5 μs  1.9 μs  +19%    152 B   152 B    +0%
nubOrd.strings_noFusion_distinct    166 μs  112 μs  -32%    611 KB  626 KB   +2%
nubOrd.strings_noFusion_repeat      3.6 μs  4.2 μs  +16%    136 B   152 B   +11%
```

### DIFF
--- a/containers-tests/benchmarks/ListUtils.hs
+++ b/containers-tests/benchmarks/ListUtils.hs
@@ -10,21 +10,26 @@ main :: IO ()
 main = do
   evaluate $ rnf [xs_distinct, xs_repeat]
   evaluate $ rnf [xss_distinct, xss_repeat]
+  evaluate $ rnf [strings_distinct, strings_repeat]
   defaultMain
     [ bgroup "nubOrd"
-      [ bench "no_fusion_distinct" $
+      [ bench "noFusion_distinct" $
           whnf (consumeNoFusion . LU.nubOrd) xs_distinct
-      , bench "no_fusion_repeat" $
+      , bench "noFusion_repeat" $
           whnf (consumeNoFusion . LU.nubOrd) xs_repeat
       , bench "issue1202_distinct" $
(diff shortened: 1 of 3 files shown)


sjakobi (commented):
I think it's very good to reduce the number of `Ord` comparisons, but it would be better to if the new implementation didn't allocate all these `Maybe`s.

meooow25:
Ah but it doesn't, the `Maybe`s disappear after optimizations. You'll notice that the allocations numbers haven't increased (except for `nubOrd.strings_noFusion_repeat`, I haven't checked what's up with that but the Core looks good so I am satisfied).

sjakobi:
> Ah but it doesn't, the `Maybe`s disappear after optimizations. You'll notice that the allocations numbers haven't increased (except for `nubOrd.strings_noFusion_repeat`, I haven't checked what's up with that but the Core looks good so I am satisfied).

Ah, I missed that! And it appears that you managed to reduce the allocations in #1229! :+1:

PR #1234, opened 2026-08-10
Add fromSetMaybe and fromSetMaybeA for maps

### LINKED ISSUE #1232: Add Map.fromSetMaybe and Map.fromSetMaybeA
Consider adding

```hs
fromSetMaybe :: (k -> Maybe a) -> Set k -> Map k a
fromSetMaybeA :: Applicative f => (k -> f (Maybe a)) -> Set k -> f (Map k a)
```

This looks like a small piece that is missing and would work well with #1227 for something like

```hs
generateMaybeMissingSet :: Applicative f => (k -> Maybe a) -> WhenMissingSet f k a
```

meooow25:
These are like the existing fromSet and fromSetA, but allow dropping
some keys from the input set. These functions will also be used by the
in-progress Map-Set merge API.

Also adjust fromSet and fromSetA docs to make it consistent with the new
functions.

Closes #1232

### DIFF
--- a/containers-tests/tests/intmap-properties.hs
+++ b/containers-tests/tests/intmap-properties.hs
@@ -37,6 +37,7 @@ import qualified Prelude (map, filter)
 import Data.List (nub,sort)
 import qualified Data.List as List
 import qualified Data.List.NonEmpty as NE
+import Data.IntSet (IntSet)
 import qualified Data.IntSet as IntSet
 import Test.Tasty
 import Test.Tasty.HUnit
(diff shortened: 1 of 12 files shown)


meooow25:
Pulling `g` out of `buildTree` and inlining it seems to have improved the performance of `fromSetA`. I made the change to only share the code with `fromSetMaybe`, so this is a nice surprise.

```
Name                     Time - - - - - - - -    Allocated - - - - -
Lazy.fromSetA inner      863 μs  753 μs  -12%    2.4 MB  1.9 MB  -21%
Lazy.fromSetA outer      678 μs  578 μs  -14%    1.9 MB  1.4 MB  -26%
Strict.fromSetA inner    1.1 ms  774 μs  -26%    3.1 MB  2.1 MB  -31%
Strict.fromSetA outer    706 μs  477 μs  -32%    2.0 MB  1.0 MB  -49%
fromSet                  555 μs  560 μs   +0%    1.4 MB  1.4 MB   +0%
```

PR #1235, opened 2026-08-11
Explain Int{Map,Set} prefixes with examples

### LINKED ISSUE #1221: Better documentation on `IntMap`'s improved `Prefix` structure?
The `Note [Okasaki-Gill]` describes the differences to the version described in the paper, but since there are several, I feel it would be helpful if the `Prefix` structure would be described directly and ideally with an example.

meooow25:
Closes #1221

### DIFF
--- a/containers/src/Data/IntSet/Internal.hs
+++ b/containers/src/Data/IntSet/Internal.hs
@@ -300,14 +300,43 @@ type BitMap = Word
 --
 -- * In the context of a Tip, the highest (WORD_SIZE - lg(WORD_SIZE)) bits of
 --   a key are called "prefix" and the lowest lg(WORD_SIZE) bits are called
---   "suffix". In Tip kx bm, kx is the shared prefix and bm is a bitmask of the
---   suffixes of the keys. In other words, the keys of Tip kx bm are (kx .|. i)
---   for every set bit i in bm.
---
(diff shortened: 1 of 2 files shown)


containers/src/Data/IntSet/Internal.hs
```diff
@@ -300,14 +300,43 @@ type BitMap = Word
...
+--
+--
+-- Now consider that we have a (Tip kx bm) with kx and bm as
+--
+-- 0b00000000001100010010010001000000  0b00000000000100000100000000000100
+--                                                  ^     ^           ^
+--                                                 20    14           2  -- suffixes
+--   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ shared prefix
+--
+-- This Tip has 3 keys, which can recovered by bitwise-or-ing the prefix and
```
sjakobi:
```suggestion
-- This Tip has 3 keys, which can be recovered by bitwise-or-ing the prefix and
```

sjakobi (approved):
Thanks, this is really helpful, @meooow25! The example for `IntSet` is particularly illuminating.

A minor tweak to make the examples even easier to grok would be to use sparser bit patterns. That would make the visual comparisons a bit easier I think.

meooow25:
> A minor tweak to make the examples even easier to grok would be to use sparser bit patterns. That would make the visual comparisons a bit easier I think.

Good idea, thanks!

PR #1237, opened 2026-08-11
Prepare v0.8.1

meooow25:
Bump version, set the since annotations and update the changelog.

For #1186.

### DIFF
--- a/containers/changelog.md
+++ b/containers/changelog.md
@@ -1,27 +1,178 @@
 # Changelog for [`containers` package](http://github.com/haskell/containers)
 
-## Next release
+## 0.8.1  *August 2026*
 
 ### Additions
 
 * Add `compareSize` for `IntSet` and `IntMap`. (Soumik Sarkar)
-  ([#1135](https://github.com/haskell/containers/pull/1135))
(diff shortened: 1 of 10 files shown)


sjakobi:
I'm really looking forward to this release, @meooow25! :)

One thing that was broken in the last release is that Haddock Quick Jump doesn't work on the main landing page (https://hackage.haskell.org/package/containers). The same thing happened in `base` too (https://gitlab.haskell.org/ghc/ghc/-/work_items/24853) and was fixed by https://gitlab.haskell.org/ghc/ghc/-/commit/14e554cf682ae975ba356b07672c0f9d465e2a78. So if you upload docs built with GHC-9.12 or 9.14, it should be fine. You could also leave it to the hackage docs builder.

meooow25:
> One thing that was broken in the last release is that Haddock Quick Jump doesn't work on the main landing page (https://hackage.haskell.org/package/containers).

Oh thanks for pointing that out. I'd prefer to leave it the Hackage docs builder as with 0.8, but I'll check on it and upload docs manually in case it's broken again.

meooow25:
Good to go now. I will make the release in a few days, if nothing else comes up.

Issue #801, opened 2021-09-10
Unboxed sums for performance (Discussion)


Boarders:
It would be useful for us to have a general discussion on places where unboxing and similar techniques might have any benefit. For example there is an old issue here: https://github.com/haskell/containers/pull/523 which tries to use them for `alter` from `IntMap`. In general I would like to gather a list of possible places where we think they could be beneficial so I (or others) can investigate the performance implications.

treeowl:
I think `lookup` for `IntMap` is a pretty obvious spot. Consider this implementation of `member` using `lookup`:

```haskell
member :: Int -> IntMap a -> Bool
member i m = case lookup i m of
  Just _ -> True
  Nothing -> False
```

This produces the following Core:

```
member :: forall a. Int -> IntMap a -> Bool
[GblId,
 Arity=2,
 Caf=NoCafRefs,
 Str=<S,1*U(U)><S,1*U>,
 Unf=Unf{Src=InlineStable, TopLvl=True, Value=True, ConLike=True,
         WorkFree=True, Expandable=True,
         Guidance=ALWAYS_IF(arity=2,unsat_ok=True,boring_ok=False)
         Tmpl= \ (@ a_aDp)
                 (i_aCM [Occ=Once!] :: Int)
                 (m_aCN [Occ=Once] :: IntMap a_aDp) ->
                 case i_aCM of { GHC.Types.I# ww1_aHF [Occ=Once] ->
                 case Data.IntMap.Internal.$wlookup @ a_aDp ww1_aHF m_aCN of {
                   Nothing -> GHC.Types.False;
                   Just _ [Occ=Dead] -> GHC.Types.True
                 }
                 }}]
member
  = \ (@ a_aDp) (i_aCM :: Int) (m_aCN :: IntMap a_aDp) ->
      case i_aCM of { GHC.Types.I# ww1_aHF ->
      case Data.IntMap.Internal.$wlookup @ a_aDp ww1_aHF m_aCN of {
        Nothing -> GHC.Types.False;
        Just ds_dHV -> GHC.Types.True
      }
      }
```

We allocate a `Just` constructor on every successful lookup, even when we immediately deconstruct it. We could rewrite `lookup` thus:

```haskell
lookup :: Key -> IntMap a -> Maybe a
lookup k m = case lookup# k m of
  (# (##) | #) -> Nothing
  (# | a #) -> Just a

lookup# :: Key -> IntMap a -> (# (##) | a #)
lookup# !k (Bin p m l r)
[...]

PR #1238, opened 2026-08-12
Use unboxed Maybe for map lookups on GHC

meooow25:
Trying this out for #801.

### DIFF
--- a/containers-tests/benchmarks/IntMap.hs
+++ b/containers-tests/benchmarks/IntMap.hs
@@ -25,25 +25,25 @@ import Utils.Random (shuffle)
 
 main = do
     let m     = M.fromAscList elems_hits   :: M.IntMap Int
-    let m'    = M.fromAscList elems_mid    :: M.IntMap Int
-    let m''   = M.fromAscList elems_most   :: M.IntMap Int
-    let m'''  = M.fromAscList elems_misses :: M.IntMap Int
-    let m'''' = M.fromAscList elems_mixed  :: M.IntMap Int
-        m_random = M.fromList elems_random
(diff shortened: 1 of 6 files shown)


meooow25:
The benchmarks results are interesting.

`Map`:

```
Name              Time - - - - - - - -    Allocated - - - - -
lookup absent     474 μs  502 μs   +5%    145 B   317 B   +118%
lookup present    469 μs  459 μs   -2%    128 KB  329 B   -99%
```

Great reduction in allocations, but no effect on time saved. I expect GHC is just very fast with short-lived allocations.

`IntMap`:

```
Name             Time - - - - - - - -    Allocated - - - - -
lookup_half      279 μs  296 μs   +6%    148 B   147 B    +0%
lookup_hits      479 μs  522 μs   +8%    1.0 KB  335 B   -67%
lookup_misses    472 μs  499 μs   +5%    309 B   305 B    -1%
lookup_mixed     246 μs  268 μs   +9%    148 B   161 B    +8%
lookup_most      438 μs  474 μs   +8%    308 B   293 B    -4%
```

Surprisingly, there are no allocations even without this change. I checked the Core and found that `lookup` is small enough that GHC inlines it and simplifies, removing the allocations. With this change, GHC no longer inlines `lookup`, which makes it slightly slower. I would argue it is still a good change because it has a positive impact on code size and compile time if `lookup` doesn't inline at every call-site.

meooow25:
Point of reference: unordered-containers does the same thing, see https://github.com/haskell-unordered-containers/unordered-containers/blob/594a3055/Data/HashMap/Internal.hs#L670-L677

PR #1239, opened 2026-08-17
Add Map curry/uncurry utils

alexfmpe:
I've been using a `curryMap` on a project of mine and realised at some point it could be made `O(n)`.
Given it's non-trivial how to do it right, upstreaming seems like it makes sense.

### DIFF
--- a/containers-tests/benchmarks/Map.hs
+++ b/containers-tests/benchmarks/Map.hs
@@ -2,7 +2,7 @@
 {-# LANGUAGE BangPatterns #-}
 module Main where
 
-import Control.Applicative (Const(Const, getConst), pure)
+import Control.Applicative (Const(Const, getConst), liftA2, pure)
 import Control.DeepSeq (rnf)
 import Control.Exception (evaluate)
 import Test.Tasty.Bench (bench, bgroup, defaultMain, whnf, nf)
@@ -31,12 +31,16 @@ main = do
     let m = M.fromList elems :: M.Map Int Int
         m_even = M.fromList elems_even :: M.Map Int Int
         m_odd = M.fromList elems_odd :: M.Map Int Int
+        m_uncurried = M.fromList elems_uncurried :: M.Map (Int, Int) Int
+        m_curried = M.curry m_uncurried :: M.Map Int (M.Map Int Int)
         s_random = Set.fromList keys_random :: Set.Set Int
     evaluate $ rnf [m, m_even, m_odd]
     evaluate $ rnf [s_random]
     evaluate $ rnf
       [elems_distinct_asc, elems_distinct_desc, elems_asc, elems_desc]
     evaluate $ rnf [keys_random]
+    evaluate $ rnf [m_uncurried]
+    evaluate $ rnf [m_curried]
     defaultMain
         [ bench "lookup absent" $ whnf (lookup evens) m_odd
         , bench "lookup present" $ whnf (lookup evens) m_even
@@ -143,6 +147,8 @@ main = do
         , bench "Strict.fromSetA outer" $ whnf (MS.fromSetA (MkSolo . pred)) s_random
         , bench "Lazy.fromSetA inner" $ whnf (getSolo . M.fromSetA (MkSolo . pred)) s_random
         , bench "Strict.fromSetA inner" $ whnf (getSolo . MS.fromSetA (MkSolo . pred)) s_random
+        , bench "curry" $ whnf M.curry m_uncurried
+        , bench "uncurry" $ whnf M.uncurry m_curried
         , bench "minView" $ whnf (\m' -> case M.minViewWithKey m' of {Nothing -> 0; Just ((k,v),m'') -> k+v+M.size m''}) (M.fromAscList $ zip [1..10::Int] [100..110::Int])
         , bench "eq" $ whnf (\m' -> m' == m') m -- worst case, compares everything
         , bench "compare" $ whnf (\m' -> compare m' m') m -- worst case, compares everything
@@ -155,7 +161,8 @@ main = do
         , bench "mapKeysWith:desc" $ whnf (M.mapKeysWith (+) (negate . (`div` 2))) m
         ]
   where
-    bound = 2^14
+    magnitude = 14
+    bound = 2^magnitude
     elems = shuffle elems_distinct_asc
     elems_even = zip evens evens
     elems_odd = zip odds odds
@@ -172,6 +179,11 @@ main = do
     sumkv k v1 v2 = k + v1 + v2
     consPair k v xs = (k, v) : xs
     keys_random = take bound (randoms gen)
+    elems_uncurried = zip xs evens
+      where
+        left = magnitude `div` 2
+        right = magnitude - left
+        xs = shuffle $ liftA2 (,) [1..2^left] $ reverse [1..2^right]
 
 add3 :: Int -> Int -> Int -> Int
 add3 x y z = x + y + z
--- a/containers-tests/tests/map-properties.hs
+++ b/containers-tests/tests/map-properties.hs
@@ -34,7 +34,7 @@ import Data.Functor
 import qualified Data.Foldable as Foldable
 import qualified Data.Bifoldable as Bifoldable
 import Data.Proxy (Proxy(..))
-import Prelude hiding (lookup, null, map, filter, foldr, foldl, foldl', take, drop, splitAt)
+import Prelude hiding (lookup, null, map, filter, foldr, foldl, foldl', take, drop, splitAt, curry, uncurry)
 import qualified Prelude
 
 import Data.List (nub,sort)
@@ -278,6 +278,7 @@ main = defaultMain $ testGroup "map-properties"
          , testProperty "fromSetMaybe"         prop_fromSetMaybe
          , testProperty "fromSetMaybeA"        prop_fromSetMaybeA
          , testProperty "fromArgSet"           prop_fromArgSet
+         , testProperty "curry"                prop_curry
          , testProperty "takeWhileAntitone"    prop_takeWhileAntitone
          , testProperty "dropWhileAntitone"    prop_dropWhileAntitone
          , testProperty "spanAntitone"         prop_spanAntitone
@@ -1576,7 +1577,7 @@ prop_fromList :: [Int] -> Bool
 prop_fromList xs
   = case fromList (zip xs xs) of
       t -> t == fromAscList (zip sort_xs sort_xs) &&
-           t == List.foldr (uncurry insert) empty (zip xs xs)
+           t == List.foldr (Prelude.uncurry insert) empty (zip xs xs)
   where sort_xs = sort xs
 
 prop_fromAscList :: SortedOnFst Int A -> Property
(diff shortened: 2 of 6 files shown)


containers-tests/benchmarks/Map.hs
```diff
@@ -143,6 +147,8 @@ main = do
         , bench "Strict.fromSetA outer" $ whnf (MS.fromSetA (MkSolo . pred)) s_random
         , bench "Lazy.fromSetA inner" $ whnf (getSolo . M.fromSetA (MkSolo . pred)) s_random
         , bench "Strict.fromSetA inner" $ whnf (getSolo . MS.fromSetA (MkSolo . pred)) s_random
+        , bench "curryMap" $ whnf M.curryMap m_uncurried
+        , bench "uncurryMap" $ whnf M.uncurryMap m_curried
```
alexfmpe: I mostly cargo-culted here. It doesn't make sense to me that `curryMap` comes out faster than `lookup` on benchmarks but then again so does `fromSet` ?

```
  lookup present:                 OK
    645  μs ±  53 μs, 128 KB allocated,  10 B  copied,  23 MB peak memory
  lookup absent:                  OK
    703  μs ±  60 μs,  54 B  allocated,   1 B  copied,  23 MB peak memory

  fromSet:               OK
    195  μs ±  14 μs, 1.3 MB allocated, 201 KB copied,  23 MB peak memory
```

containers-tests/tests/map-properties.hs
```diff
@@ -2013,6 +2014,10 @@ prop_fromArgSet :: [(OrdA, B)] -> Property
 prop_fromArgSet ys =
   fromArgSet (Set.fromList $ List.map (uncurry Arg) ys) === fromList ys
 
+prop_curryMap :: Map Int (Map Int A) -> Property
+prop_curryMap m = m' === uncurryMap (curryMap m')
+  where m' = uncurryMap m
```
alexfmpe: I wanted to simply have
```
prop_curryMap :: Map (Int, Int) A -> Property
prop_curryMap m = m === curryMap (uncurryMap m)
```
but that yields `• No instance for ‘IsInt (Int, Int)’` for some reason.

containers/src/Data/Map/Internal.hs
```diff
@@ -3427,6 +3431,35 @@ fromArgSet :: Set.Set (Arg k a) -> Map k a
...
+{--------------------------------------------------------------------
+  Maps
+--------------------------------------------------------------------}
+-- | \(O(n)\). Group map entries by the first component.
+--
+-- > curryMap $ fromList [((1,2),12),((1,3),13)] == fromList [(1,fromList [(2,12),(3,13)])]
+--
+-- @since FIXME
+curryMap :: (Ord a, Ord b) => Map (a,b) c -> Map a (Map b c)
+curryMap m = fmap (fromDescList . ($ [])) $ fromAscListWith (.) $ fmap (\((a,b),c) -> (a, ((b,c):))) $ toAscList m
```
alexfmpe: Surprisingly to me this consistently comes up slightly faster and with significantly lower allocations 

```
  curryMap:   OK
    262  μs ±  15 μs, 2.3 MB allocated, 116 KB copied,  23 MB peak memory
``` 

than 
```
  curryMap:   OK
    273  μs ±  13 μs, 2.4 MB allocated, 142 KB copied,  23 MB peak memory

curryMap m = fmap M.fromDescList $ M.fromAscListWith (<>) $ fmap (\((a,b),c) -> (a, [(b,c)])) $ M.toAscList m
```
on `cabal bench map-benchmarks --benchmark-options "-p curryMap +RTS -T"`. 


I guess it's because we skip the pattern match on singleton lists, but I expected the overhead of a closure to be worse. I guess GHC gets to optimize this case a lot more.

containers/src/Data/Map/Strict/Internal.hs
```diff
@@ -1471,6 +1475,35 @@ fromArgSet :: Set.Set (Arg k a) -> Map k a
...
+{--------------------------------------------------------------------
+  Maps
+--------------------------------------------------------------------}
+-- | \(O(n)\). Group map entries by the first component.
+--
+-- > curryMap $ fromList [((1,2),12),((1,3),13)] == fromList [(1,fromList [(2,12),(3,13)])]
+--
+-- @since FIXME
+curryMap :: (Ord a, Ord b) => Map (a,b) c -> Map a (Map b c)
+curryMap m = fmap (fromDescList . ($ [])) $ fromAscListWith (.) $ fmap (\((a,b),c) -> (a, ((b,c):))) $ toAscList m
```
Ericson2314:
```suggestion
curry :: (Ord a, Ord b) => Map (a,b) c -> Map a (Map b c)
curry m = fmap (fromDescList . ($ [])) $ fromAscListWith (.) $ fmap (\((a,b),c) -> (a, ((b,c):))) $ toAscList m
```
I think just this because the intended use-case would be to import qualified (`Data.Map.curry`) and use that for disambiguation.

meooow25:
This is interesting to consider for sure, but I am not not seeing the utility. In what situation do you use it in your project?

containers/src/Data/Map/Strict/Internal.hs
alexfmpe: Agreed, I forgot to change the names when upstreaming

alexfmpe:
> This is interesting to consider for sure, but I am not not seeing the utility. In what situation do you use it in your project?

I am given a `Map (a,b) Word` and want to then compute some stats over it keyed by `a` so I do `stats . curryMap`. e.g.
```
stats :: Map a (Map b Word) -> Map a (Word, Word)
stats = fmap $ (toEnum . length) &&& sum
```

I added `uncurry` purely out of symmetry

IIRC @Ericson2314 also had a FRP use-case? `curry`-ing `Map` to create hierarchical UI lists or so?

meooow25:
How do you have a `Map (a,b) Word` in the first place? I think that's the part which seems most like a wart to me. Almost all functions in the interface are polymorphic on the key type, which is a good property to have since the users can use their types as keys.

Consider that somebody wants

```hs
data Foo = Foo !Bar !Baz deriving Ord

curryFooMap :: Map Foo a -> Map Bar (Map Baz a)
curryFooMap = ??
```

but they are unable to use `curryMap` because it's not `(a,b)`, though they should ideally be able to use what is offered from the Map interface.

alexfmpe:
It goes something like this
1. build a `Map ByteString c` 
1. parse the `ByteString` into `(a,b)` 
1. call `stats` from above comment

(1) is done by streaming lines from a file (with `Map.insertWith (+)` to count ocurrences)
(2) is done by going via `List` and `mapMaybe` because *in theory* the parse can fail.  

`ByteString` is basically the concatenation of encoded `a` and `b`.
Parsing after is way faster than before regardless of whether I accumulate into `Map (a,b) c` or `Map a (Map b c)`, presumably because by doing it after Map de-dup there's less much less parsing and `Ord` on `ByteString` is pretty fast? So I accumulate, then parse, then curry. Roughly, that gives me
- 50ms - parse before, flat map
- 48ms - parse before, nested map
- 35ms - parse after

Btw, as of https://github.com/haskell/containers/pull/1230, (2) can be done with `Map.mapAssocsMonotonic` and `fromJust`. `mapAssocsMonotonicMaybe` anyone?

> Almost all functions in the interface are polymorphic on the key type, which is a good property to have since the users can use their types as keys.

I mean, it is polymorphic in `a` and `b`. By that logic we'd also reject `Map.catMaybes` or a one-pass `Map.unzip`
Some sort of `curryMonotonic :: (a -> (b,c)) -> Map a d -> Map b (Map c d)` would pass this criteria, but it's decomposable into `curryMonotonic f = curry . mapKeysMonotonic f`

meooow25:
Thanks, that sheds some light on it. I think it also echoes my point about it being odd to have `(a,b)` as the key. In this case what is required overall is

```hs
foo :: (ByteString -> Maybe (Key1, Key2)) -> Map ByteString Word -> Map Key1 (Map Key2 Word)
```

and `Map (Key1, Key2) Word` is an intermediate which is not strictly necessary.

I'm trying to think if there is a nice way to offer this functionality...

By the way, there is a third implementation you could try using `fromAscListUpsert (\x -> maybe [x] (x:))`, instead of `fromAscListWith (.)` or `fromAscListWith (++)`. This is on master but not yet released on Hackage.

> I mean, it is polymorphic in `a` and `b`. By that logic we'd also reject `Map.catMaybes` or a one-pass `Map.unzip`

Yes I do feel the same way about `catMaybes` and `unzip` :)

alexfmpe:
> By the way, there is a third implementation you could try using fromAscListUpsert (\x -> maybe [x] (x:)), instead of fromAscListWith (.) or fromAscListWith (++). This is on master but not yet released on Hackage.

Ooooh nice I've wanted exactly this for something else but couldn't find it. I've been able to exploit some structure of my input to go straight to nested maps which then I accumulate with
```haskell
IntMap.fromListWith (Map.unionWith (+))
```
but the `Map.unionWith (+)` felt wasteful since one of them is always a singleton. 
```haskell
IntMap.fromListUpsert (\(e,c) -> maybe (Map.singleton e c) (Map.insertWith (+) e c))
```
made for 7% speedup on the merging part. Seems to make no difference for `curry` besides slightly larger allocations:

```
fromAscListWith:   OK
  256  μs ±  15 μs, 2.3 MB allocated, 116 KB copied,  23 MB peak memory

fromAscListUpsert:   OK
  256  μs ±  14 μs, 2.4 MB allocated, 119 KB copied,  23 MB peak memory
```

> Yes I do feel the same way about catMaybes and unzip :)

Well, the venerable `join :: [[a]] -> [a]` also fails the "not polymorphic" criteria.
It's easy to get a sequence of sequences, that's how we view text files, as a sequence of lines which are themselves sequences of chars. It's also the output of many `group` functions.

> Map (Key1, Key2) Word is an intermediate which is not strictly necessary.

It's not strictly needed in my use-case, but having a product/record type as `Map` key happens all the time no? Even if we were confident there are ways to avoid producing it, one cannot control what third-party APIs return.

> I'm trying to think if there is a nice way to offer this functionality...

Tricky. At first glance one might think that offering the small pieces covers every use-case, e.g.
```haskell
foo :: (a -> Maybe (a1, a2)) -> Map a b -> Map a1 (Map a2 b)
foo f = curryMap . mapKeys f
```
but notice for `IntMap` the intermediate does not exist. At most there's
```haskell
module Data.IntMap
curry :: Map (Int,a) b -> IntMap (Map a b)
uncurry :: IntMap (Map a b) -> Map (Int,a) b
```
This pattern extends far beyond `containers`: [`concatMap`](https://hackage-content.haskell.org/package/text-2.1.4/docs/Data-Text.html#v:concatMap) exists for `Text`, but `join` cannot.

alexfmpe:
That last point made me notice there's other possible combinations, since `IntMap a ~ Map Int a` and `Set a ~ Map a ()`
```haskell
curry :: Set (a,b) -> Map a (Set b)
uncurry :: Map a (Set b) -> Set (a,b)
```
```haskell
curry :: Set (Int,b) -> IntMap (Set b)
uncurry :: IntMap (Set b) -> Set (Int,b)
```

Issue #227, opened 2016-05-16
mergeWithKey' may be too general for its own good


treeowl:
It's cute that `mergeWithKey'` can implement so many different things, but there's a performance cost. Different operations necessitate different `Nil` checking, which `mergeWithKey'` has no way to deal with. The fact that the functions passed to it are often partial is also a bit disturbing, suggesting there may be a better approach. Having the combining function take maps instead of keys and values to enhance sharing sounds very reasonable until you see that it's actually only called for `Tip` trees. So the sharing is probably significant, but not earth-shattering. Using something like the evil pointer equality check could potentially have a much bigger impact, I imagine. Short of that, the partial functions could be avoided by passing both the `Tip` and its contents in separate arguments, although that's kind of ugly.

PR #1241, opened 2026-08-21
IntMap: improve sharing in `difference` via pointer equality

sjakobi:
Replace the mergeWithKey-based difference with a direct implementation that uses ptrEq to return the original (sub)tree when a recursive call removes nothing, instead of allocating an equal copy. This mirrors the sharing behavior Data.Map.difference gets from its size check and what Data.Set does with ptrEq.

Context: #835, #227

----

TODO:

* [ ] run benchmarks
* [ ] check perf implications of SpecConstr-workaround via `lazy`
* [ ] tests for sharing?!
* [ ] Improve Note to read less "LLM-y"

### DIFF
--- a/containers/src/Data/IntMap/Internal.hs
+++ b/containers/src/Data/IntMap/Internal.hs
@@ -354,13 +354,14 @@ import Data.IntSet.Internal.IntTreeCommons
 import Utils.Containers.Internal.BitUtil (shiftLL, shiftRL, iShiftRL, wordSize)
 import Utils.Containers.Internal.Strict
   (StrictPair(..), StrictTriple(..), toPair)
+import Utils.Containers.Internal.PtrEquality (ptrEq)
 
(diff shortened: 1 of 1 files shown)


meooow25:
Thanks for being interested in this, but before you spend more time on it let me say right away that I don't intend to review or accept any LLM-generated code. Of course, I don't speak for other maintainers, and maybe there is a discussion to be had about it, but my position is clear.

sjakobi:
Thanks for the heads-up, @meooow25! :) I should have probably made this clearer in the description, but I didn't mean to request your review at this stage. I opened the draft mostly to signal "I am interested in this".

So I've started looking into SpecConstr on the GHC side for now, but eventually I'd like to see some sharing optimizations for IntMap and IntSet land here, including `difference`. Do you have any advice for me how to go about this? Would you consider reviewing and potentially accepting a fresh PR that is handwritten by me? Or is there another process that you would prefer?

PR #1255, opened 2026-09-17
Add some missing Set benchmarks

alexfmpe:


meooow25 (approved):
Seems fine to add these.

PR #865, opened 2022-10-29
Faster disjoint/isSubsetOf for Set via unbalanced splitting.

alexfmpe:
Unlike, say, `union`/`intersection`, `disjoint` doesn't return a new structure. It can avoid the re-balancing work because it immediately inspects and forgets the produced tree. This allows significant constant factor speedups.

```All
  member:                    OK (0.14s)
    133  μs ±  12 μs,   0 B  allocated,   0 B  copied, 8.0 MB peak memory,       same as baseline
  insert:                    OK (0.18s)
    678  μs ±  47 μs, 2.6 MB allocated,  55 KB copied, 8.0 MB peak memory,       same as baseline
  map:                       OK (0.22s)
    106  μs ± 6.0 μs, 557 KB allocated,  16 KB copied, 8.0 MB peak memory,       same as baseline
  filter:                    OK (0.17s)
    80.2 μs ± 6.4 μs,  80 KB allocated, 1.2 KB copied, 8.0 MB peak memory,       same as baseline
  partition:                 OK (0.16s)
    148  μs ±  15 μs, 239 KB allocated, 4.0 KB copied, 8.0 MB peak memory,       same as baseline
  fold:                      OK (0.31s)
    34.9 ns ± 1.5 ns, 695 B  allocated,   0 B  copied, 8.0 MB peak memory,       same as baseline
  delete:                    OK (0.12s)
    454  μs ±  44 μs, 1.3 MB allocated, 372 B  copied, 8.0 MB peak memory,       same as baseline
  findMin:                   OK (0.27s)
    15.7 ns ± 784 ps,  31 B  allocated,   0 B  copied, 8.0 MB peak memory,       same as baseline
  findMax:                   OK (0.30s)
    17.1 ns ± 842 ps,  31 B  allocated,   0 B  copied, 8.0 MB peak memory,       same as baseline
  deleteMin:                 OK (0.25s)
    114  ns ± 7.0 ns, 471 B  allocated,   0 B  copied, 8.0 MB peak memory,       same as baseline
  deleteMax:                 OK (0.25s)
    117  ns ± 7.0 ns, 511 B  allocated,   0 B  copied, 8.0 MB peak memory,       same as baseline
  unions:                    OK (0.31s)
    147  μs ± 7.2 μs, 472 KB allocated,  10 KB copied, 8.0 MB peak memory,       same as baseline
  union:                     OK (0.16s)
[...]

PR #1256, opened 2026-09-17
List-based implementation of boolean algebra operations for Set 

alexfmpe:
Here's an experiment: how do simple linear merge implementations compare in practice to the asymptotical optimal ones?
Seems to be competitive depending a lot on size/benchmark.
I think this is worth exploring further but wanted to get feedback first.

We could merely add new variants but it might also be worth looking at the sizes/ratios of input trees to pick the implementation. In either case, I think we need a _lot_ more benchmarks to have a better idea when any of this is actually worth it. In particular, it needs to be easy to see how results vary as size(s) vary.
There's a lot of charts in the papers linked but I imagine the constant factors are all kinds of different to what we're running.

The current implementations grab the root of one tree as a proxy for median value, then find its place in the other tree to split that one as well so that the problem is divided into two smaller ones on smaller tree pairs.
IIUC, this interpolates/generalises between two extremes, with `n >= m`:
1. when one list is a singleton, this acts as an insert in `O(lg n)`
2. when both lists are of similar size, we end up with `O(n)`

The recursive structure of that algorithm also has the advantage that it's highly parallelizable. 
However, I don't think we're actually leveraging that?
Seems to me that always doubling parallelism on every layer would backfire from contention and overhead so we'd need to require/choose a tree size/depth cutoff, take `numCapabilities` or so into account, that sort of thing.
Though I can imagine variants that receive a capabilities/workers count and use that to divide up the work when branching.

Given we can convert to/from lists in linear time, and that we can merge two ordered sequences in linear time, I had a go at the list-based implementations for `Set`. This in principle should be extendable to the `Map` equivalents and to `Map`/`Set` mixes like `restrictKeys`. 

## Benchmarks

I ran the benchmarks on top of https://github.com/haskell/containers/pull/1255 with
```
cabal run set-benchmarks -- +RTS -T -RTS --stdev 5 --baseline set-baseline.csv -p '$0 ~ /difference/ || $0 ~ /intersection/ || $0 ~ /union/ || $0 ~ /disjoint/ || $0 ~ /symmetricDifference/ || $0 ~ /isSubsetOf/'
```

### `bound = 2^12`
This is the current value of `bound`
#### baseline
```
  unions:                  OK
    82.5 μs ± 7.8 μs, 477 KB allocated,  10 KB copied,  10 MB peak memory
  union:                   OK
    82.0 μs ± 7.7 μs, 477 KB allocated,  10 KB copied,  10 MB peak memory
  difference:              OK
    84.1 μs ± 7.6 μs, 343 KB allocated, 3.4 KB copied,  10 MB peak memory
  symmetricDifference:     OK
    88.3 μs ± 4.7 μs, 320 KB allocated, 2.9 KB copied,  10 MB peak memory
  intersection:            OK
    86.5 μs ± 3.3 μs, 322 KB allocated, 2.9 KB copied,  10 MB peak memory
  disjoint:false:          OK
    397  ns ±  26 ns, 2.0 KB allocated,   0 B  copied,  10 MB peak memory
  disjoint:true:           OK
    84.1 μs ± 7.9 μs, 236 KB allocated, 132 B  copied,  10 MB peak memory
  isSubsetOf:false:        OK
    93.2 ns ± 6.5 ns, 904 B  allocated,   0 B  copied,  10 MB peak memory
  isSubsetOf:true:         OK
    38.4 μs ± 3.3 μs, 188 KB allocated,  87 B  copied,  10 MB peak memory
  null.intersection:false: OK
    86.4 μs ± 7.2 μs, 322 KB allocated, 2.9 KB copied,  10 MB peak memory
  null.intersection:true:  OK
    70.3 μs ± 6.9 μs, 305 KB allocated, 167 B  copied,  10 MB peak memory
```
#### list-based
```
  unions:                  OK
    106  μs ± 7.8 μs, 897 KB allocated,  25 KB copied, 9.0 MB peak memory, 28% more than baseline
  union:                   OK
    68.9 μs ± 6.8 μs, 608 KB allocated,  16 KB copied, 9.0 MB peak memory, 16% less than baseline
  difference:              OK
    40.5 μs ± 1.7 μs, 288 KB allocated, 3.2 KB copied,  10 MB peak memory, 51% less than baseline
  symmetricDifference:     OK
    40.7 μs ± 3.5 μs, 288 KB allocated, 3.2 KB copied,  10 MB peak memory, 53% less than baseline
  intersection:            OK
    43.1 μs ± 3.3 μs, 304 KB allocated, 3.2 KB copied,  10 MB peak memory, 50% less than baseline
  disjoint:false:          OK
    63.1 ns ± 3.3 ns, 1.2 KB allocated,   0 B  copied,  10 MB peak memory, 84% less than baseline
  disjoint:true:           OK
    54.0 μs ± 4.1 μs, 224 KB allocated,  44 B  copied,  10 MB peak memory, 35% less than baseline
  isSubsetOf:false:        OK
    47.4 ns ± 3.7 ns, 1.1 KB allocated,   0 B  copied,  10 MB peak memory, 49% less than baseline
  isSubsetOf:true:         OK
    62.8 μs ± 3.8 μs, 336 KB allocated,  66 B  copied,  10 MB peak memory, 63% more than baseline
  null.intersection:false: OK
    43.2 μs ± 1.7 μs, 304 KB allocated, 3.2 KB copied,  10 MB peak memory, 49% less than baseline
  null.intersection:true:  OK
    26.7 μs ± 1.8 μs, 112 KB allocated,  14 B  copied,  10 MB peak memory, 62% less than baseline
```
Very promising results overall. Usually less allocation and time. Not sure why `isSubsetOf:true` swims against the tide. As for `unions` it likely ends up slower is because it starts folding with `empty`, so we get an extra round of merging unless we keep the `Tip` short-circuits on `union`. 
I tried adding the short-circuits back to all the top-level bindings and while that made `unions` mirror the speedups (the benchmark uses a 2-element list afterall), a lot of the gains elsewhere got nerfed for some reason?

#### short-circuit
```
  unions:                  OK
    67.6 μs ± 6.6 μs, 609 KB allocated,  16 KB copied, 9.0 MB peak memory, 18% less than baseline
  union:                   OK
    67.8 μs ± 6.6 μs, 608 KB allocated,  16 KB copied, 9.0 MB peak memory, 17% less than baseline
  difference:              OK
    76.4 μs ± 7.3 μs, 513 KB allocated, 4.6 KB copied, 9.0 MB peak memory,  9% less than baseline
  symmetricDifference:     OK
    78.1 μs ± 5.7 μs, 512 KB allocated, 4.6 KB copied, 9.0 MB peak memory, 11% less than baseline
  intersection:            OK
    80.3 μs ± 4.2 μs, 529 KB allocated, 4.6 KB copied, 9.0 MB peak memory,  7% less than baseline
  disjoint:false:          OK
    61.8 ns ± 3.3 ns, 1.2 KB allocated,   0 B  copied, 9.0 MB peak memory, 84% less than baseline
  disjoint:true:           OK
    52.7 μs ± 3.6 μs, 224 KB allocated,  44 B  copied, 9.0 MB peak memory, 37% less than baseline
  isSubsetOf:false:        OK
    50.1 ns ± 3.8 ns, 1.2 KB allocated,   0 B  copied, 9.0 MB peak memory, 46% less than baseline
  isSubsetOf:true:         OK
    61.4 μs ± 4.9 μs, 336 KB allocated,  66 B  copied, 9.0 MB peak memory, 59% more than baseline
  null.intersection:false: OK
    78.6 μs ± 7.2 μs, 529 KB allocated, 4.6 KB copied, 9.0 MB peak memory,  8% less than baseline
  null.intersection:true:  OK
    37.7 μs ± 3.5 μs, 224 KB allocated,  45 B  copied, 9.0 MB peak memory, 46% less than baseline
```

Now, as we increase `bound`, the list-based implementations start performing worse and worse
#### baseline
```
  unions:                  OK
    4.57 ms ± 305 μs, 7.5 MB allocated, 6.5 MB copied, 101 MB peak memory
  union:                   OK
    4.56 ms ± 377 μs, 7.5 MB allocated, 6.5 MB copied, 101 MB peak memory
  difference:              OK
    2.53 ms ± 229 μs, 5.3 MB allocated, 1.5 MB copied, 101 MB peak memory
  symmetricDifference:     OK
    2.63 ms ± 171 μs, 5.0 MB allocated, 1.2 MB copied, 101 MB peak memory
  intersection:            OK
    2.51 ms ±  78 μs, 5.0 MB allocated, 1.2 MB copied, 101 MB peak memory
  disjoint:false:          OK
    103  ns ± 7.4 ns, 584 B  allocated,   0 B  copied, 101 MB peak memory
  disjoint:true:           OK
    1.65 ms ± 131 μs, 3.7 MB allocated, 3.6 KB copied, 101 MB peak memory
  isSubsetOf:false:        OK
    110  ns ± 7.7 ns, 1.2 KB allocated,   0 B  copied, 101 MB peak memory
  isSubsetOf:true:         OK
    1.46 ms ± 132 μs, 2.9 MB allocated, 2.4 KB copied, 101 MB peak memory
  null.intersection:false: OK
    2.51 ms ± 110 μs, 5.0 MB allocated, 746 KB copied, 101 MB peak memory
  null.intersection:true:  OK
    1.42 ms ±  93 μs, 4.8 MB allocated, 4.5 KB copied, 101 MB peak memory
```
#### list-based
```
  unions:                  OK
    9.83 ms ± 422 μs,  14 MB allocated,  13 MB copied, 102 MB peak memory, 115% more than baseline
  union:                   OK
    7.91 ms ± 244 μs, 9.5 MB allocated,  11 MB copied, 105 MB peak memory, 73% more than baseline
  difference:              OK
    1.72 ms ± 125 μs, 4.5 MB allocated, 1.2 MB copied, 105 MB peak memory, 32% less than baseline
  symmetricDifference:     OK
    1.74 ms ±  82 μs, 4.5 MB allocated, 1.2 MB copied, 105 MB peak memory, 33% less than baseline
  intersection:            OK
    2.16 ms ± 123 μs, 4.8 MB allocated, 1.3 MB copied, 105 MB peak memory, 14% less than baseline
  disjoint:false:          OK
    116  ns ± 9.2 ns, 1.8 KB allocated,   0 B  copied, 105 MB peak memory, 12% more than baseline
  disjoint:true:           OK
    1.36 ms ± 124 μs, 3.5 MB allocated, 921 B  copied, 105 MB peak memory, 17% less than baseline
  isSubsetOf:false:        OK
    95.0 ns ± 7.3 ns, 1.7 KB allocated,   0 B  copied, 105 MB peak memory, 13% less than baseline
  isSubsetOf:true:         OK
    1.63 ms ± 133 μs, 5.3 MB allocated, 1.3 KB copied, 105 MB peak memory, 11% more than baseline
  null.intersection:false: OK
    1.79 ms ±  74 μs, 4.8 MB allocated, 832 KB copied, 105 MB peak memory, 28% less than baseline
  null.intersection:true:  OK
    606  μs ±  60 μs, 1.8 MB allocated, 278 B  copied, 105 MB peak memory, 57% less than baseline
```

#### baseline
```
  unions:                  OK
    80.7 ms ± 7.8 ms, 120 MB allocated,  86 MB copied, 1.2 GB peak memory
  union:                   OK
    101  ms ± 6.5 ms, 120 MB allocated, 124 MB copied, 1.5 GB peak memory
  difference:              OK
    69.8 ms ± 5.6 ms,  86 MB allocated,  37 MB copied, 1.5 GB peak memory
  symmetricDifference:     OK
    64.1 ms ± 3.6 ms,  80 MB allocated,  51 MB copied, 1.5 GB peak memory
  intersection:            OK
    63.1 ms ± 2.4 ms,  80 MB allocated,  53 MB copied, 1.5 GB peak memory
  disjoint:false:          OK
    677  ns ±  62 ns, 3.7 KB allocated,   1 B  copied, 1.5 GB peak memory
  disjoint:true:           OK
    28.6 ms ± 1.8 ms,  59 MB allocated,  83 KB copied, 1.5 GB peak memory
  isSubsetOf:false:        OK
    137  ns ±  13 ns, 1.4 KB allocated,   0 B  copied, 1.5 GB peak memory
  isSubsetOf:true:         OK
    38.1 ms ± 383 μs,  47 MB allocated,  60 KB copied, 1.5 GB peak memory
  null.intersection:false: OK
    61.5 ms ± 3.2 ms,  80 MB allocated,  50 MB copied, 1.5 GB peak memory
  null.intersection:true:  OK
    27.0 ms ± 347 μs,  76 MB allocated, 103 KB copied, 1.5 GB peak memory
```
#### list-based
```
  unions:                  OK
    254  ms ± 9.3 ms, 224 MB allocated, 388 MB copied, 1.6 GB peak memory, 215% more than baseline
  union:                   OK
    165  ms ± 2.7 ms, 152 MB allocated, 275 MB copied, 1.6 GB peak memory, 62% more than baseline
  difference:              OK
    90.1 ms ± 4.9 ms,  72 MB allocated, 112 MB copied, 1.6 GB peak memory, 29% more than baseline
  symmetricDifference:     OK
    91.7 ms ± 3.0 ms,  72 MB allocated, 118 MB copied, 1.6 GB peak memory, 43% more than baseline
  intersection:            OK
    80.6 ms ± 1.0 ms,  76 MB allocated, 110 MB copied, 1.6 GB peak memory, 27% more than baseline
  disjoint:false:          OK
    96.4 ns ± 7.7 ns, 2.2 KB allocated,   0 B  copied, 1.6 GB peak memory, 85% less than baseline
  disjoint:true:           OK
    45.3 ms ± 949 μs,  56 MB allocated,  41 MB copied, 1.6 GB peak memory, 58% more than baseline
  isSubsetOf:false:        OK
    84.1 ns ± 6.7 ns, 2.1 KB allocated,   0 B  copied, 1.6 GB peak memory, 38% less than baseline
  isSubsetOf:true:         OK
    66.7 ms ± 4.4 ms,  84 MB allocated,  64 MB copied, 1.6 GB peak memory, 74% more than baseline
  null.intersection:false: OK
    76.1 ms ± 3.3 ms,  76 MB allocated, 105 MB copied, 1.6 GB peak memory, 23% more than baseline
  null.intersection:true:  OK
    20.6 ms ± 1.8 ms,  28 MB allocated,  17 MB copied, 1.6 GB peak memory, 23% less than baseline
```

If we instead decrease `bound`, we remain with positive looking results, though the gains seem to decrease. Maybe with small trees, the asymptotical optimal algorithm tends more toward (1)?

#### baseline
```
  unions:                  OK
    4.43 μs ± 423 ns,  29 KB allocated,  41 B  copied, 6.0 MB peak memory
  union:                   OK
    4.42 μs ± 429 ns,  29 KB allocated,  43 B  copied, 6.0 MB peak memory
  difference:              OK
    4.25 μs ± 424 ns,  21 KB allocated,  17 B  copied, 6.0 MB peak memory
  symmetricDifference:     OK
    4.58 μs ± 444 ns,  19 KB allocated,  13 B  copied, 6.0 MB peak memory
  intersection:            OK
    4.63 μs ± 413 ns,  20 KB allocated,  14 B  copied, 6.0 MB peak memory
  disjoint:false:          OK
    327  ns ±  27 ns, 1.6 KB allocated,   0 B  copied, 6.0 MB peak memory
  disjoint:true:           OK
    4.73 μs ± 414 ns,  14 KB allocated,   3 B  copied, 6.0 MB peak memory
  isSubsetOf:false:        OK
    57.1 ns ± 3.2 ns, 504 B  allocated,   0 B  copied, 6.0 MB peak memory
  isSubsetOf:true:         OK
    2.11 μs ± 209 ns,  11 KB allocated,   2 B  copied, 6.0 MB peak memory
  null.intersection:false: OK
    4.63 μs ± 449 ns,  20 KB allocated,  14 B  copied, 6.0 MB peak memory
  null.intersection:true:  OK
    3.92 μs ± 227 ns,  18 KB allocated,   3 B  copied, 6.0 MB peak memory
```
#### list-based
```
  unions:                  OK
    5.43 μs ± 530 ns,  57 KB allocated,  81 B  copied, 6.0 MB peak memory, 22% more than baseline
  union:                   OK
    3.80 μs ± 222 ns,  38 KB allocated,  53 B  copied, 6.0 MB peak memory, 14% less than baseline
  difference:              OK
    2.30 μs ± 214 ns,  18 KB allocated,  13 B  copied, 6.0 MB peak memory, 45% less than baseline
  symmetricDifference:     OK
    2.29 μs ± 206 ns,  18 KB allocated,  13 B  copied, 6.0 MB peak memory, 49% less than baseline
  intersection:            OK
    2.46 μs ± 222 ns,  19 KB allocated,  14 B  copied, 6.0 MB peak memory, 46% less than baseline
  disjoint:false:          OK
    50.1 ns ± 4.0 ns, 752 B  allocated,   0 B  copied, 6.0 MB peak memory, 84% less than baseline
  disjoint:true:           OK
    3.28 μs ± 214 ns,  14 KB allocated,   1 B  copied, 6.0 MB peak memory, 30% less than baseline
  isSubsetOf:false:        OK
    34.2 ns ± 3.3 ns, 720 B  allocated,   0 B  copied, 6.0 MB peak memory, 40% less than baseline
  isSubsetOf:true:         OK
    3.28 μs ± 219 ns,  21 KB allocated,   2 B  copied, 6.0 MB peak memory, 55% more than baseline
  null.intersection:false: OK
    2.47 μs ± 217 ns,  19 KB allocated,  14 B  copied, 6.0 MB peak memory, 46% less than baseline
  null.intersection:true:  OK
    1.51 μs ± 104 ns, 7.0 KB allocated,   0 B  copied, 6.0 MB peak memory, 61% less than baseline
```

#### baseline
```
  unions:                  OK
    244  ns ±  13 ns, 1.7 KB allocated,   0 B  copied, 6.0 MB peak memory
  union:                   OK
    236  ns ±  15 ns, 1.7 KB allocated,   0 B  copied, 6.0 MB peak memory
  difference:              OK
    234  ns ±  13 ns, 1.1 KB allocated,   0 B  copied, 6.0 MB peak memory
  symmetricDifference:     OK
    291  ns ±  28 ns, 1.2 KB allocated,   0 B  copied, 6.0 MB peak memory
  intersection:            OK
    296  ns ±  30 ns, 1.2 KB allocated,   0 B  copied, 6.0 MB peak memory
  disjoint:false:          OK
    63.2 ns ± 3.7 ns, 224 B  allocated,   0 B  copied, 6.0 MB peak memory
  disjoint:true:           OK
    242  ns ±  14 ns, 704 B  allocated,   0 B  copied, 6.0 MB peak memory
  isSubsetOf:false:        OK
    35.4 ns ± 3.5 ns, 224 B  allocated,   0 B  copied, 6.0 MB peak memory
  isSubsetOf:true:         OK
    123  ns ± 6.6 ns, 664 B  allocated,   0 B  copied, 6.0 MB peak memory
  null.intersection:false: OK
    293  ns ±  28 ns, 1.2 KB allocated,   0 B  copied, 6.0 MB peak memory
  null.intersection:true:  OK
    215  ns ±  14 ns, 984 B  allocated,   0 B  copied, 6.0 MB peak memory
```

#### list-based
```
  unions:                  OK
    379  ns ±  27 ns, 3.8 KB allocated,   0 B  copied, 6.0 MB peak memory, 56% more than baseline
  union:                   OK
    252  ns ±  13 ns, 2.5 KB allocated,   0 B  copied, 6.0 MB peak memory,  6% more than baseline
  difference:              OK
    148  ns ±  14 ns, 1.3 KB allocated,   0 B  copied, 6.0 MB peak memory, 36% less than baseline
  symmetricDifference:     OK
    157  ns ± 8.2 ns, 1.3 KB allocated,   0 B  copied, 6.0 MB peak memory, 46% less than baseline
  intersection:            OK
    161  ns ±  13 ns, 1.4 KB allocated,   0 B  copied, 6.0 MB peak memory, 45% less than baseline
  disjoint:false:          OK
    41.5 ns ± 3.7 ns, 360 B  allocated,   0 B  copied, 6.0 MB peak memory, 33% less than baseline
  disjoint:true:           OK
    204  ns ±  14 ns, 920 B  allocated,   0 B  copied, 6.0 MB peak memory, 14% less than baseline
  isSubsetOf:false:        OK
    25.3 ns ± 1.7 ns, 384 B  allocated,   0 B  copied, 6.0 MB peak memory, 27% less than baseline
  isSubsetOf:true:         OK
    206  ns ±  15 ns, 1.4 KB allocated,   0 B  copied, 6.0 MB peak memory, 68% more than baseline
  null.intersection:false: OK
    161  ns ±  14 ns, 1.4 KB allocated,   0 B  copied, 6.0 MB peak memory, 44% less than baseline
  null.intersection:true:  OK
    90.2 ns ± 7.5 ns, 496 B  allocated,   0 B  copied, 6.0 MB peak memory, 57% less than baseline
```


containers/src/Data/Set/Internal.hs
```diff
@@ -535,26 +535,6 @@ lazy :: a -> a
 lazy a = a
 #endif
 
--- Insert an element to the set only if it is not in the set.
--- Used by `union`.
-
--- See Note: Type of local 'go' function
--- See Note: Avoiding worker/wrapper (in Data.Map.Internal)
-insertR :: Ord a => a -> Set a -> Set a
```
alexfmpe, line 543: Removed to get rid of unused warning due to me experimenting with everything else in-place rather than new variants

meooow25:
Pretty interesting to see this comparison, thanks for benchmarking these.

> We could merely add new variants but it might also be worth looking at the sizes/ratios of input trees to pick the implementation.

Yes, if this strategy is decently better for comparable sizes it is an option to implement it and use it only for those cases.

> In either case, I think we need a lot more benchmarks to have a better idea when any of this is actually worth it. In particular, it needs to be easy to see how results vary as size(s) vary.

Try the `set-operations-set` benchmark, it tests less functions but with a few different sizes: https://github.com/haskell/containers/blob/0c3b9ee93af23f86fb8ad5ffe7729cb82fc4cd3a/containers-tests/benchmark-utils/SetOperations.hs#L34-L37

---

Regarding the list-based implementation, I would also try using the iterator bits here (https://github.com/haskell/containers/blob/0c3b9ee93af23f86fb8ad5ffe7729cb82fc4cd3a/containers/src/Data/Set/Internal.hs#L1303) in place of `toAscList`. It doesn't affect the complexity, but might be faster in practice.

I recall you also had an experiment with unbalanced splitting at #865. I messed around a little bit with that in https://github.com/haskell/containers/pull/865#issuecomment-1374524234 but never looked into it properly after that. So it might be possible to make the current implementations faster as well.

jwaldmann:
The search tree invariant gives for each sub-tree, an interval that contains all the keys, and the tree-based (splitting) implementation uses this (implicitly), and it helps when intervals have empty intersection. How does the suggested toList/fromList code perform there? E.g. (with `ghci -package containers`)

```
:set +s

import qualified Data.Set as S

a = S.fromList  [1 .. 10^6::Int ]
sum a -- to force evaluation

b = S.map (+ 10^6) a
sum b

S.intersection a b  -- this is the actual test
fromList []
(0.00 secs, 93,688 bytes)
```
Now this  uses disjoint-ness at the roots, which may be uncommon, but the effect may happen for subtrees, and that's harder to predict.

alexfmpe:
> Try the set-operations-set benchmark, it tests less functions but with a few different sizes: 

Oh was unaware we had that. Gave it a spin.
Outside of the `nn` cases, tree-splitting is often orders of magnitude faster than list-merging. That much was expected but the `nn` cases are also 2-6x faster for tree-splitting than list-merging

> Yes, if this strategy is decently better for comparable sizes it is an option to implement it and use it only for those cases.

Turns out it's not merely about size, but also about the relative order of elements. Consider `disjoint xs ys` where every `x` is smaller than every `y`. Tree-splitting will first grab the root of `xs`, say, `Bin _ x l r`, then split `ys` with it in `lg |ys|` time. Given `x` is smaller than any `y`, this will split `y` into itself and an empy set. We then recurse into `disjoint mempty l && disjoint y r` and halve `xs` on each iteration so we end up with `lg |xs| * lg |ys|` regardless of their sizes. 

AFAICT, list-match only pays off both when 
  1. sets are of "close enough" size
  2. their values are sufficiently interleaved order-wise that split ends up taking `O(n)` size

Put another way, the interleaving of the two sets must be of size "close enough" to that of the larger set.

Which is exactly what's going on with our benchmarks based on `s`, `s_odd`, `s_even`, so that had painted a much more favorable picture.

While we could probably figure out (1), (2) is at most known to the caller, so I think if we're doing anything with the list implementations, they only make sense as separate bindings with comments detailing the happy path conditions.

> Regarding the list-based implementation, I would also try using the iterator bits here

Ah nice, I've been wondering how to avoid the conversion into lists without `fold`
Will try it out. By-product: https://github.com/haskell/containers/pull/1258

PR #1258, opened 2026-09-19
 Generalise EqM/OrdM to decidable monoids 

alexfmpe:
Name is tentative

```
Set 
  eq:                           OK
    32.6 μs ± 2.1 μs, 128 KB allocated,  29 B  copied,  44 MB peak memory,       same as baseline
  compare:                      OK
    32.3 μs ± 1.9 μs, 128 KB allocated,  28 B  copied,  44 MB peak memory,       same as baseline


All
  eq:      OK
    167  μs ±  16 μs, 640 KB allocated,  60 B  copied,  21 MB peak memory,       same as baseline
  compare: OK
    167  μs ±  13 μs, 640 KB allocated,  58 B  copied,  21 MB peak memory,       same as baseline
```


containers/src/Utils/Containers/Internal/ShortCircuit.hs
```diff
@@ -1,38 +1,22 @@
...
+-- | Composes left-to-right, short-circuits on monoid identity
+newtype ShortCircuit m a = ShortCircuit { runShortCircuit :: a -> StrictPair m a }
 
--- | Composes left-to-right, short-circuits on False
-instance Semigroup (EqM a) where
-  f <> g = EqM $ \x -> case runEqM f x of
-    r@(e :*: x') -> if e then runEqM g x' else r
+instance (Eq m, Monoid m) => Semigroup (ShortCircuit m a) where
+  f <> g = ShortCircuit $ \x -> case runShortCircuit f x of
+    r@(e :*: x') -> if e == mempty then runShortCircuit g x' else r
```
alexfmpe, line 16: I assume `e == All True` will always optimize to `if e`

meooow25:
I'm not sure about this. `EqM` and `OrdM` are easy to understand on their own, and I find that this change makes it harder to understand. The relationship with monoids is not reflected in the code at all: it doesn't use `<>` and there is no explanation for why it should test against `mempty`.  
It would be nice if it were generalizing to some well-known structure, but that's not the case either. I can't find anything on "decidable monoids" with a Google search.

---

If I think of generalizing this, I would proceed a bit differently.

First, we need the property that `<>` returns one of its arguments (I don't know what this is called either).

```hs
data Which = L | R
instance Semigroup a => ChoosesOne a where
  -- Law: (x <> y) == case sappendWhich x y of { L -> x; R -> y }
  sappendWhich :: a -> a -> Which
```

Then we can have

```hs
choosePair :: ChoosesOne a => StrictPair s a -> StrictPair s a -> StrictPair s a
choosePair p1@(_ :*: x1) p2(_ :*: x2) = case sappendWhich x1 x2 of { L -> p1; R -> p2 }

newtype ChooseWithState s a = CWS (s -> StrictPair s a)

instance ChoosesOne a => Semigroup (ChooseWithState s a) where
  CWS f1 <> CWS f2 = CWS $ \s0 ->
    let p1@(_ :*: s1) = f1 s0
    in choosePair p1 (f2 s1)
```

With another property that the semigroup has left-zeros, we can optimize it to the current implementation.

```hs
instance Semigroup a => HasLeftZeros a where
  -- Law: isLeftZero x <=> x <> y == x
  isLeftZero :: a -> Bool

newtype ShortCircuitChooseWithState s a = SCCWS (s -> StrictPair s a)

instance (ChoosesOne a, HasLeftZeros a) => Semigroup (ShortCircuitChooseWithState s a) where
  SCCWS f1 <> SCCWS f2 = SCCWS $ \s0 ->
    let p1@(x1 :*: s1) = f1 s0
    in if isLeftZero x1
       then p1
       else f2 s1
```

All this is to say that I find it convoluted and don't see the benefit in trying to generalize it.

